PR #36, opened 2022-12-28
Hot fix: result incorrect if fallback solution is gsu kernel

jichangjichang:
Without this fix, the test result is incorrect by running test command:
./example_hipblaslt_preference --datatype fp16 -m 16 -n 1 -k 4608 --beta 0 --alpha 1 -V
This patch can fix that settings are mismatch for solution searching and solution running.


jichangjichang:
CQE has confirmed that this PR is ok to merge.

PR #42, opened 2023-01-09
Run TensileCreateLibrary once in TensileConfig.cmake

KKyang:
The loading and create solution time of the logic yamls are too long, so we change the behavior in cmake.

Before:
Create manifest at config time -> create CreateLibrary custom builds -> use the files in the manifest as DEPENDS

New:
Create CreateLibrary custom builds -> use the manifest itself as DEPENDS, and check if the generated files exist in Python


jichangjichang (approved):
LGTM

PR #52, opened 2023-03-07
Add AUX Support

KKyang:
Add AUX support.

Bug fix:
1. Fix no branch code if number of lines in activation exceeds 16384.
2. Fix hip invalid arguments error when bound check is set to GuardPageBack.


tensilelite/Tensile/Source/lib/source/ContractionSolution.cpp
```diff
@@ -941,6 +773,7 @@ namespace Tensile
     {
         TensorDescriptor const& c = problem.c();
         TensorDescriptor const& d = problem.d();
+        TensorDescriptor const& e = problem.d();
```
hcman2, line 776: Should it be "problem.e();" ?

KKyang: fixed

aazz44ss (approved):

PR #55, opened 2023-03-10
Add gradient gelu

KKyang:
Add gradient gelu with the following limitations:
1. Bias is temporarily disabled.

This version does not have a NaN guard.


aazz44ss (approved):
LGTM

PR #56, opened 2023-03-10
hipblaslt-bench: remove the dependency with hipblas library

jichangjichang:
hipblaslt-bench doesn't need to link hipblas library


Jinp800125 (approved):
approve

PR #60, opened 2023-03-16
hipblaslt-test: remove the dependency with hipblas library

jichangjichang:
hipblaslt-test doesn't need to link to libhipblas.so


jichangjichang (commented):
LGTM

PR #84, opened 2023-04-19
Fix compiler warnings and check if algo is valid

KKyang:
As tile.


library/src/amd_detail/rocblaslt/src/rocblaslt_mat.cpp
```diff
@@ -235,7 +236,8 @@ rocblaslt_status rocblaslt_groupedgemm_create_impl(rocblaslt_groupedgemm
         if(validArgs == rocblaslt_status_success)
             continue;
         if(validArgs != rocblaslt_status_continue)
-            return validArgs;;
+            return validArgs;
+        ;
```
jichangjichang, line 240: remove this line

jichangjichang (approved):
LGTM

PR #125, opened 2023-05-25
Fallback to C++17

vin-huang:
googletest v13 and lower has bug at using std::u8string on RedHeat8 with C++20
 (refer https://github.com/google/googletest/pull/3703)
so fallback to C++17.


jichangjichang (approved):
LGTM

vin-huang:
local test passed, hipblaslt-test and tensile tests.

PR #130, opened 2023-05-30
Fix incorrect precision type of norm_check.

vin-huang:


vin-huang:
local test passed

jichangjichang (approved):
LGTM

PR #135, opened 2023-05-31
Fix tensile problem not update correctly and move contraction problem to matmul_desc

KKyang:
Will cause error if the transpose type changed. Fixed after adding index to updateProblem Tensile API


KKyang:
hipblas-test passed

jichangjichang (approved):
LGTM

aazz44ss (approved):

PR #140, opened 2023-06-02
Fix shared_ptr memory leak issue

KKyang:


jichangjichang (commented):
Please find anothor way that the mem can be actaullly freed along with algo struct.

PR #143, opened 2023-06-05
Add ext to gtest and change GG API setProblem

KKyang:
The GG API setProblem input arg problem type was
```
std::vector<GemmProblemType>& problemtype
```
is changed to
```
GemmProblemType& problemtype
```
to avoid confusion since currently hipblaslt only supports same GemmProblemType for all GG


KKyang:
Local hipblaslt-test passed

jichangjichang (approved):
LGTM

PR #144, opened 2023-06-05
Save solution index in algo instead of shared_ptr

KKyang:


KKyang:
Local hipblaslt-test passed

jichangjichang (approved):
LGTM

PR #154, opened 2023-06-20
Added softmax extension op

Serge45:
## Brief ##
This PR adds softmax extension op, including
 - ASM code generator for softmax kernel
 - Integration of code generator and hipBLASLt
 - `hipblasltExtSoftmax` API was added.

## Limitations ##
Currently this API has the following limitations
 - Support F32 as input/output tensor only
 - Support softmax along dimension 1 only
 - The dimension 1 must <= 256
 - It assumes input tensor is row-major.


TonyYHsieh (approved):
LGTM

PR #157, opened 2023-06-26
Update build options and docs

cgmb:
I was looking at hipBLASLt in relation to SWDEV-407636, and I had a few suggestions. rocSOLVER and rocFFT provide this same `WERROR` CMake option. I also noticed a couple minor documentation issues and figured they should be addressed as well.

#### Disable -Werror by default

The -Werror option should be disabled by default, as it is likely that users building with different versions of the compiler, different versions of third-party dependencies, or different build flags will encounter at least one stray warning. The option can be explicitly enabled in controlled environments (like the math-ci).

#### Update URLs in README

There are a few URLs in README.md that are broken or outdated. I've updated them to the most current page.

#### Update package description

The package description appears to still have been the hipBLAS description. I'm not very familiar with the hipBLASLt library, so feel free to suggest something better.


cgmb:
The main issue was fixed in https://github.com/ROCmSoftwarePlatform/hipBLASLt/commit/3b0b5c0ed105dfbf8ea25f0f641d41424699ad99.

PR #169, opened 2023-07-05
[Tensilelite] Fix GSU for int8

vin-huang:
1. fix gsu for i8
2. add test case for gsu: I8/I8/s, I8/I/s, I8/I8/I, I8/I/I
3. enable i8 gsu tests for sparse


jichangjichang (approved):
LGTM, please merge after CI pass

PR #177, opened 2023-07-13
Use TensileInstructions to instead globalParameters when need to get the CurrentISA

vin-huang:


tensilelite/Tensile/KernelWriter.py
```diff
@@ -448,8 +448,7 @@ def __init__( self, kernelMinNaming, kernelSerialNaming ):
   ##############################################################################
   def makeSchedule(self, kernel, tensorParametersA, tensorParametersB, localWriteEndIter, uDu=0, skipGlobalReadInc=False, firstIter=False, lastLoop=False, lastLc=False):
 
-    currentIsa = globalParameters["CurrentISA"]
-    maxVmcnt = globalParameters["AsmCaps"][currentIsa]["MaxVmcnt"]
+    maxVmcnt = self.ti.getAsmCaps()["MaxVmcnt"]
```
jichangjichang, line 451: here we can git Asmcaps from self.states

vin-huang: done

PR #179, opened 2023-07-16
[Tensilelite] use TensileInstructions to instead globalParameters when

vin-huang:
need to get the CurrentISA


jichangjichang (approved):
LGTM

PR #210, opened 2023-07-28
Bias stride

vin-huang:
The support of the stride batch for bias is needed by hipSparseLt.
hipSparselt allows the user to set stride=0 to broadcast the same bias vector to each batch,
 or stride >=M, in this case, each batch will have its own bias vector

1. Support bias stride for the bias source = D
2. fix lds conflict between bias and Matrix C
3. add test case


KKyang (approved):

jichangjichang:
Could you share the background for this feature here?

PR #241, opened 2023-08-08
alpha vector sample

Jinp800125:


clients/samples/gemm/sample_hipblaslt_gemm_ext.cpp
KKyang, line 1: Use a new sample instead of modifying the old one, e.g. ``samples/gemm_scaleAlphavec/sample_hipblaslt_gemm_scaleAlphavec_ext.cpp``

KKyang (changes requested):

clients/samples/gemm/sample_hipblaslt_gemm_ext.cpp
Jinp800125: Sure. Before this already told the user this change just a temp will separate to another folder if needed.

jichangjichang (dismissed):
you should go legacy cublaslt api but not hipblaslt owned ext api for alpha vector sample.

jichangjichang:
> you should go legacy cublaslt api but not hipblaslt owned ext api for alpha vector sample.

I heared from yangwen that the ext sample request is from customer. So I removed the my request change.

PR #257, opened 2023-08-14
414231 - folder structure for cmake config files

arvindcheru:
Proposed change
- Update cmake config files install folder to ROCM_PATH with out prefix


jichangjichang (approved):
lgtm

PR #269, opened 2023-08-23
Add FP8FP16_FP8SS

KKyang:


KKyang:
Local hipblaslt-test passed

tensilelite/Tensile/KernelWriterAssembly.py
```diff
@@ -5237,6 +5238,7 @@ def globalReadGuardK(self, kernel, tP):
 
     def globalReadGuardKBody(tP):
       tc = tP["tensorChar"]
+      tc2 = "" if tc == "Metadata" else tc
```
jichangjichang, line 5241: can we deprecate tc2, and just replase tc as "" for Metadata

KKyang: Chaaged to tcDataType instead

PR #281, opened 2023-08-29
Change Ti to TiA/TiB in testing_matmul

KKyang:


CMakeLists.txt
```diff
@@ -25,7 +25,7 @@
 cmake_minimum_required(VERSION 3.5 FATAL_ERROR)
 
 # This will add compile option: -std=c++17
-set(CMAKE_CXX_STANDARD 17 )
+set(CMAKE_CXX_STANDARD 20 )
```
jichangjichang, line 28: @vin-huang 
Do you remember that we countered some errors if change the CXX version? I fotgot the details.

vin-huang: In the previous case, C++20 will suffer problems building GTest on the SLES, Redheat OS.
in this PR, Gtest Version has also been updated to v1.14. may check this version to see if the Gtest bug was fixed or not.

KKyang: Passed on rhel-8.x

jichangjichang (approved):
LGTM

PR #325, opened 2023-09-18
Add new options to hipblaslt-bench

KKyang:
Add options ``algo_method`` and ``requested_solution``.

The option ``algo_method`` is to choose which searching API.
0-	getHeuristic
1-	getAllAlgos
2-	getSolutionByIndex
Currently 2 does not allow the user to choose their own index but only use the API to call the first runnable kernel for any input problem.

The option ``api_method`` is to choose which APIs to call,
0-	The cublaslt style API
1-	The cublaslt style matrix declaration + C++ style API
2-	Pure C++ style API

``requested_solution`` is only valid when ``algo_method`` is set to ``0``.

Example command to run getAllAlgos with C++ style API with default size,
```
./clients/staging/hipblaslt-bench --api_method 2 --algo_method 1
```

### DIFF
--- a/clients/benchmarks/README.md
+++ b/clients/benchmarks/README.md
@@ -30,7 +30,7 @@ cd hipBLASLt; cd build/release
 --d_type <value>           Precision of matrix D. Options: f32_r,f16_r,bf16_r
 --compute_type <value>     Precision of computation. Options: s,f32_r,x,xf32_r,f64_r,i32_r                     (Default value is: f32_r)
 --scale_type <value>       Precision of scalar. Options: f16_r,bf16_r
---initialization <value>   Intialize matrix data.Options: rand_int, trig_float, hpl(floating)                  (Default value is: hpl)
+--initialization <value>   Intialize matrix data.Options: rand_int, trig_float, hpl(floating)                  (Default value is: rand_int)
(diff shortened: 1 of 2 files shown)


jichangjichang:
hit errors on gfx90a:
./clients/staging/hipblaslt-bench --requested_solution 2
corrupted size vs. prev_size
Aborted (core dumped)

clients/include/testing_matmul.hpp
```diff
@@ -832,8 +832,9 @@ void testing_matmul(const Arguments& arg)
 
     // Get Heuristic results
     std::vector<hipblasLtMatmulHeuristicResult_t> heuristicResult{1};
```
jichangjichang, line 834: cannot hard-code it as 1 element now

jichangjichang (changes requested):
check comment

clients/include/testing_matmul.hpp
KKyang: >> besides that, don't use vector for non-ext path.
This should open another PR for that instead of here cause it's not related to the commit.

jichangjichang (approved):
LGTM

Jinp800125 (commented):
done

Jinp800125 (approved):

PR #332, opened 2023-09-19
Enable preload kernel args feature on all GroupedGemm kernels for gfx941

jichangjichang:


jichangjichang:
I would enable CI once the complier changes are in math CI build.

KKyang (approved):

PR #334, opened 2023-09-19
replace build.image with build.os in the .readthedocs.yaml config

samjwu:
https://docs.readthedocs.io/en/stable/config-file/v2.html

build.image (currently used implicitly) will be deprecated in October 2023


KKyang (approved):

jichangjichang (approved):
lgtm

PR #339, opened 2023-09-22
Disable fallback when epilogue is not set to DEFAULT

KKyang:


library/src/amd_detail/rocblaslt/src/tensile_host.cpp
```diff
@@ -1232,13 +1234,15 @@ rocblaslt_status gemmCreate(RocblasltContractionProblem const& problem,
             std::shared_ptr<TensileDataGemm> data
                 = std::static_pointer_cast<TensileDataGemm>(gemmData);
             updateTensileProblem(false, problem, data->problem);
-            data->inputs = GetTensileInputs(problem);
+            data->inputs         = GetTensileInputs(problem);
+            data->enableEpilogue = problem.epilogue == ROCBLASLT_EPILOGUE_DEFAULT ? true : false;
```
jichangjichang, line 1238: roblem.epilogue == ROCBLASLT_EPILOGUE_DEFAULT -> data->enableEpilogue = true?

KKyang: fixed

jichangjichang (approved):
lgtm

vin-huang (approved):

PR #357, opened 2023-10-03
Amend ext ops library runtime/generating flow

Serge45:
## Brief ##
This PR attempts to amend runtime/generating flow of ext op library.

## Implemented ##
 - Properly setup output dependency of ext op target.
 - Expose part of logging APIs to `rocblaslt` layer.
 - Convert ext op initialization failures to error log and throwing the exception during the first ext op API call.
 - Fix undefined behavior caused by nondeterministic initialization order of static variables across different translation units(log singleton and ext op library singleton).


TonyYHsieh (approved):
LGTM

KKyang (approved):

PR #359, opened 2023-10-04
Fix yamls

KKyang:
The default distance type in LibraryIO is Euclidean. This commit changes to None.
All yamls must assign their distance types.

Also fix lazy loading build error.


jichangjichang (approved):
LGTM

aazz44ss (approved):

PR #364, opened 2023-10-05
Use hipblaslt_initialization for input data initialization of ext op

Serge45:
## Brief ##
Use `hipblaslt_initialization` to initialize input data for ext op samples.

## Implemented
 - [x] updated init method for softmax, layernorm and matrix transform
 - [x] added missing license agreement in samples.

### DIFF
--- a/clients/benchmarks/client_extop_matrixtransform.cpp
+++ b/clients/benchmarks/client_extop_matrixtransform.cpp
@@ -25,6 +25,8 @@
  *******************************************************************************/
 #include <algorithm>
 #include <hip/hip_runtime.h>
+#include <hipblaslt_datatype2string.hpp>
+#include <hipblaslt_init.hpp>
 #include <hipblaslt/hipblaslt.h>
 #include <iostream>
 #include <memory>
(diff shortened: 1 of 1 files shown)


clients/benchmarks/client_extop_layernorm.cpp
```diff
@@ -173,13 +214,36 @@ void compare(const char* title, const std::vector<T>& cpuOutput, const std::vect
     std::cout << "max error : " << maxErr << std::endl;
 }
 
+template <typename DType>
+void initData(DType* data, std::size_t numElements, hipblaslt_initialization initMethod)
+{
+    switch(initMethod)
+    {
+    case hipblaslt_initialization::rand_int:
+        hipblaslt_init<DType>(data, numElements, 1, 1);
```
TonyYHsieh, line 223: minus quest : should lda value be numElements?

lda == M. is it correct ?
template <typename T>
inline void hipblaslt_init(
    std::vector<T>& A, size_t M, size_t N, size_t lda, size_t stride = 0, size_t batch_count = 1)
{
    hipblaslt_init(A.data(), M, N, lda, stride, batch_count);
}

TonyYHsieh (approved):

KKyang (approved):

PR #367, opened 2023-10-06
[Fix] sizeWritten supports nullptr

wenchuanchen:
From this JIRA ticket: https://ontrack-internal.amd.com/browse/SWDEV-423172, where users want to use sizeWritten with nullptr as well. Also add testcases to gtest bad arg.


KKyang (approved):

jichangjichang (approved):
LGTM

PR #370, opened 2023-10-06
Add --solution_index support to hipblaslt-bench

dllehr-amd:
When using `--algo_method 2`.  The current method will find the first solution that works and benchmark it.  The preferred behavior is to take a `--solution_index <sol>`  which hipblaslt-bench can use as the preferred solution and benchmark with that.

Since `arg.solution_index` wasn't used for anything at the moment, I used it's placeholder to pass in the desired solution to the `testing_matmul` code.

### DIFF
--- a/clients/benchmarks/client.cpp
+++ b/clients/benchmarks/client.cpp
@@ -407,12 +407,12 @@ try
 
         ("algo_method",
          value<int32_t>(&arg.algo_method)->default_value(0),
-         "Use different algorithm search API. 0: Get heuristic, 1: Get all algorithm, 2: Get solutuion by index."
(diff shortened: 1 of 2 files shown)


shajrawi (approved):
I am not a fan of algo_method being an integer number instead of a descriptive string that is easier for an outsider to grok, but that's just a small nit :)

KKyang (commented):
Could you change the code to "if the solution index is not given, it'll go the original behavior" else gtest will fail at this moment. Thanks.

We use gtest to see if the API works normally so it is designed like this.

KKyang:
Closing this PR due to #371 merged.

PR #371, opened 2023-10-07
Add --solution_index support

hcman2:
This change is based on PR #370.
Add an error handling and keep the original behavior when user didn't set solution index.
@dllehr-amd , Could you try this patch to make sure it works well for your requirements.

### DIFF
--- a/clients/benchmarks/client.cpp
+++ b/clients/benchmarks/client.cpp
@@ -426,8 +426,8 @@ try
          "Use different algorithm search API. Options: heuristic, all, index.")
 
         ("solution_index",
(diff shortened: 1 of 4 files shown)


dllehr-amd:
@hcman2  I'll try it out on my system.  This looks like it will satisfy both of our cases.  I'll let you know as soon as I test it!  Thanks for the PR!

shajrawi (approved):
I'll leave final approval to @dllehr-amd but LGTM

dllehr-amd (approved):
LGTM!

PR #388, opened 2023-10-18
Add HSS to gfx94x

KKyang:
gfx94x is missing gridbased HSS yamls. Add them in this commit.

AB type: float16
CD type: float
Bias: FP32 or FP16


amathews-amd (approved):
LGTM

PR #412, opened 2023-11-01
Keep debug info for debug and release with debug info build

Serge45:
## Brief ##
This PR adds ability to keep debug information for assembly kernels in debug and release with debug build.

## Details ##
 - Using `-g` or `-k` in `install.sh` will build whole library with assmembly debug information
 - For `TensileCreateLibrary`, use `--asm-debug` to enable it
 - For `Tensile`, simply add `AsmDebug: True` to tuning YAML's global parameter section.

## Misc ##
To check the debug information is kept or not, use `file <code_object_path>`, it shows

```
./release-debug/Tensile/library/TensileLibrary_gfx942.co: ELF 64-bit LSB shared object, *unknown arch 0xe0* version 1, dynamically linked, with debug_info, not stripped, too many notes (256)
```

for building with debug info. And it shows

```
./TensileLibrary_gfx942.co: ELF 64-bit LSB shared object, *unknown arch 0xe0* version 1, dynamically linked, not stripped, too many notes (256)
```
without debug info,

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -367,7 +367,6 @@ tensile_version=
 build_tensile=true
 tensile_msgpack_backend=true
 update_cmake=true
-asm_debug=false
 
 
 rocm_path=/opt/rocm
@@ -415,7 +414,6 @@ while true; do
             shift ;;
         -g|--debug)
             build_release=false
-            asm_debug=true
             shift ;;
         --hip-clang)
             build_hip_clang=true
@@ -429,7 +427,6 @@ while true; do
         -k|--relwithdebinfo)
             build_release=false
             build_release_debug=true
-            asm_debug=true
             shift ;;
         --codecoverage)
             build_codecoverage=true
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -707,6 +710,10 @@ pushd .
     tensile_opt="${tensile_opt} -DTensile_LIBRARY_FORMAT=yaml"
   fi
 
+  if [[ "${asm_debug}" == true ]]; then
```
jichangjichang, line 713: That checking build_release == false should be enough.
I didn't see the necessary to add asm_debug here unless you want to add another argument for install.sh to control the TENSILE_ASM_DEBUG directly.

KKyang:
Isn't d stands for dependency?

install.sh
Serge45: Yea, you're right, will change soon.

Serge45:
Yea, typo, I meant `-g`.

KKyang (approved):

jichangjichang (approved):
lgtm

PR #422, opened 2023-11-03
Correcting argument parsing for short form

mahmoodw:
Originally noticed that -a flags was not working instead of using --architecture. Corrected all options missing the ":"

Additionally, there are many install script options that are not documented in the help section. Is this because they are not tested yet? e.g. cov, fork, branch, no-tensile, etc.


KKyang (approved):

jichangjichang (approved):
lgtm

PR #429, opened 2023-11-09
Add rotating buffer

KKyang:
Auto calculated, only size in MB is needed.

### TensileLite
Supports groupedgemm/ gemm.

#### Usage
globalParameters["RotatingBufferSize"] # Size in MB

### hipBLASLt
Currently only supports gemm + api_method c only.

#### Usage
--rotating X (X is size in MB)

### DIFF
--- a/clients/benchmarks/client.cpp
+++ b/clients/benchmarks/client.cpp
@@ -522,6 +522,10 @@ try
          value<bool>(&arg.print_kernel_info)->default_value(false),
          "Print solution, kernel name and solution index.")
 
+        ("block_count",
+         value<int32_t>(&arg.block_count)->default_value(1),
+         "Use rotating memory blocks for each iteration. Only works for api_method = 0.")
+
         ("help,h", "produces this help message")
 
         ("version", "Prints the version number");
--- a/clients/benchmarks/client.cpp
+++ b/clients/benchmarks/client.cpp
@@ -522,9 +522,9 @@ try
          value<bool>(&arg.print_kernel_info)->default_value(false),
          "Print solution, kernel name and solution index.")
 
-        ("block_count",
-         value<int32_t>(&arg.block_count)->default_value(1),
-         "Use rotating memory blocks for each iteration. Only works for api_method = 0.")
+        ("rotating",
+         value<int32_t>(&arg.rotating)->default_value(0),
+         "Use rotating memory blocks for each iteration, size in MB. (Only supports GEMM + api_method c)")
 
         ("help,h", "produces this help message")
 
@@ -581,6 +581,14 @@ try
     else
     {
         hipblaslt_cerr << "Invalid algo method: " << algo_method_str << std::endl;
+        return 1;
+    }
+
+    // Unblock this after rotating supports all modes
+    if((arg.rotating > 0) && (api_method != 0 || arg.grouped_gemm))
+    {
+        hipblaslt_cerr << "Currently rotating buffer only supports GEMM + api_method c."
+                       << std::endl;
         return 1;
     }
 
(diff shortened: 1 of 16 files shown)


aazz44ss (approved):

clients/benchmarks/client.cpp
```diff
@@ -522,6 +522,10 @@ try
          value<bool>(&arg.print_kernel_info)->default_value(false),
          "Print solution, kernel name and solution index.")
 
+        ("rotating",
+         value<int32_t>(&arg.rotating)->default_value(0),
+         "Use rotating memory blocks for each iteration, size in MB. (Only supports GEMM + api_method c)")
+
```
jichangjichang, line 528: return  error if user enable rotating buffer for non-supported cases.

clients/include/testing_matmul.hpp
```diff
@@ -631,6 +662,37 @@ void testing_matmul(const Arguments& arg)
...
+            dE[i] = nullptr;
+        }
+
+        if(block_count > 1)
+        { // rotating blocks
+            alt_dA[i]      = new device_vector<TiA>(size_A[i] * (block_count - 1), 1, HMM);
+            alt_dB[i]      = new device_vector<TiB>(size_B[i] * (block_count - 1), 1, HMM);
+            alt_dC[i]      = new device_vector<To>(size_C[i] * (block_count - 1), 1, HMM);
+            alt_dD[i]      = new device_vector<To>(size_D[i] * (block_count - 1), 1, HMM);
+            alt_dE[i]      = new device_vector<To>(size_E[i] * (block_count - 1), 1, HMM);
```
jichangjichang: don't need to check arg.use_e?

clients/include/testing_matmul.hpp
```diff
@@ -811,21 +863,51 @@ void testing_matmul(const Arguments& arg)
...
+                CHECK_HIP_ERROR(alt_dBias[i]->transfer_from(*hBias[i], block_count - 1));
+                CHECK_HIP_ERROR(alt_dBias_C[i]->transfer_from(*hBias_C[i], block_count - 1));
+            }
         }
 
+        if(arg.scaleAlpha_vector)
+        {
+            CHECK_HIP_ERROR(dScaleAlphaVec[i]->transfer_from(*hScaleAlphaVec[i]));
+            alpha_in[i] = *(dScaleAlphaVec[i]);
+            h_alpha[i]  = 1.0; // use dScaleAlphaVec instead, original alpha = 1.0 for verify
```
jichangjichang, line 895: if it is debug code or unused code, please remove it

clients/include/testing_matmul.hpp
```diff
@@ -631,6 +662,37 @@ void testing_matmul(const Arguments& arg)
...
+        {
+            dE[i] = new device_vector<To>(size_E[i], 1, HMM);
+            CHECK_DEVICE_ALLOCATION(dE[i]->memcheck());
+        }
+        else
+        {
+            dE[i] = nullptr;
+        }
+
+        if(block_count > 1)
```
jichangjichang: is that possible to allocate block_count x size_X in one shot?
That separating dX[i] and alt_dX[i] produces many duplicated codes.

clients/benchmarks/client.cpp
KKyang: It'll just ignore instead. Change to return error?

clients/include/testing_matmul.hpp
KKyang: Not necessarily but can.

clients/benchmarks/client.cpp
jichangjichang: better to returing error to let user know, since user may expect some perf different comparing with no rotating buffer.

clients/include/testing_matmul.hpp
KKyang: Maybe next version with other changes?

clients/benchmarks/client.cpp
KKyang: Fixed

clients/include/testing_matmul.hpp
KKyang: Fixed

jichangjichang (approved):
lgtm

PR #431, opened 2023-11-10
Force GemmInstance move only

Serge45:
## Brief ##
Make `GemmInstance` and its derives move only.

## Details ##
The auto-conversion feature for gfx90a makes the `GemmInstance` non-copyable, this PR adds
 - Refactored conversion details to `ConversionHelper`
 - Forced `GemmInstance` and its derives to use their move constructor, this makes compiler generates move only codes.


jichangjichang:
add original failure code into sample app which would be verified by  ctest

Serge45:
Yea, I'm thinking about adding to test suite or sample app, but both of them need a scenario to put the strcuts to STL container. Anyway, I can design an artificial scenario for testing first.

aazz44ss (approved):

KKyang (approved):

PR #435, opened 2023-11-14
Universal GSU Kernel

KKyang:


tensilelite/Tensile/Common.py
```diff
@@ -999,7 +1009,10 @@ def getArchitectureName(gfxName):
     "WorkGroupReduction":          [False],
 
     # 4:2 Structured Sparse A Matrix, 0=Non Sparse, 1=Sparse Matrix A, 2=Sparse Matrix B
-    "Sparse":                      [0, 1, 2]
+    "Sparse":                      [0, 1, 2],
+
+    # Information about user input internal kernel argument support
+    "SupportUserGSU":              [False, True],
```
jichangjichang: Don't add New Kernel Feature just for information purpose.

KKyang: You have to add this since you can't tell if a custom kernel supports GSU. @jichangjichang

KKyang: Or you can just turn it off if it is a custom kernel.

aazz44ss (approved):

tensilelite/Tensile/Common.py
KKyang: Fixed

jichangjichang (approved):
lgtm

PR #443, opened 2023-11-21
Add sample_hipblaslt_gemm_tuning_splitk_ext

KKyang:


clients/samples/gemm_tuning/sample_hipblaslt_gemm_tuning_splitk_ext.cpp
```diff
@@ -0,0 +1,170 @@
...
+        }
+    }
+
+    if(validIdx.empty())
+    {
+        std::cerr << "No valid solution found!" << std::endl;
+        return;
+    }
+
+    // In this sample, the workspace is already allocated with max_workspace_size
```
jichangjichang: if requried workspace size is less than ma_workspace_size, we should allocate it here.
There example is for GSU which is workspace sensitive feature. we should let user know it.

KKyang: fixed

jichangjichang (approved):
lgtm

aazz44ss (approved):

PR #445, opened 2023-11-21
Rotating, SplitK and bug fix

KKyang:
[Bench]
Support rotating for all modes
Support multi-splitk
Bug fix
1. Fix using gemm instance to check if grouped gemm solution is supported.
2. log_args
3. Recalculate _WorkspaceSizePerElem*


aazz44ss (approved):

hcman2 (approved):
cool

PR #449, opened 2023-11-22
Add option for gpu kernel time

wenchuanchen:
From: https://ontrack-internal.amd.com/browse/SWDEV-430621
Instead of using cpu time to measure kernel elapsed time, use gpu timer hipEventElapsedTime.
Add user option --use_gpu_timer to enable this new feature or not.


clients/benchmarks/client.cpp
```diff
@@ -531,6 +531,10 @@ try
          valueVec<uint32_t>(&gsu_vector),
          "[Tuning parameter] Set split K for a solution, 0 is use solution's default value. (Only support GEMM + api_method mix or cpp)")
 
+        ("use_hip_time",
+         value<bool>(&arg.use_hip_time)->default_value(false),
```
jichangjichang: suggest using "gpu_timer" as argument name

jichangjichang (approved):
lgtm

KKyang (approved):

PR #451, opened 2023-11-23
Add F8F8S NN/TN/NT FreeSize yaml

KKyang:


aazz44ss:
do we need f8f8S of gfx942?

PR #480, opened 2023-12-05
max GSU workspace size in predicate checking is 128MB

jichangjichang:
This PR prevent user from trying GSU for large size which would never get any benefit from GSU.
We should reject it directly even user gives enough workspace size.


hcman2 (approved):

Serge45 (approved):

KKyang (approved):

aazz44ss:
maybe can extend to 256MB?
like [2048,2048,65536], we can use MT256x256_GSU14 to fully utilize all CUs, also have a good granularity 0.98
it requires 2048x2048x4x14/1024/1024=224MB of workspace

jichangjichang:
> 48,6

The real tuning case for 2048,2048,65536 won't select 256x256 GSU14. 
256x112 with GSU 2 can fully utilize all CU for this size.

PR #482, opened 2023-12-05
Fix workspace usage when epilogue is set to DGELU only

KKyang:


jichangjichang (approved):
LGTM,
Is that possible to add a single command in ctest to test original failure case?

aazz44ss (approved):

PR #485, opened 2023-12-06
add log level bench

wenchuanchen:
https://ontrack-internal.amd.com/browse/SWDEV-427775

To print out log which can be applied on hipblaslt-bench like this:

hipblaslt-bench -m 1024 -n 512 -k 1024 --lda 1024 --ldb 1024 --ldc 1024 --ldd 1024 --lde 1024 --stride_a 0 --stride_b 0 --stride_c 0 --stride_d 0 --stride_e 0 --alpha 1 --beta 1 --transA N --transB N --batch_count 1 --a_type f16_r --b_type f16_r --c_type f16_r --d_type f16_r --compute_type f32_r --scale_type f32_r

Enabled by export HIPBLASLT_LOG_MASK=32

### DIFF
--- a/library/src/amd_detail/rocblaslt/src/rocblaslt_mat.cpp
+++ b/library/src/amd_detail/rocblaslt/src/rocblaslt_mat.cpp
@@ -200,7 +200,7 @@ rocblaslt_status rocblaslt_matmul_impl(const rocblaslt_handle       handle,
 
 // To deal with some arguments may be invalid
 #define GEN_BENCH_ARG(_fun, _type_str, _type) \
-    strlen(_fun(_type)) ? _type_str : "", \
+    strlen(_fun(_type)) && strcmp(_fun(_type), "invalid") ? _type_str : "", \
     _fun(_type)
 
         log_bench(__func__,
--- a/library/src/amd_detail/rocblaslt/src/rocblaslt_mat.cpp
+++ b/library/src/amd_detail/rocblaslt/src/rocblaslt_mat.cpp
@@ -197,12 +197,18 @@ rocblaslt_status rocblaslt_matmul_impl(const rocblaslt_handle       handle,
                                         workspaceSizeInBytes,
                                         stream};
     if (get_logger_layer_mode() & rocblaslt_layer_mode_log_bench) {
+
+// To deal with some arguments may be invalid
+#define GEN_BENCH_ARG(_fun, _type_str, _type) \
+    strlen(_fun(_type)) ? _type_str : "", \
+    _fun(_type)
+
         log_bench(__func__,
-                  "--sizem",
+                  "-m",
                   m,
-                  "--sizen",
+                  "-n",
                   n,
-                  "--sizek",
+                  "-k",
                   k,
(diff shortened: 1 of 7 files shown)


library/src/amd_detail/rocblaslt/src/rocblaslt_mat.cpp
```diff
@@ -195,54 +196,61 @@ rocblaslt_status rocblaslt_matmul_impl(const rocblaslt_handle       handle,
...
-                  "transA",
-                  opA == HIPBLAS_OP_N ? "N" : "T",
-                  "transB",
-                  opB == HIPBLAS_OP_N ? "N" : "T",
-                  "batch_count",
-                  num_batches_a,
-                  "m",
+    if (get_logger_layer_mode() & rocblaslt_layer_mode_log_bench) {
+        log_bench(__func__,
+                  "--sizem",
```
jichangjichang: better to use "-m"

library/src/amd_detail/rocblaslt/src/rocblaslt_mat.cpp
```diff
@@ -195,54 +196,61 @@ rocblaslt_status rocblaslt_matmul_impl(const rocblaslt_handle       handle,
...
+                  hipblasltDatatype_to_bench_string(bias_type),
+                  "--grouped_gemm",
                   grouped_gemm,
-                  "compute_type",
-                  compute_type,
-                  "bias",
-                  bias,
-                  "scaleAlphaVec",
+                  "--scaleAlpha_vector",
                   scaleAlphaVec);
```
jichangjichang: missing: scaleA,B,C,D, and some arguments for backward pass epilogues

KKyang: Doesn't this look too hardcoded? Why can't we just print the argv in client.cpp if it is only used in bench?

library/src/amd_detail/rocblaslt/src/utility.cpp
```diff
@@ -82,6 +82,31 @@ const char* hipblasltDatatype_to_string(hipblasltDatatype_t type)
...
+    case HIPBLASLT_R_8I:
+        return "i8_r";
+    case HIPBLASLT_R_32I:
+        return "i32_r";
+    case HIPBLASLT_R_8F_E4M3:
+        return "f8_r";
+    case HIPBLASLT_R_8F_E5M2:
+        return "bf8_r";
+    default:
+        return "";
```
jichangjichang: better return "invalid" for unkown type

jichangjichang (commented):
Do we need to support this logging feature for extension cpp API path?

KKyang (approved):
Better think of another way to represent the rocblaslt problem type in stead of the bench parameters.

wenchuanchen:
> Do we need to support this logging feature for extension cpp API path?

This can be supported in another PR, it's more complicated to decode arguments in ext path than default path.

jichangjichang (approved):
lgtm

PR #508, opened 2023-12-14
Fix unable to init when capturing.

KKyang:


library/include/hipblaslt.h
```diff
@@ -381,12 +381,14 @@ hipblasStatus_t hipblasLtGetArchName(char** archName);
...
  * context.
+ *  @param[in]
+ *  stream  hipStream used in hipGraph mode.
  *
  *  \retval HIPBLAS_STATUS_SUCCESS The allocation completed successfully.
  *  \retval HIPBLAS_STATUS_INVALID_VALUE \p handle == NULL.
  */
 HIPBLASLT_EXPORT
-hipblasStatus_t hipblasLtCreate(hipblasLtHandle_t* handle);
+hipblasStatus_t hipblasLtCreate(hipblasLtHandle_t* handle, hipStream_t stream = NULL);
```
jeffdaily, line 391: This does not match the cublasLtCreate() API spec.

https://docs.nvidia.com/cuda/cublas/index.html?highlight=cublasltcreate#cublasltcreate

Why do we need an optional stream?

KKyang: To get the capturing status. If we have another API without stream input, we may not need an optional stream.

PR #519, opened 2023-12-26
[Tensilelite] Reject the soluion when LBSPPA and LDSPAD is not 0, but the lds read/write address is inconsistence

vin-huang:
Under some MT, that said MT=96x32x32, MIWT=3x1, FP16, LBSPPA=128, LdsPADA=4, LSCA=64
The LDS address will not same under these 2 formulas below, then causes the read/write address inconsistent.

1. int(MTA / LBSPPA ) + int( LSCA * [0 ~ MIWTA-1] / LBSPPA)
2.  int((MTA+ LSCA * [0 ~ MIWTA-1])  / LBSPPA) )

### DIFF
--- a/tensilelite/Tensile/SolutionStructs.py
+++ b/tensilelite/Tensile/SolutionStructs.py
@@ -2799,8 +2799,8 @@ class Solution(collections.abc.Mapping):
           state["LdsPadA"] = 0
           if state["MatrixInstB"] == 1 and state["MatrixInstM"] == 16:
             state["LdsPadA"] = ((16 * state["VectorWidthA"] * state["ProblemType"]["DataType"].numBytes() + state["MacroTile0"] * state["ProblemType"]["DataType"].numBytes() * state["LocalReadVectorWidth"]) % 128) // state["ProblemType"]["DataType"].numBytes()
-          if state["GlobalReadVectorWidthA"] * state["ProblemType"]["DataTypeA"].numBytes() == 16 and state["LdsPadA"] == 0:
-            state["LdsPadA"] = 8 // state["ProblemType"]["DataTypeA"].numBytes()
+          if state["GlobalReadVectorWidthA"] * state["ProblemType"]["DataType"].numBytes() == 32 and state["LdsPadA"] == 0:
+            state["LdsPadA"] = 16 // state["ProblemType"]["DataType"].numBytes()
             if auto_LdsBlockSizePerPadA_for_mix:
               state["LdsBlockSizePerPadA"] = 128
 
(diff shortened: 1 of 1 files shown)


tensilelite/Tensile/SolutionStructs.py
```diff
@@ -2807,6 +2807,16 @@ def assignDerivedParameters(state):
           state["LdsPadB"] = 0
       assert(state["LdsPadB"] >= 0)
 
+    if state["LdsBlockSizePerPadA"] != 0:
```
jichangjichang: what will happen if we set LdsBlockSizePerPadA=-1 for odd MIWT0 ?

vin-huang: when state["LdsBlockSizePerPadA"] = -1, that will be assigned a value at [SolutionStructs.py#L2671](https://github.com/ROCm/hipBLASLt/blob/develop/tensilelite/Tensile/SolutionStructs.py#L2671)
Hence, at line 2810, there will be a real value (not -1).  then we can use this value to check if the LDS address is consistent.

jichangjichang: if user set -1 which mean let other kern-genator to produce the most proper setting for their config, then we should generate a least one kernel. If -1 for those odd MT0 would produce a non-zero value at [SolutionStructs.py#L2671](https://github.com/ROCm/hipBLASLt/blob/develop/tensilelite/Tensile/SolutionStructs.py#L2671) and rejected by here, it means -1 doesn't work.
What I mean is that we should give a solution at least for -1 setting.

vin-huang: OK, so we can set state["LdsBlockSizePerPadA"] to 0 for the inconsistent case instead of the reject.

aazz44ss (approved):

vin-huang:
gfx90a, 

local test tensilelite passed:
> =========== 15 passed, 25 skipped, 14 warnings in 1113.21s (0:18:33) ===========

local test hipblaslt

> [  PASSED  ] 10719 tests.
> hipBLASLt version: 700

tensilelite/Tensile/SolutionStructs.py
```diff
@@ -2696,6 +2696,26 @@ def assignDerivedParameters(state):
     if state["LdsBlockSizePerPadMetadata"] == -1:
       state["LdsBlockSizePerPadMetadata"] = state["LdsBlockSizePerPadA"]
 
+    if state["LdsBlockSizePerPadA"] != 0 :
```
jichangjichang, line 2699: If the value of ldsBlockSizePerPadA is set by user, we can reject it.
- Reject section can be here.
- Setting as 0 section should be in the handling of LdsBlockSizePerPad = -1.

vin-huang: yes, if the user sets LdsBlockSizePerPadA = -1 will set LdsBlockSizePerPadA to 0, otherwise will be rejected.

          if auto_LdsBlockSizePerPadA_for_mix:   # LdsBlockSizePerPadA=-1
            print("Paded addres is inconisstent, set LdsBlockSizePerPadA=0.")
            state["LdsBlockSizePerPadA"] = 0
            break
          else:
            reject(state, "A's paded addres is inconisstent")

jichangjichang (approved):
lgtm

vin-huang:
gfx941 ci PASSED
http://math-ci.amd.com/blue/organizations/jenkins/gfx94x%2Fprecheckin%2FhipBLASLt/detail/PR-519/4/pipeline

PR #527, opened 2024-01-02
fix launch kernel return status

wenchuanchen:
https://ontrack-internal.amd.com/browse/SWDEV-439162

### DIFF
--- a/library/src/amd_detail/rocblaslt/src/tensile_host.cpp
+++ b/library/src/amd_detail/rocblaslt/src/tensile_host.cpp
@@ -262,6 +262,21 @@ namespace
                    : typeB;
     }
 
+    rocblaslt_status hip2RocStatus(hipError_t status)
+    {
+        switch(status)
+        {
+        case hipSuccess:
+            return rocblaslt_status_success;
+        case hipErrorUnknown:
+        case hipErrorRuntimeOther:
+        case hipErrorInvalidDevice:
+            return rocblaslt_status_internal_error;
+        default:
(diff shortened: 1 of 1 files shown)


library/src/amd_detail/rocblaslt/src/tensile_host.cpp
```diff
@@ -262,6 +262,21 @@ namespace
...
     }
 
+    rocblaslt_status hip2RocStatus(hipError_t status)
+    {
+        switch(status)
+        {
+        case hipSuccess:
+            return rocblaslt_status_success;
+        case hipErrorUnknown:
+            return rocblaslt_status_internal_error;
```
jichangjichang: you can remove this line if both of these return same error

library/src/amd_detail/rocblaslt/src/tensile_host.cpp
```diff
@@ -262,6 +262,20 @@ namespace
...
     }
 
+    rocblaslt_status hip2RocStatus(hipError_t status)
+    {
+        switch(status)
+        {
+        case hipSuccess:
+            return rocblaslt_status_success;
+        case hipErrorUnknown:
+        case hipErrorRuntimeOther:
```
jichangjichang, line 272: The error code of the ticket is hipErrorInvalidDevice. Sould we return rocblaslt_status_internal_error but rather than rocblaslt_status_not_implemented?

jichangjichang (approved):
LGTM

aazz44ss (approved):
LGTM

PR #530, opened 2024-01-04
pick 1 hotifx into 6.1 staging, revert the PR 519

vin-huang:
this condition will reject too many solutions. so revert this change.


vin-huang:
please ignore this PR

PR #533, opened 2024-01-05
[TensileLite] Fix the incorrect rejection of lbspp

vin-huang:
Simulated the local write instruction, calculated the local write address and offset to validate that the pad can be added correctly when lbspp is not zero.


KKyang (approved):

vin-huang:
The ci failed at gfx94x is another issue related to CK, which is not caused by this PR.

aazz44ss (approved):

PR #532, opened 2024-01-04
lazy load adapters

jeffdaily:
Fixes SWDEV-435811.

PR #385, opened 2023-10-16
Use lazy init for ext op libs

Serge45:
## Brief ##
Use lazy initialization for ext op library to avoid devices visibility conflicts among hipBLASLt and other libraries caused by environment variables.

PR #536, opened 2024-01-08
lazy load adapaters, thread safe

jeffdaily:
Follow-up to PR #385 and #532.

### DIFF
--- a/library/src/amd_detail/hipblaslt-ext-op.cpp
+++ b/library/src/amd_detail/hipblaslt-ext-op.cpp
@@ -145,35 +145,35 @@ namespace {
         return lib;
     }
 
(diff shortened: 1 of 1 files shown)


jeffdaily:
This PR starts from recently merged #385 but adapts the hotfix from rocm-rel-6.0 branch to make it thread safe.  I found this PR diff easier to review in github if I turned off whitespace changes.

Also, made the changes without attempting a build, so this does need a CI signal before merging.

Serge45 (approved):

jeffdaily:
This really needs to be tested on MI100 and some navi system before landing.  So let's not merge until this change can be staged appropriately.

jichangjichang:
close old PR

PR #545, opened 2024-01-10
Add UseUniversalArgs

KKyang:


jichangjichang:
please seperate it as 2 commits, too many custom_kernel would cause github Oops

KKyang:
> please seperate it as 2 commits, too many custom_kernel would cause github Oops

Seperating it into two commits isn't recommended because resetting to any commit only won't work.

KKyang:
Oh GitHub has squash

aazz44ss (approved):

PR #562, opened 2024-01-16
Used the placeholder to insert the branch instruction.

vin-huang:
Used the placeholder and calculated the length between the branch inst and target label using the long branch if necessary.


vin-huang:
the changes had been redesigned and added to the #566

PR #565, opened 2024-01-17
Update logic yaml

hcman2:
Update logic yaml of LSU4.


KKyang (approved):

Jinp800125 (approved):

jichangjichang (changes requested):
gfx941 also need same kernels

hcman2:
> gfx941 also need same kernels

updated

jichangjichang (approved):

hcman2:
need to check the result again due to performance change on server.

PR #584, opened 2024-01-24
[Tensilelite] Enable Silu

AndySu12:
Silu(x) = x * Sigmoid(x)

### DIFF
--- a/tensilelite/Tensile/Activation.py
+++ b/tensilelite/Tensile/Activation.py
@@ -140,6 +140,7 @@ class ActivationType:
                           ('tanh',        ActivationTypeRegister('tanh', False, 2,        True,  True, False,   False, False, False, False)), \
                           ('dgelu',       ActivationTypeRegister('dgelu', True, 0,       False,  True, False,   False, False, False, False)), \
                           ('geluscaling', ActivationTypeRegister('geluscaling', False, 1, True,  True, False,   False, False, False, False)), \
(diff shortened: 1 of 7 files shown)


KKyang (changes requested):
1. Please use rebase instead of merge.
2. The two commits have the same message, please fix it.

AndySu12:
@KKyang Done, thanks.

tensilelite/Tensile/Activation.py
```diff
@@ -765,6 +768,44 @@ def getDGeluModule(self, cDataType, vgprIn, vgprOut):
             raise RuntimeError("Unsupported data type %s."%cDataType.toDevice("HIP"))
         return module
 
+    def getSiluModule(self, cDataType, vgprIn, vgprOut):
```
KKyang, line 771: we already have sigmoid module, any reasons why can't call sigmoid here then add a mul x?
def getSiluModule(...):
    module.addModuleAsFlatItem(self.getSigmoidModule(...))
    module.add(VMulF32(...))

KKyang (changes requested):

tensilelite/Tensile/Activation.py
AndySu12: Good advice, I will use Sigmoid + mul. Initially I just use dedicated function, no further reason or consideration.

AndySu12: Done.

KKyang (approved):

jichangjichang (approved):
lgtm

PR #601, opened 2024-01-30
Change cmake version to 3.25.2 in install.sh

KKyang:
Can't compile on Ubuntu 20.04 because 3.16.8 is too old for ext module.


jichangjichang (approved):
lgtm

hcman2 (approved):

PR #623, opened 2024-02-13
cmake: add hipblas to public include dirs

evetsso:
hipblaslt.h directly includes hipblas.h.  But the CMake files do not declare that the hipblaslt header needs to have the hipblas header.

If hipblas is installed to a prefix other than /opt/rocm, the build fails even when `-DCMAKE_PREFIX_PATH=/path/to/that/prefix` is supplied:

```
In file included from /home/steleung/hipBLASLt/library/src/amd_detail/hipblaslt-ext.cpp:27:
In file included from /home/steleung/hipBLASLt/library/include/hipblaslt-ext.hpp:41:
/home/steleung/hipBLASLt/build/include/hipblaslt/hipblaslt.h:49:10: fatal error: 'hipblas/hipblas.h' file not found
   49 | #include <hipblas/hipblas.h>
      |          ^~~~~~~~~~~~~~~~~~~
1 error generated when compiling for gfx90a.
```

Fix this by telling CMake about this public include directory requirement.

This probably hasn't been noticed because hipcc automatically adds /opt/rocm to the include paths.

### DIFF
--- a/library/CMakeLists.txt
+++ b/library/CMakeLists.txt
@@ -141,6 +141,7 @@ target_include_directories(hipblaslt
                            PUBLIC  $<BUILD_INTERFACE:${CMAKE_SOURCE_DIR}/library/include>
                                    $<BUILD_INTERFACE:${PROJECT_BINARY_DIR}/include>
                                    $<BUILD_INTERFACE:${HIP_INCLUDE_DIRS}>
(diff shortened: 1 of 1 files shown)


jichangjichang (approved):
lgtm

KKyang (approved):

PR #624, opened 2024-02-15
Update NN/NT/TN F32 Performance

ssuyuanchang:


hcman2:
Do we have this kind of test yaml in Tensile/Tests? If not , should we add one?

jichangjichang (approved):
lgtm

hcman2 (approved):

ssuyuanchang:
> Do we have this kind of test yaml in Tensile/Tests? If not , should we add one?

This tuning patch is for demo; no need to add tests for now. I will add in the future if needed.

PR #625, opened 2024-02-15
Add a custom kernel.

hcman2:
Add a custom kernel for better performance.


jichangjichang:
It seems that we also need same kernel for gfx941 for stable branch

PR #629, opened 2024-02-15
add nt flag for gfx94x to develop

aazz44ss:


KKyang (approved):

tensilelite/Tensile/KernelWriterAssembly.py
```diff
@@ -5284,7 +5284,8 @@ def globalReadGuardKBody(tP):
       loadWidth = tP["globalReadInstruction"].totalWidth
 
       isGlc = True if tP["NonTemporal"]%2==1 else False
-      isSlc = True if tP["NonTemporal"]//2==1 else False
+      isSlc = True if (tP["NonTemporal"]//2)%2==1 else False
```
jichangjichang: Just a suggestion, better to use bitwise operation. ex:  isSlc = True if tP["NonTemporal"] & 0x2 else False

aazz44ss: good idea! done

jichangjichang (approved):
lgtm

PR #633, opened 2024-02-16
Add gfx941/942 NN HHS yamls for large m size

AndySu12:


jichangjichang:
Please attach issue link if have

AndySu12:
Data type is incorrect, abandon PR.

PR #651, opened 2024-02-22
Add icache flush to bench

KKyang:


hcman2 (approved):

jichangjichang (approved):
lgtm

randyh62 (commented):
Looks good to me.

clients/include/testing_matmul.hpp
```diff
@@ -2171,6 +2193,39 @@ void testing_matmul(const Arguments& arg)
         int         number_cold_calls = arg.cold_iters;
         int         number_hot_calls  = arg.iters;
 
+        int    flush_iter      = 100000;
+        double flush_time_used = 0;
+        if(arg.flush)
+        {
+            for(int i = 0; i < flush_iter; i++)
+                hipLaunchKernelGGL(flush_icache, dim3(1024), dim3(1), 0, stream);
```
dcyca: Hardcoding the number of thread blocks to 1024 doesn't seem very robust.  We should probably query the number of CUs and account for occupancy.  Otherwise we are depending on CU scheduling behavior that I don't think is guaranteed.  Would hipLaunchCooperativeKernel() be better here?

PR #661, opened 2024-02-26
Support lagency useBias (type:bool)

vin-huang:
* transfer useBias to int when the input useBias type is bool
 * set default UseBiasEnum when running the logic yaml directly (step 4 of tunning yaml)
 * currect the value of BetaOnlyUseBias


KKyang (approved):

jichangjichang (changes requested):
I don't see the neccessary for this PR.
Please don't  reserver this backward compaility. All user should follow new definition of UseBias.

PR #662, opened 2024-02-26
Correct the value of  UseBiasEnum and BetaOnlyUseBias

vin-huang:
* set default UseBiasEnum when running the logic yaml directly (step 4 of tunning yaml)
 * correct the value of BetaOnlyUseBias


KKyang (approved):

jichangjichang (approved):
lgtm

PR #664, opened 2024-02-29
Fix matrix transform edge handling

Serge45:
## Brief ##
This PR fixes edge handling for `hipblasLtMatrixTransform` API.

## Tasks ##
 - Fixed incorrect edge handling for VW == 1 case
 - Added support to null scales
 - Amended unit test for matrix transform
    - allocate memory that aligned with page size for detection of out of bound access
    - simplify test cases for null A/B input.


KKyang (approved):

jichangjichang (approved):
lgtm

PR #667, opened 2024-03-01
add ck of wave specialization for F8H HHS

aazz44ss:


msujon-AMD:
It is actually a prototype version of wave-specialization kernel, not production ready yet. This approach has very high potential to get better performance. 
Known issues: no tailloop support, need special NLL for odd iteration count, etc.

PR #669, opened 2024-03-05
Add sequence client

KKyang:


hcman2 (approved):

vin-huang (approved):

library/src/include/auxiliary.hpp
```diff
@@ -170,6 +183,39 @@ constexpr hipblasComputeType_t string_to_hipblas_computetype(const std::string&
...
+    return computetytpe;
+}
+
+HIPBLASLT_EXPORT
+constexpr hipblasLtEpilogue_t string_to_epilogue_type(const std::string& value)
+{
+    return
+        value == "HIPBLASLT_EPILOGUE_BIAS" ? HIPBLASLT_EPILOGUE_BIAS  :
+        value == "HIPBLASLT_EPILOGUE_DEFAULT" || value == "" ? HIPBLASLT_EPILOGUE_DEFAULT :
+        static_cast<hipblasLtEpilogue_t>(0);
```
jichangjichang, line 215: Better to support all epilogue since this function is in library part but not client part.
once you add it, should consider it from library perspective but not client perspective

library/src/include/auxiliary.hpp
```diff
@@ -170,6 +183,49 @@ constexpr hipblasComputeType_t string_to_hipblas_computetype(const std::string&
...
+}
+
+HIPBLASLT_EXPORT
+constexpr hipblasLtEpilogue_t string_to_epilogue_type_assert(const std::string& value)
+{
+    auto epilogue = string_to_epilogue_type(value);
+    if(static_cast<int>(epilogue) == 0)
+    {
+        std::cout << "Currently supported types are HIPBLASLT_EPILOGUE_DEFAULT, HIPBLASLT_EPILOGUE_BIAS." << std::endl;
```
jichangjichang: revise the logging as well

jichangjichang (approved):
lgtm

library/src/include/auxiliary.hpp
KKyang: fixed

PR #674, opened 2024-03-08
gtest: Updated f8/b8 out test cases: some cases should use norm_check

solaslin:
resolve: https://ontrack-internal.amd.com/browse/SWDEV-449221

The test failures are not only seen in that ticket but also can be seen "**occasionally**" in our CIs
The observations are: (for output type is F8 or B8)
1. **If ScaleC/D is false** (scaleC, D is 0/false), then the test initialization will use value **"1"**, (same in client code and tensile). In this case, using unit_check is good since all initializations are int.
2. **If ScaleC/D is true** when output type if F8/B8, then the initialization will use **random_small [0.0~1.0]** value with fraction. In this case, using unit check is not good since the tolerance of F8 (0.125) or B8 (0.25) can not be ignored. We should use norm_check instead.
3. Separated test cases into two different groups: 
   * One with ScaleCD=0 --> value would be 1 --> **unit_check** is good. 
   * And the other one with ScaleCD=1 --> value would be rand[0.0~0.1] --> **norm_check** is good.

### DIFF
--- a/clients/include/norm.hpp
+++ b/clients/include/norm.hpp
@@ -324,5 +324,9 @@ bool norm_check(double norm_error)
         return norm_error < 0.01;
     if(std::is_same<T, hipblasLtInt32>{})
         return norm_error < 0.0001;
+    if(std::is_same<T, hipblaslt_f8_fnuz>{})
+       return norm_error < 0.125;
+    if(std::is_same<T, hipblaslt_bf8_fnuz>{})
+        return norm_error < 0.25;    
     return false;
 }
--- a/clients/include/norm.hpp
+++ b/clients/include/norm.hpp
@@ -325,8 +325,8 @@ bool norm_check(double norm_error)
     if(std::is_same<T, hipblasLtInt32>{})
         return norm_error < 0.0001;
     if(std::is_same<T, hipblaslt_f8_fnuz>{})
-       return norm_error < 0.125;
+        return norm_error < 0.125;
     if(std::is_same<T, hipblaslt_bf8_fnuz>{})
-        return norm_error < 0.25;    
+        return norm_error < 0.25;
     return false;
 }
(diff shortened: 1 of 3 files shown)


clients/include/norm.hpp
```diff
@@ -324,5 +324,9 @@ bool norm_check(double norm_error)
         return norm_error < 0.01;
     if(std::is_same<T, hipblasLtInt32>{})
         return norm_error < 0.0001;
+    if(std::is_same<T, hipblaslt_f8_fnuz>{})
+       return norm_error < 0.125;
+    if(std::is_same<T, hipblaslt_bf8_fnuz>{})
+        return norm_error < 0.25;    
     return false;
```
solaslin: Added the norm check for F8/B8 (same as what we do in tensile, AlmostEqual)

library/include/hipblaslt_float8.h
```diff
@@ -76,7 +76,7 @@ struct HIPBLASLT_EXPORT hipblaslt_f8_fnuz
     // default constructor
     HIP_HOST_DEVICE hipblaslt_f8_fnuz() = default;
 
-#if defined(__gfx940__)
+#if defined(__gfx940__) || defined(__gfx941__) || defined(__gfx942__)
     // device specific optimized F8 down-conversion code
```
solaslin, line 80: Seems not directly related to this PR, but just updated this together.

vin-huang (approved):

Serge45 (approved):

clients/include/norm.hpp
```diff
@@ -324,5 +324,9 @@ bool norm_check(double norm_error)
         return norm_error < 0.01;
     if(std::is_same<T, hipblasLtInt32>{})
         return norm_error < 0.0001;
+    if(std::is_same<T, hipblaslt_f8_fnuz>{})
+       return norm_error < 0.125;
```
vin-huang: please fix the code format here

solaslin: done, thanks for catching

PR #718, opened 2024-03-26
fix cvt after ds schedule error

TonyYHsieh:
fix cvt after ds schedule error


KKyang (approved):

hcman2 (approved):

hcman2 (approved):
ok but 94x ci is terminated before end?

PR #725, opened 2024-04-01
Refactor generator

KKyang:
1. Add allocTmpGprList
    Able to get separated sgprs in a single call.
2. Return Inst instead of index for defineSgpr


aazz44ss (approved):
LGTM

jichangjichang (approved):
lgtm

PR #744, opened 2024-04-22
[tensilelite] Update NumWarmups and MaxEnqueuesPerSync

AndySu12:
- NumWarmups: This factor will be restricted to 30ms if set, and also affect MaxEnqueuesPerSync.
- MaxEnqueuesPerSync: max solution enqueues to perform per synchronization. This factor will be restricted if NumWarmups touches 30ms.

### DIFF
--- a/tensilelite/Tensile/Source/client/main.cpp
+++ b/tensilelite/Tensile/Source/client/main.cpp
@@ -649,7 +649,7 @@ int main(int argc, const char* argv[])
                                 kernels.push_back(kernel);
                             }
 
-                            size_t       eventCount        = gpuTimer ? kernels[0].size() : 0;
+                            size_t eventCount = gpuTimer ? kernels[0].size() : 0;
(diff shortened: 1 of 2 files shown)


AndySu12:
Question: should warmup introduce GPU time like enqueue?

Serge45:
FYI, there is an [experimental implementation](https://github.com/Serge45/hipBLASLt/tree/feature/time-based-warmup) for GPU time-based warmup mechanism for tensilelite and hipblaslt-bench.

jichangjichang (changes requested):
- Should not directly limit the time as 30ms, which totally make NumWarmups malfunction
- preWarmup and postWarmup count cpu time only but not GPU execution time.
- If you make CPU wait GPU execution in every postWarmup() to get the GPU execution time, the GPU tasks would not be executed back to back. This would directly change warm up behavior.

jayhawk-commits:

PR #752, opened 2024-04-30
[do not merge yet]Amax optimization

TonyYHsieh:
Amax optimization


library/include/hipblaslt-ext-op.h
```diff
@@ -150,6 +150,12 @@ HIPBLASLT_EXPORT hipblasStatus_t hipblasltExtLayerNorm(hipDataType datatype,
  *  input 2-D tensor buffer. can't be nullptr.
  *
  *  @param[in]
+ *  workspace Amax tensor buffer (4k). can't be nullptr.
```
jichangjichang: Finding anchors and sections in /home/docs/checkouts/readthedocs.org/user_builds/advanced-micro-devices-hipblaslt/checkouts/752/library/include/hipblaslt-ext-op.h:138: error: argument 'workspace' of command @param is not found in the argument list of hipblasltExtAMax(const hipDataType datatype, const hipDataType outDatatype, void *output, void *input, void *workSpace, void *sync, uint32_t m, uint32_t n, hipStream_t stream) (warning treated as error, aborting now)
the documentation...

library/src/amd_detail/hipblaslt-ext-op.cpp
```diff
@@ -511,27 +536,34 @@ hipblasStatus_t hipblasltAMaxWithScaleRun(const hipDataType datatype,
 
     const auto kernelName = sol->name();
     err                   = adapter->initKernel(kernelName);
+    int workSize          = 131072;
```
jichangjichang, line 637: hardcode??

TonyYHsieh: yes

TonyYHsieh: best value when we do experiment.

jichangjichang (commented):
Please add new APIs for new cases. Old API should be kept for supporting backward compatibilities.

library/include/hipblaslt-ext-op.h
TonyYHsieh: update

TonyYHsieh:
> Please add new APIs for new cases. Old API should be kept for supporting backward compatibilities.

update

PR #758, opened 2024-05-03
Update RTD config to use Python 3.10 and latest rocm-docs-core

samjwu:
Update RTD config to use Python 3.10 and rocm-docs-core 

Summary of proposed changes:

    ReadtheDocs will use Python 3.10 instead of Python 3.8 to build docs
        To use newer packages (eg: Sphinx 7)
    Doc builds will use latest rocm-docs-core version


yhuiYH (approved):
Looks good.  Not sure I have approval.  May need hipBLASLt code-owners for help on approval.

PR #776, opened 2024-05-19
Added hipblaslt_ext::matmulIsTuned API

Serge45:
## Brief ##
Added `hipblaslt_ext::matmulIsTuned` API, it can be used to check a given GEMM problem is exactly tuned or not.

## Implementation ##
 - added `rocblaslt_matmul_is_tuned` bridge function.
 - added `MatchingTag` in `ContractionSolution` struct.
 - in `ExactLogicLibrary`, set `MatchingTag` of matched solution to `Euqal` if the predicate is `Predicates::Contraction::EqualityMatching`.
 - `hipblaslt_ext::matmulIsTuned` returns `1` if tag of the best solution is `Equal` else `0`.
 - An example was also added.


Serge45:
Tests are passed on Ubuntu.

KKyang (approved):

hcman2 (approved):

PR #803, opened 2024-06-06
Amax fast algorithm version 2

TonyYHsieh:
use buffer load to handle 8x size


solaslin:
Will use #804 , close this one

PR #804, opened 2024-06-06
Amax codegen: improved tail block

solaslin:
Another option of #803 

Using the behavior of returning 0 when buffer load is out-of-range, we still treat tail block as a non-tail block, but make it smaller.
(Aligned with minimum-load-size instead of an entire worksize-block)

### DIFF
--- a/tensilelite/Tensile/Ops/AMaxGenerator.py
+++ b/tensilelite/Tensile/Ops/AMaxGenerator.py
@@ -517,23 +517,41 @@ class AMaxKernelGenerator:
         mod.addComment0("sum_per_blocksize")
 
         label_sum_per_blocksize = ti.Label("sum_per_blocksize", 'sum_per_blocksize')
+        label_tail_adj_WS_end = ti.Label("tail_adj_WS_end", 'tail_adj_WS_end')
         label_loop = ti.Label("loop", 'loop')
         label_last_loop = ti.Label("last_loop", 'last_loop')
-        label_loop_end = ti.Label("loop_end", 'loop_end')
         label_sum_per_blocksize_end = ti.Label("sum_per_blocksize_end", 'sum_per_blocksize_end')
 
         mod.add(ti.SLShiftRightB32(ti.sgpr("MainLoop"), ti.sgpr("LogWorkSize"), ti.sgpr("SizeLength")))
         mod.add(label_sum_per_blocksize)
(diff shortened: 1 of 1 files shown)


TonyYHsieh (approved):
LGTM

tensilelite/Tensile/Ops/AMaxGenerator.py
```diff
@@ -517,23 +517,41 @@ def sum_per_blocksize(self) -> ti.Module:
...
         mod.add(ti.SLShiftRightB32(ti.sgpr("MainLoop"), ti.sgpr("LogWorkSize"), ti.sgpr("SizeLength")))
         mod.add(label_sum_per_blocksize)
-        mod.add(ti.SCmpGeI32(ti.sgpr("WGIdx"), ti.sgpr("MainLoop")))
+        mod.add(ti.SCmpGtI32(ti.sgpr("WGIdx"), ti.sgpr("MainLoop")))
         mod.add(ti.SCBranchSCC1(label_sum_per_blocksize_end.getLabelName()))
         mod.addSpaceLine()
+        mod.add(ti.SCmpEQI32(ti.sgpr("WGIdx"), ti.sgpr("MainLoop")))
+        mod.add(ti.SCBranchSCC0(label_tail_adj_WS_end.getLabelName()))
+        mod.addSpaceLine()
+
```
solaslin, line 533: WGIdx == MainLoop -> might be a tail block

tensilelite/Tensile/Ops/AMaxGenerator.py
```diff
@@ -517,23 +517,41 @@ def sum_per_blocksize(self) -> ti.Module:
...
+        # WGIdx == MainLoop, last one, might have a tail block or nothing left
+        mod.addComment0("Last WG, check if this might be a tail block")
+        mod.add(ti.SLShiftLeftB32(ti.sgpr("Tmp"), ti.sgpr("LogWorkSize"), ti.sgpr("MainLoop"), "temp = non-tail elems = loop time * worksize"))
+        mod.add(ti.SCmpEQI32(ti.sgpr("Tmp"), ti.sgpr("SizeLength"), "if non-tail-elem == full-size --> no tail block, jump to reduction"))
+        mod.add(ti.SCBranchSCC1(self.label_reduction.getLabelName()))
+        mod.addSpaceLine()
```
solaslin, line 539: but if the sizeLength result in no tail, then we go to reduction directly, else we do tail block.

PR #851, opened 2024-06-20
Fix errors when calculating ldsPad and ldsNumBytes of sparse metadata.

vin-huang:
Moved the related calculation into the calcLdsPad() and calcLdsNumBytes()


vin-huang:
the test of gfx942 is passed at ubuntu22

aazz44ss (approved):

hcman2 (approved):

PR #859, opened 2024-06-22
gfx942 80cu BBS/HHS TN GridBased additional tuning

nakajee:


nakajee:
- Tuned kernels are added to gfx942 80cu GridBased logic (HHS/BBS)
    - 63 sizes tuned: 21 new sizes, 42 exising sizes
- removed 43 sizes from base gfx942_80cu BBS TN GridBased logic file (same change for HHS TN)
- Added/updated 1 size [8192, 8192, 1, 8192] to HHS/BBS Equality logic
    - looks like the size in gfx942 Equality (not 80cu) can be used if 80cu Equality does not include the size
    - in that case, we need to add the size in gfx942 80cu Equality logic file

babakpst (approved):

KKyang (approved):

PR #864, opened 2024-06-25
fix incorrect calcLdsNumBytes calculation

aazz44ss:
fix incorrect ldsPad value in calcLdsNumBytes

### DIFF
--- a/tensilelite/Tensile/SolutionStructs.py
+++ b/tensilelite/Tensile/SolutionStructs.py
@@ -2389,7 +2389,7 @@ class Solution(collections.abc.Mapping):
         ldsNumBytesA = state["_DepthUA"] * (state["MacroTileA"] + ldsPadA) * bpeA
       padInterval = LdsBlockSizePerPadA
       if padInterval != 0:
-        ldsNumBytesA = int((state["_DepthUA"] * state["MacroTileA"] * bpeA) / padInterval * (padInterval + state["LdsPadA"] * bpeA))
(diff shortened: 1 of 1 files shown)


KKyang (approved):

hcman2:
CI Failed.
mci/precheckin/test/gfx90a(main) — Failed in Test stage on gfx90a

aazz44ss:
> CI Failed. mci/precheckin/test/gfx90a(main) — Failed in Test stage on gfx90a

Re-run CI again, it passed.

hcman2 (approved):
OK if gfx94x ci also pass.

hcman2:
Should we add a test yaml to cover this bug?

aazz44ss:
> Should we add a test yaml to cover this bug?

I just re-ran CI and it passed. Not sure what happened.

hcman2:
Not sure which bug this change fixed? if we have bug before this change, the test yaml in this codebase cannot hit the bug this change fixed. That is why I ask for a test yaml to hit this bug.

aazz44ss:
> Not sure which bug this change fixed? if we have bug before this change, the test yaml in this codebase cannot hit the bug this change fixed. That is why I ask for a test yaml to hit this bug.

OK, I misunderstood I thought you were saying CI failed bug, that's what I an not sure, I re-ran CI and it passed, I also check CI on my local gfx942 and passed.

This PR fix wrong value used in the function, not able to reproduce after fix.

PR #872, opened 2024-06-28
Port negative WGM from Tensile

KKyang:
https://github.com/ROCm/Tensile/commit/d76408a4a939ebc8187145162247b333e9ee47a9

### DIFF
--- a/tensilelite/Tensile/KernelWriterAssembly.py
+++ b/tensilelite/Tensile/KernelWriterAssembly.py
@@ -1143,7 +1143,10 @@ class KernelWriterAssembly(KernelWriter):
       commonArgs.add(self.argLoader.loadKernArg(sgprNumsOfGemm, "KernArgAddress", hex(0), dword=1))
       commonArgs.addComment1("Load GSU data")
       commonArgs.add(self.argLoader.loadKernArg("GSU", "KernArgAddress", hex(4), dword=1))
-      loadedCommonArgs = 8 # TODO: Need to make this automatically calculated
+      commonArgs.addComment1("Load WGM data")
+      commonArgs.add(self.argLoader.loadKernArg("WGM", "KernArgAddress", hex(8), dword=1))
+      self.commonArgNum = 3 # TODO: Need to make this automatically calculated
+      self.loadedCommonArgs = self.commonArgNum * 4
       ########################################
       # kernel args parameters
       load = self.states.numSgprToLoad
@@ -1155,9 +1158,9 @@ class KernelWriterAssembly(KernelWriter):
       if (self.states.kernel["WorkGroupMappingXCC"] > 1):
         tmpSgprNumWorkGroups = self.sgprPool.checkOut(1, preventOverflow=0)
         hbmArgs.addComment1("Grouped Gemm: Load num of WGs")
-        hbmArgs.add(self.argLoader.loadKernArg(tmpSgprNumWorkGroups, "KernArgAddress", hex(16), dword=1))
+        hbmArgs.add(self.argLoader.loadKernArg(tmpSgprNumWorkGroups, "KernArgAddress", hex(self.loadedCommonArgs+8), dword=1))
       hbmArgs.addComment1("Load address of kernel arguments")
-      hbmArgs.add(self.argLoader.loadKernArg("KernArgAddress", "KernArgAddress", hex(8), dword=2))
+      hbmArgs.add(self.argLoader.loadKernArg("KernArgAddress", "KernArgAddress", hex(self.loadedCommonArgs), dword=2))
 
       moduleArgs.addModuleAsFlatItems(fastdeepcopy(commonArgs))
       moduleArgs.add(SWaitCnt(0))
@@ -1171,15 +1174,15 @@ class KernelWriterAssembly(KernelWriter):
           moduleArgs.add(SCmpEQU32(src0=sgpr(sgprArgType), src1=2, comment="ArgType == 2 ?"))
           moduleArgs.add(SCBranchSCC0(labelName=extReadEpilogueLabeltmp.getLabelName()))
         moduleArgs.addComment1("Grouped Gemm: Load address of external kernel arguments")
-        moduleArgs.add(self.argLoader.loadKernArg("AddressTD", "KernArgAddress", hex(28), dword=2))
-        moduleArgs.add(self.argLoader.loadKernArg("Synchronizer", "KernArgAddress", hex(20), dword=2))
+        moduleArgs.add(self.argLoader.loadKernArg("AddressTD", "KernArgAddress", hex(self.loadedCommonArgs+20), dword=2))
+        moduleArgs.add(self.argLoader.loadKernArg("Synchronizer", "KernArgAddress", hex(self.loadedCommonArgs+12), dword=2))
         moduleArgs.add(extReadEpilogueLabeltmp)
 
       moduleArgs.add(SCmpEQU32(src0=sgpr(sgprArgType), src1=(0), comment="Is kernel args"))
       labelHBM = Label("HBMArgs", comment="")
       labelLoadEnd = Label("LoadArgsEnd", comment="")
       moduleArgs.add(SCBranchSCC0(labelName=labelHBM.getLabelName()))
-      moduleArgs.add(SAddU32(dst=sgpr("KernArgAddress"), src0=sgpr("KernArgAddress"), src1=hex(loadedCommonArgs), comment="Shift common args"))
+      moduleArgs.add(SAddU32(dst=sgpr("KernArgAddress"), src0=sgpr("KernArgAddress"), src1=hex(self.loadedCommonArgs), comment="Shift common args"))
       moduleArgs.add(SAddCU32(dst=sgpr("KernArgAddress+1"), src0=sgpr("KernArgAddress+1"), src1=hex(0)))
       moduleArgs.addModuleAsFlatItems(self.getKernelArgLoadModule(kernel, sgprStart, load, 0))
       if self.states.numSgprPreload > 0:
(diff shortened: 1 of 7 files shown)


tensilelite/Tensile/KernelWriterAssembly.py
```diff
@@ -1916,40 +1918,64 @@ def graWorkGroup(self, kernel):
     ########################################
     # Blocked rows or columns
     # Do branch
-    wgmLabel = Label(label=self.labels.getNameInc("WGM"), comment="")
-    module.add(SCmpLeU32(src0=sgpr("WGM"), src1=1, comment="WGM <= 1 ?"))
-    module.add(SCBranchSCC1(labelName=wgmLabel.getLabelName(), comment="branch if WGM <= 1"))
+    wgmLabel         = Label(label=self.labels.getNameInc("WGM"), comment="")
+    wgmLabelNegative = Label(label=self.labels.getNameInc("WGMNegative"), comment="")
+    module.add(SCmpLtI32(src0=sgpr("WGM"), src1=1, comment="WGM < 1 ?"))
```
nakajee: Shouldn't this be WGM < 0?

tensilelite/Tensile/KernelWriterAssembly.py
```diff
@@ -1916,40 +1918,63 @@ def graWorkGroup(self, kernel):
     ########################################
     # Blocked rows or columns
     # Do branch
-    wgmLabel = Label(label=self.labels.getNameInc("WGM"), comment="")
-    module.add(SCmpLeU32(src0=sgpr("WGM"), src1=1, comment="WGM <= 1 ?"))
-    module.add(SCBranchSCC1(labelName=wgmLabel.getLabelName(), comment="branch if WGM <= 1"))
+    wgmLabel         = Label(label=self.labels.getNameInc("WGM"), comment="")
+    wgmLabelNegative = Label(label=self.labels.getNameInc("WGMNegative"), comment="")
+    module.add(SCmpLeI32(src0=sgpr("WGM"), src1=1, comment="WGM <= 1 ?"))
+    module.add(SCBranchSCC1(labelName=wgmLabelNegative.getLabelName(), comment="branch if WGM <= 1"))
```
nakajee: To avoid jumping twice in WGM= 0 or 1 case, we can put negative version first.
Then, we can do like this.
if (WGM > 1)
  jump to positive code
Negative code
    here, we convert WGM to positive value
    if WGM<=1
        Jump to the end
Positive:
Positive code

tensilelite/Tensile/KernelWriterAssembly.py
```diff
@@ -1916,40 +1918,63 @@ def graWorkGroup(self, kernel):
...
+          numWorkgroupsSecond = "NumWorkGroups1"
+
+        if wgmType:
+          module.add(wgmLabelNegative)
+          module.add(SCmpGeI32(src0=sgpr("WGM"), src1=-1, comment="WGM >= -1 ?"))
+          module.add(SCBranchSCC1(labelName=wgmLabel.getLabelName(), comment="branch if WGM >= -1"))
+          module.add(SAbsI32(dst=sgpr("WGM"), src=sgpr("WGM"), comment="abs(WGM)"))
+          module.add(Label("WGMTMP", comment=""))
+        # note this overwrites blockId2+1
+        module.add(scalarUInt32DivideAndRemainder(qReg=blockId2, dReg=workgroupSecond, divReg="WGM", rReg=wgSerial2, tmpVgprRes=tmpVgprRes, wavewidth=kernel["WavefrontSize"], doRemainder=False, comment="WGM"))
```
nakajee, line 1951: Ideally, we would not like to generate WGM calculation for both positive and negative.
We can unify the code by using some tmp sgpr for first and second.
I think we can overwrite WorkGroup0 and 1 here.
So, what we can do is
   Swap WorkGroup0 and 1 in negative case
   allocate 2 sgpr for numWorkgroupsFirst and second. Store the values properly for positive and negative.

tensilelite/Tensile/KernelWriterAssembly.py
nakajee: I understand the logic.
I put another comment below.

tensilelite/Tensile/Source/lib/source/ContractionSolution.cpp
```diff
@@ -796,15 +793,21 @@ namespace Tensile
         }
 
         args.template append<uint32_t>("internalArgs", internalArg0);
+
+        int32_t internalArg1 = 0;
+        if(internalArgsSupport.wgm)
+        {
+            if(wgm == -1)
+                wgm = 1;
+            args.template append<int32_t>("internalArgs1", wgm);
```
nakajee, line 814: I am not sure if we really need -1024<= WGM <= 1024.
It might be OK to limit the range of WGM from -128 to 127 to fit 8bit range.
we can use S_SEXT_I32_I8 to convert signed int8 to 32bit.

nakajee: But, I think it is a good idea to expand the range of GSU to 16bit.
GSU>256 can be needed.
I think GSU range change is important and better to mention in your commit comment.

KKyang: I think we can keep this because we don't have any bits left for internalArgs0

tensilelite/Tensile/KernelWriterAssembly.py
KKyang: I'm not sure if we have enough sgpr. I have to check.

KKyang: Currently some kernels use too much sgprs, so still not able to allocate 2 tmp sgprs.
I added a TODO.

tensilelite/Tensile/KernelWriterAssembly.py
KKyang: Fixed

tensilelite/Tensile/KernelWriterAssembly.py
KKyang: Maybe we can swap  numWorkgroups0 and  numWorkgroups1 and swap back?

nakajee: I was thinking the same way.
I think we can technically do that (the code might look a bit complicated).
We can use xor operation to swap without tmp register.

KKyang: > I was thinking the same way. I think we can technically do that (the code might look a bit complicated). We can use xor operation to swap without tmp register.

But store requires a sgpr to cache the state.

KKyang: We do the abs(WGM) so we can't know if it's negative or not.

nakajee: OK. I do not come up with any other ideas.
We can keep current code for now....

KKyang: Just thinking if we use the same code, we'll need at least 9 lines to exchange data for negative cases.

aazz44ss:
grouped gemm failed to run with WGM<0

terminate called after throwing an instance of 'std::runtime_error'
  what():  Error in /root/hipBLASLt/tensilelite/Tensile/Source/lib/source/ContractionSolution.cpp:964: sizeMapping.workGroupMapping >= 0

tensilelite/Tensile/Source/lib/source/ContractionSolution.cpp
```diff
@@ -775,36 +775,46 @@ namespace Tensile
...
         }
 
         args.template append<uint32_t>("internalArgs", internalArg0);
+
+        int32_t internalArg1 = 0;
+        if(internalArgsSupport.wgm && sizeMapping.customKernelName == "")
+        {
+            if(wgm == -1)
+                wgm = 1;
+            args.template append<int32_t>("internalArgs1", wgm);
```
nakajee, line 814: I think we need to add wgm even if wgm is 0.
Kernel code read this regardless of wgm value.

tensilelite/Tensile/Source/lib/source/ContractionSolution.cpp
```diff
@@ -775,36 +775,46 @@ namespace Tensile
...
-
-        const uint32_t mask         = 0xFF;
-        uint32_t       internalArg0 = mask & gsu;
-        uint32_t       wgShift8     = (mask & wgm) << 8;
-        internalArg0                = internalArg0 | wgShift8;
+        int32_t wgm  = param.wgm() != 0 ? param.wgm() : sizeMapping.workGroupMapping;
+
+        const uint32_t mask16       = 0xFFFF;
+        const uint32_t mask8        = 0xFF;
+        uint32_t       internalArg0 = mask16 & gsu;
```
nakajee: Strictly speaking, this should be mask8 in customKernel case.
(it might not be an issue unless we use GSU>255)

tensilelite/Tensile/Source/lib/source/ContractionSolution.cpp
KKyang: Wgm=0 will still write. The flag is use to check whether wgm is a kernel argument or not.

tensilelite/Tensile/Source/lib/source/ContractionSolution.cpp
KKyang: Will change

KKyang:
> grouped gemm failed to run with WGM<0
> 
> terminate called after throwing an instance of 'std::runtime_error' what(): Error in /root/hipBLASLt/tensilelite/Tensile/Source/lib/source/ContractionSolution.cpp:964: sizeMapping.workGroupMapping >= 0

fixed

tensilelite/Tensile/KernelWriterAssembly.py
```diff
@@ -1916,40 +1918,64 @@ def graWorkGroup(self, kernel):
...
+        # WorkGroup1 = wgSerial2 - (wgmDivisor * WorkGroup0)
+        tmpVgpr = self.vgprPool.checkOut(2, "div")
+        tmpVgprRes = RegisterPoolResource(idx=tmpVgpr, size=2)
+        module.add(scalarUInt32DivideAndRemainder(qReg=workgroupFirst, dReg=wgSerial2, divReg=wgmDivisor, rReg=workgroupSecond, tmpVgprRes=tmpVgprRes, wavewidth=kernel["WavefrontSize"]))
+        self.vgprPool.checkIn(tmpVgpr)
+        module.add(SMulI32(dst=sgpr(workgroupSecond), src0=sgpr(workgroupFirst), src1=sgpr(wgmDivisor), comment="quotient * non-magic divisor"))
+        module.add(SSubU32(dst=sgpr(workgroupSecond), src0=sgpr(wgSerial2), src1=sgpr(workgroupSecond), comment="%s=remainder"%workgroupSecond))
+        module.add(SMulI32(dst=sgpr(blockId2), src0=sgpr(blockId2), src1=sgpr("WGM"), comment="blockId * WGM"))
+        module.add(SAddU32(dst=sgpr(workgroupSecond), src0=sgpr(workgroupSecond), src1=sgpr(blockId2), comment="wg1 += blockId * WGM"))
+        module.add(SBranch(wgmLabel.getLabelName()))
```
nakajee: We can remove unnecessary branch by adding "if wgmType:" here.

KKyang: Done

tensilelite/Tensile/KernelWriterAssembly.py
```diff
@@ -1916,40 +1918,65 @@ def graWorkGroup(self, kernel):
...
+        module.add(SCmpEQU32(src0=sgpr(wgmDivisorMagicNumber), src1=0, comment="remainder == 0 ?"))
+        module.add(SCMovB32(dst=sgpr(wgmDivisorMagicNumber), src=sgpr("WGM"), comment="remainder = WGM if remainder == 0"))
+
+        module.add(SCmpGeU32(src0=sgpr(blockId2), src1=sgpr(wgmDivisor), comment="blockId >= numFullBlocks ?"))
+        module.add(SCSelectB32(dst=sgpr(wgmDivisor), src0=sgpr(wgmDivisorMagicNumber), src1=sgpr("WGM")))
+
+        # For WGM >= 1
+        # WorkGroup0 = wgSerial2 / wgmDivisor
+        # WorkGroup1 = wgSerial2 - (wgmDivisor * WorkGroup0)
+        tmpVgpr = self.vgprPool.checkOut(2, "div")
```
nakajee: This is not caused by your change, but I think tmpVgpr allocation should be before line 1952.
tmpVgpr is allocated at line 1902 and checked in at 1905.
However, it is used again at line 1952 before reallocating tmpVgpr.

nakajee (approved):
Thanks for your update.
I do not have any further comment.
Please make sure all tests pass (excluding known issues).

aazz44ss (approved):
LGTM

KKyang:
gfx942 local test passed

PR #887, opened 2024-07-08
Add KernArgsVersion

KKyang:
Add KernArgsVersion

Since there are too many custom kernels in TensileLite, we add
KernArgsVersion to check which custom kernel goes to which kernel
arguments.

The following config is needed else the generator will raise an
error. You can find the version inder "Optimizations and Config:".

```
/******************************************/
/* Optimizations and Config:              */
/******************************************/
/* Version: 1 */ <-- Should be here
.....
.amdgpu_metadata
```

If your generated kernel does not have version shown, use 0.

```
custom.config:
   InternalSupportParams:
      KernArgsVersion: 0
```

### DIFF
--- a/tensilelite/Tensile/Common.py
+++ b/tensilelite/Tensile/Common.py
@@ -314,6 +314,7 @@ internalParameters = {
 
 # These parameters are used in ContractionSolutions for user arguments support.
 defaultInternalSupportParams = {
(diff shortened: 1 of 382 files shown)


Jinp800125 (approved):

aazz44ss (approved):
LGTM

PR #897, opened 2024-07-12
support scale alpha vector on n-dim

vin-huang:
-  Redefined UseScaleAlphaVec

    0 - Alpha Vector Scaling is disabled
    1 - Calculated Alpha Vector Scaling along M-dim
    2 - Calculated Alpha Vector Scaling along N-dim
    3 - Calculated Alpha Vector Scaling along either M-dim or N-dim.
 
-  if UseScaleAlphaVec does not equal 0, then the setting of UseScaleAlphaVec must be the same as UseBias, and vice versa.

    This means when UseScaleAlphaVec = 3, then UseBias can only be 0 or 3.
    
-  Rename kernel argument BiasDim to FactorDim. which is used to control alpha vector scaling and bias should be calculated along either M-dim or N-dim. (only use this kernel argument when one of UseScaleAlphaVec or UseBias is 3)

   0 - along M-dim
   1 - along N-dim

-    Replaced keyword BiasDim by FactorDim.


KKyang (approved):

jichangjichang (approved):
lgtm

Jinp800125 (approved):

PR #898, opened 2024-07-12
Fix Lazy Lib Load to Improve Library Loading Time

wenchuanchen:
This PR is from https://ontrack-internal.amd.com/browse/SWDEV-465673. The long library loading time is fixed by removing deviceSet from LoadLibraryFilePreload. However, this introduces a seg fault in getSolutionsFromIndex and isSolutionSupported due to empty library->solutions. Since getSolutionsFromIndex and isSolutionSupported are not the target of the ticket and aren't supported by rocblas as well, there is a workaround. getSolutionsFromIndex should be further designed to load corresponding placeholder library when specific index is passed in.

When using hipblaslt-bench, "--algo_method heuristic" benefits from this PR. "--algo_method all" and "--algo_method index" won't change.

### DIFF
--- a/library/src/amd_detail/rocblaslt/src/tensile_host.cpp
+++ b/library/src/amd_detail/rocblaslt/src/tensile_host.cpp
@@ -931,6 +931,8 @@ namespace
 #else
         std::shared_ptr<hipDeviceProp_t> m_deviceProp;
 #endif
+        std::string m_tensileLibPath;
+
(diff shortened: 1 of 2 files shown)


jichangjichang (approved):
lgtm
descript more on which hipblaslt-bench path can speed up on init time and which hipblaslt-bench cannot speed up.

KKyang (approved):

PR #901, opened 2024-07-14
fix mismatch issue with xfp32 + TLU + VectorWidth

nakajee:
- disabled wider local read for xfp32 + TLU
- added test cases

### DIFF
--- a/tensilelite/Tensile/KernelWriter.py
+++ b/tensilelite/Tensile/KernelWriter.py
@@ -2682,7 +2682,7 @@ class KernelWriter(metaclass=abc.ABCMeta):
 
     # force lrvwTile = 1 for numBytes >= 4 + MIInputPerThread > 1
     # TODO: implement extra logic to swap vgprs after local read to suport lrvwTile > 1 for umBytes >= 4 + MIInputPerThread > 1
-    forceLrvwTile1 = kernel["ProblemType"]["DataType"].numBytes() >= 4 and kernel["MIInputPerThread"] > 1
+    forceLrvwTile1 = kernel["ProblemType"]["DataType"].numBytes() >= 4 and (kernel["EnableMatrixInstruction"] and kernel["MIInputPerThread"] > 1)
     if not kernel["UnrollMajorLDSA"] and not forceLrvwTile1:
       self.states.lrvwTileA = kernel["VectorWidthA"]
     else:
--- a/tensilelite/Tensile/KernelWriter.py
+++ b/tensilelite/Tensile/KernelWriter.py
@@ -2680,12 +2680,14 @@ class KernelWriter(metaclass=abc.ABCMeta):
     if kernel["ProblemType"]["Sparse"] and not kernel["DirectToVgprSparseMetadata"]:
       vwm = kernel["GlobalReadVectorWidthMetadata"]
 
-    if not kernel["UnrollMajorLDSA"]:
+    # use lrwvTile = 1 for XFP32
+    isXFP32 = kernel["EnableF32XdlMathOp"] and kernel["ProblemType"]["F32XdlMathOp"].isXFloat32()
+    if not kernel["UnrollMajorLDSA"] and not isXFP32:
       self.states.lrvwTileA = kernel["VectorWidthA"]
     else:
       self.states.lrvwTileA = 1
 
-    if not kernel["UnrollMajorLDSB"]:
+    if not kernel["UnrollMajorLDSB"] and not isXFP32:
       self.states.lrvwTileB = kernel["VectorWidthB"]
     else:
       self.states.lrvwTileB = 1
--- a/tensilelite/Tensile/KernelWriter.py
+++ b/tensilelite/Tensile/KernelWriter.py
@@ -2680,14 +2680,15 @@ class KernelWriter(metaclass=abc.ABCMeta):
     if kernel["ProblemType"]["Sparse"] and not kernel["DirectToVgprSparseMetadata"]:
       vwm = kernel["GlobalReadVectorWidthMetadata"]
 
-    # use lrwvTile = 1 for XFP32
-    isXFP32 = kernel["EnableF32XdlMathOp"] and kernel["ProblemType"]["F32XdlMathOp"].isXFloat32()
-    if not kernel["UnrollMajorLDSA"] and not isXFP32:
+    # force lrvwTile = 1 for numBytes >= 4 + MIInputPerThread > 1
+    # TODO: implement extra logic to swap vgprs after local read to suport lrvwTile > 1 for umBytes >= 4 + MIInputPerThread > 1
+    forceLrvwTile1 = kernel["ProblemType"]["DataType"].numBytes() >= 4 and kernel["MIInputPerThread"] > 1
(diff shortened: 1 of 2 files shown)


AlexBrownAMD (approved):

nakajee:
More detail about what is happening
(1) VectorWidthA=1 case (correct result)
  vgprVALUA0 = ds_read_b32 (for K=0-3)
  vgprVALUA1 = ds_read_b32 (for K=4-7)
  v_mfma  vgprVALUA0:1,....
(2) VectorWidthA=2 case (incorrect result)
  vgprVALUA0:1 = ds_read_b64 (for K=0-3)
  vgprVALUA2:3 = ds_read_b64 (for K=4-7)
  v_mfma  vgprVALUA0:1,....   <-incorrect, we should combine VALUA0 and 2
We need extra code to swap VALUA1 and 2.
So far, I need this fix soon and did not have enough time to enable wider local read (=VectorWidth) for XFP32.

I just disabled wider local read and always use ds_read_b32 for XFP32 (TLU case).

jichangjichang:
This is workaround definitely. I would recommend apply "forcing the VectorWidth as 1" or "rejecting VectorWidth > 1 solution" in SolutionStruct.py but not in kernelWriter.py.

nakajee:
> This is workaround definitely. I would recommend apply "forcing the VectorWidth as 1" or "rejecting VectorWidth > 1 solution" in SolutionStruct.py but not in kernelWriter.py.

VectorWidth affects more than local read (such as StoreVectorWidth).
I would not like to disable other optimizations.

nakajee:
I do not see any benefit of always forcing VectorWidth to 1 compared to disabling wider local read.
Ideally, we should have a way to specify width of local read for TLU=True case.
(It should be LocalReadVectorWidth, but it currently works only for TLU=False)
So far, we can control it only with VectorWidth, but as I said, VectorWidth affects much more than local read.

jichangjichang:
> I do not see any benefit of always forcing VectorWidth to 1 compared to disabling wider local read. Ideally, we should have a way to specify width of local read for TLU=True case. (It should be LocalReadVectorWidth, but it currently works only for TLU=False) So far, we can control it only with VectorWidth, but as I said, VectorWidth affects much more than local read.

Got it

tensilelite/Tensile/KernelWriter.py
```diff
@@ -2680,12 +2680,14 @@ def initKernel(self, kernel, tensorParametersA, tensorParametersB):
     if kernel["ProblemType"]["Sparse"] and not kernel["DirectToVgprSparseMetadata"]:
       vwm = kernel["GlobalReadVectorWidthMetadata"]
 
-    if not kernel["UnrollMajorLDSA"]:
+    # use lrwvTile = 1 for XFP32
+    isXFP32 = kernel["EnableF32XdlMathOp"] and kernel["ProblemType"]["F32XdlMathOp"].isXFloat32()
+    if not kernel["UnrollMajorLDSA"] and not isXFP32:
```
jichangjichang: Could you simply check if data type is smaller than 4 bytes?  The issue is not only for XF32, but for datatype >= 4 bytes.

if not kernel["UnrollMajorLDSA"] and kernel["ProblemType"]["DataType"].numRegisters() < 1:
      self.states.lrvwTileA = kernel["VectorWidthA"]
 else:
      self.states.lrvwTileA = 1

nakajee: This issue is XFP32 specific.
It is because MatrixInstK is doubled for XFP32.
S,D,C,Z GEMM do not have same issue.

nakajee: As I explained in the above comment,

====

(2) VectorWidthA=2 case (incorrect result)
vgprVALUA0:1 = ds_read_b64 (for K=0-3)
vgprVALUA2:3 = ds_read_b64 (for K=4-7)
v_mfma vgprVALUA0:1,.... <-incorrect, we should combine VALUA0 and 2

====

This happens only with XFP32 because MatrixInstK is 8 (MI16x16x8x1).

tensilelite/Tensile/KernelWriter.py
```diff
@@ -2680,12 +2680,14 @@ def initKernel(self, kernel, tensorParametersA, tensorParametersB):
...
       self.states.lrvwTileA = kernel["VectorWidthA"]
     else:
       self.states.lrvwTileA = 1
 
-    if not kernel["UnrollMajorLDSB"]:
+    if not kernel["UnrollMajorLDSB"] and not isXFP32:
```
KKyang: Please add # FIXME for easy search.

tensilelite/Tensile/KernelWriter.py
KKyang: Please add # FIXME for easy search.

KKyang (changes requested):
Please add # FIXME for easy search.

tensilelite/Tensile/KernelWriter.py
nakajee: I feel FIXME is not exactly expressing this.
This specific case (XFP32+ TLU + VW) requires extra logic, but that logic is simply not implemented yet.
If you would like to put some comment, I can put TODO here.
like this,
   TODO: need to implement extra logic (swap vgprs) to support XFP32+ TLU + VW

Is this OK for you?

tensilelite/Tensile/KernelWriter.py
jichangjichang: > This happens only with XFP32 because MatrixInstK is 8 (MI16x16x8x1).

This is common code. You shouldn't assume that "MatrixInstK is 8" is for all platform.
The current un-supported condition for not UnrollMajorLDS is : data type >= 4 byte and MIInputPerThread is > 1

tensilelite/Tensile/KernelWriter.py
nakajee: It is not due to incorrect logic to fix, but the implementation itself is missing.

tensilelite/Tensile/KernelWriter.py
nakajee: OK. I will make another change as you suggest.

KKyang (approved):

tensilelite/Tensile/KernelWriter.py
nakajee: I pushed the code change as you suggested.
I confirmed xfp32 test (I added in this PR) passed with the new change.

tensilelite/Tensile/KernelWriter.py
nakajee: I added TODO comment.

jichangjichang (approved):
lgtm

nakajee:
TensileCreateLibrary failed because MIInputPerThread is not defined if EnableMatrixInstruction is false.
I added code to check EnableMatrixInstruction before referring MIInputPerThread.

nakajee:
I confirmed 
- TensileCreateLibrary failed with navi* before the fix
- TensileCreateLibrary passed with navi* after the fix

nakajee:
fp8.yaml failed but I can find the same fail in precheckin test with the latest develop branch.
Not related to my change.

PR #905, opened 2024-07-16
Add FP8 row-wise gemm (tensilelite)

KKyang:


tensilelite/Tensile/Common.py
```diff
@@ -1164,7 +1164,8 @@ def getArchitectureName(gfxName):
     "Gradient":                 False,            # =True set globalWriteElements to gradient mode
     "UseBias":                  0,                # =1 support bias vector on M direction, =2 support bias vector on N direction, =3 support bias vector on both M,N direction
     "BiasSrc":                  "D",              # This parameter is used in gradient + bias. Support A, B, D.
-    "UseScaleAB":               False,            # =True use scaleA, scaleB
+    "UseScaleAB":               False,            # =True use scaleA, scaleB, may be deprecated
```
jichangjichang: If ScaleAB and scaleABvec cannot exist at the same time, It is better use same problem argument but just extend its scope. Please don't create new "UseScaleABVec"
UseScaleAB: 0 # not support
UseScaleAB: 1 # legacy scaling factor
UseScaleAB: 2 # scale vector

KKyang: Scalar AB is not a legacy scaling factor. We are still figuring out a way to make them both exist in one kernel but not increase the kernel size too much.

tensilelite/Tensile/Tests/common/gemm/fp8.yaml
```diff
@@ -219,3 +221,62 @@ BenchmarkProblems:
...
+      HighPrecisionAccumulate: True
+      TransposeA: 1
+      TransposeB: 0
+      UseBeta: True
+      Batched: True
+      BiasDataTypeList: [s]
+      UseBias:       1
+      UseE:          False
+      Activation:    True
+      UseScaleABVec: True
```
jichangjichang: correct this one

jichangjichang (approved):
lgtm

vin-huang (approved):

PR #907, opened 2024-07-16
one AMax fix and ValueDividedbyAMaxWithRcp Gtest

TonyYHsieh:
one AMax fix and ValueDividedbyAMaxWithRcp Gtest
1. AMax fix conversion error when output is H type with value divided
2. ValueDividedbyAMaxWithRcp GTest

### DIFF
--- a/clients/gtest/hipblaslt_gtest_ext_op.cpp
+++ b/clients/gtest/hipblaslt_gtest_ext_op.cpp
@@ -83,13 +83,13 @@ namespace
     template <typename T>
     T abs(T a)
     {
-        return (a > 0) ? a : -a;
+        return (double(a) > 0) ? a : -a;
     }
 
     template <typename T>
     T max(T a, T b)
     {
-        return (a > b) ? a : b;
+        return (double(a) > double(b)) ? a : b;
     }
 
     template <typename Ti, typename To>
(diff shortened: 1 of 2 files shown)


clients/gtest/hipblaslt_gtest_ext_op.cpp
```diff
@@ -456,6 +488,56 @@ void AMaxTestWithScale(hipDataType    type,
...
+    std::vector<To> cpuOutput(1, -999.f);
+    std::vector<To> cpuOutputRcp(1, -999.f);
+    std::vector<Ti> cpuInput(m * n, 0.f);
+
+    hipblaslt_init_hpl(cpuInput, m * n, 1, m * n);
+
+    hipErr = hipMemcpyHtoD(gpuInput, cpuInput.data(), m * n * sizeof(Ti));
+    hipErr = hipMemset(gpuWorkSpace, 0, m * n * sizeof(Ti));
+    hipErr = hipMemset(gpuSync, 0, sizeof(std::int32_t));
+
```
solaslin, line 519: LGTM but just a minor suggestion: Why don't we use CHECK_HIP_ERROR on these hipErr. I believe it would make it more robust.
image

TonyYHsieh: good suggestion. but all of extop is not using it. so maybe we change it in another independent PR.

solaslin (approved):
LGTM, the CHECK_HIP_ERROR can be done in another PR

PR #910, opened 2024-07-18
Amax2D which support leading dimension

TonyYHsieh:
Amax2D which support leading dimension
TODO: M < 8
Limitation: address of input memory should below 4G memory.


TonyYHsieh:
hipblase-test pass

PR #926, opened 2024-07-22
gfx942_80CU HHS/BBS NN GridBased

babakpst:


babakpst:
- The same library is used for BBS and HHS. 
- Total number of sizes: 6855

nakajee:
Regarding this,
- The same library is used for BBS and HHS.

It is not a big deal, but the number of kernel is not exactly same between HHS and BBS.

babakpst:
Correct. And thanks for the comment. After benchmarking, there were a few sizes different in HHS and BBS with perf drop compared to gfx942. After adjusting for the perf drops, the number of kernels changed in HHS and BBS.

nakajee (approved):
Looks good

KKyang (approved):

PR #930, opened 2024-07-23
Gridbased batch map

yenong-amd:
This PR enables batch mapping in gridbased logic according to SWDEV-468488

Implementation
- Passing batch dimension for tuned sizes from library logic to solution library
- Adding batch to Gridbased debug printouts
- Checking for key size and assigning batch and K dimension correctly
- Implementing batch as nearest distance in gridbased distance function
- Modifying HHS NN, NT and TN logic yaml's to include batch tuned sizes

Testing
- Tested non-batched sizes to ensure no regression

### DIFF
--- a/tensilelite/Tensile/Source/lib/include/Tensile/Distance.hpp
+++ b/tensilelite/Tensile/Source/lib/include/Tensile/Distance.hpp
@@ -359,7 +359,8 @@ namespace Tensile
 
                 double M = p2[0];
                 double N = p2[1];
-                double K = p2[2];
+                double B = p2.size() > 3 ? p2[2] : 1;
+                double K = p2.size() > 3 ? p2[3] : p2[2];
 
                 // This is hard coding workaround solution //
                 // If incoming_size falls inside grid boundary (32768 in this case), searching toward larger M,N.
@@ -379,7 +380,12 @@ namespace Tensile
                         * std::pow(std::pow(p1[0] - p2[0], 2) + std::pow(p1[1] - p2[1], 2), 0.5));
                 }
                 // and nearest K
-                distance += (std::abs(K - p1[2]) / (K + p1[2] * 8));
+                double stepK = p1.size() > 3 ? p1[3] : p1[2];
+                distance += (std::abs(K - stepK) / (K + stepK * 8));
+
+                // and nearest B
+                double stepB = p1.size() > 3 ? p1[2] : 1;
+                distance += (std::abs(B - stepB) / (B + stepB * 8));
 
                 return distance;
             }
--- a/tensilelite/Tensile/SolutionLibrary.py
+++ b/tensilelite/Tensile/SolutionLibrary.py
@@ -89,7 +89,7 @@ class MatchingLibrary:
             3: Properties.Property("FreeSizeB", index=0),
             1: Properties.Property("BoundSize", index=0)
         }
-        if distance == "Equality":
+        if distance == "Equality" or distance == "GridBased":
             propertyKeys[0] = Properties.Property("BatchSize", index=0)
 
         properties = list([propertyKeys[i] for i in indices if i in propertyKeys])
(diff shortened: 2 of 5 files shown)


aazz44ss (approved):
LGTM

nakajee:
I am guessing you meant to use batch_count=32.

  - - [64, 16, 4096, 32]
    - [535, 0.0]
  
However, this is actually setting batch count = 4096 (MxNxK=64x16x32).
Please make sure if you set batch count as you intended or not.

tensilelite/Tensile/Source/lib/include/Tensile/Distance.hpp
```diff
@@ -379,7 +380,12 @@ namespace Tensile
                         * std::pow(std::pow(p1[0] - p2[0], 2) + std::pow(p1[1] - p2[1], 2), 0.5));
                 }
                 // and nearest K
-                distance += (std::abs(K - p1[2]) / (K + p1[2] * 8));
+                double stepK = p1.size() > 3 ? p1[3] : p1[2];
+                distance += (std::abs(K - stepK) / (K + stepK * 8));
+
+                // and nearest B
+                double stepB = p1.size() > 3 ? p1[2] : 1;
+                distance += (std::abs(B - stepB) / (B + stepB * 8));
```
nakajee, line 388: What is the meaning of * 8 here?
Should this be exactly same as K?

KKyang (commented):
Please make sure env TENSILE_GRIDBASED_KDTREE works with this change, thanks.

aazz44ss:
Is it better to search greater B, instead of nearest B?

tensilelite/Tensile/Source/lib/include/Tensile/Distance.hpp
aazz44ss: Code here is very old from previous searching algorithm before binary search.
We can simply use L1 or L2 distance for K

yenong-amd:
> I am guessing you meant to use batch_count=32.
> 
> * * [64, 16, 4096, 32]
>   * [535, 0.0]
> 
> However, this is actually setting batch count = 4096 (MxNxK=64x16x32). Please make sure if you set batch count as you intended or not.

This is as intended. The MNK are small but batch count is large up to 32768.

nakajee:
> > I am guessing you meant to use batch_count=32.
> > 
> > * * [64, 16, 4096, 32]
> >   * [535, 0.0]
> > 
> > However, this is actually setting batch count = 4096 (MxNxK=64x16x32). Please make sure if you set batch count as you intended or not.
> 
> This is as intended. The MNK are small but batch count is large up to 32768.

OK. Thanks for confirmation.
I did not think we needed very large batch count in grid.
Unlike K dimension, batch_count simply increase number of equivalent work groups.
In rocblas case, we did tuning for batch count up to number of CU.

aazz44ss:
@yenong-amd 
I found maybe there is something wrong with batch search.
if we have non batch grid
[128, 128, 1, 32]
[128, 160, 1, 32]
and batch grid
[128, 140, 304, 32]

in search steps, it binary search greater M, then binary search greater N, then search nearest batch and K
size [128, 128, 300, 32] will find [128, 128, 1, 32] instead of [128, 140, 304, 32].
size [128, 140, 1, 32] will find [128, 140, 304, 32] instead of [128, 160, 1, 32].
@nakajee is this expected?

yenong-amd:
I did not add MNK points that were not previously on the grid. The batch just projects those points onto batch dimension. Many existing grid points are missing batch dimension so the fallback is batch 1. I will add more batch tuning points in a future PR.

nakajee:
> @yenong-amd I found maybe there is something wrong with batch search. if we have non batch grid [128, 128, 1, 32] [128, 160, 1, 32] and batch grid [128, 140, 304, 32]
> 
> in search steps, it binary search greater M, then binary search greater N, then search nearest batch and K size [128, 128, 300, 32] will find [128, 128, 1, 32] instead of [128, 140, 304, 32]. size [128, 140, 1, 32] will find [128, 140, 304, 32] instead of [128, 160, 1, 32]. @nakajee is this expected?

We probably need to add batch_count sizes for existing M,N,K only. If we do not have exact matching [M,N,1,K] size, most likely there is some possibility that batched sizes is used for unexpected sizes.

aazz44ss (approved):
LGTM

nakajee (approved):
I am OK with this as the initial version of batch support.
We can try further improvement in the future.

PR #948, opened 2024-07-25
remove F8HS/HSS customkernel

Jinp800125:
remove F8HS/HSS customkernel


KKyang (approved):

aazz44ss (approved):
LGTM

aazz44ss:
There is also some code related to these custom kernel in ContractionSolution.cpp, do we need to also remove them?

PR #952, opened 2024-07-25
Switch to hipBLASCommon

bethune-bryant:
This PR is now ready for reviews!

In this PR we remove the dependency on hipBLAS and switch to using the hipBLAS-common package.


bnemanich (dismissed):

KKyang (dismissed):
Please use squash merge, thanks!

jichangjichang (changes requested):
Have you confirmed that rocm mainline build already has hipblas-common package installed?

jichangjichang (dismissed):
lgtm, please use squash merge.
It is better to update https://github.com/ROCm/hipBLASLt/blob/develop/README.md as well.

library/CMakeLists.txt
```diff
@@ -142,6 +142,7 @@ target_include_directories(hipblaslt
                                    $<BUILD_INTERFACE:${PROJECT_BINARY_DIR}/include>
                                    $<BUILD_INTERFACE:${HIP_INCLUDE_DIRS}>
                                    $<BUILD_INTERFACE:${HIPBLAS_INCLUDE_DIRS}>
+                                   $<BUILD_INTERFACE:${HIPBLAS-COMMON_INCLUDE_DIRS}>
```
AlexBrownAMD: Is HIPBLAS_INCLUDE_DIRS still required if this is using hipblas-common instead?

bnemanich: Oh, that is a good point. This should not be needed anymore. @awhittle3 can you please remove this?

awhittle3: Reference removed, builds as expected. I did not find any other references in the CMake files.

jichangjichang (approved):
lgtm, please use squash merge

bnemanich (approved):

PR #967, opened 2024-07-30
Fix potential integer overflow in KD-tree

Serge45:
## Brief ##
This PR fixes potential integer overflow for distance calculation in KD-tree based solution searching.


KKyang (approved):

yenong-amd (approved):
Tests passed.

PR #968, opened 2024-07-30
Modify data init

aazz44ss:
remove data initialization dependency of lda in hipblaslt-bench, so that we benchmark same data when leading dimension is different.

### DIFF
--- a/tensilelite/Tensile/Source/client/include/DataInitialization.hpp
+++ b/tensilelite/Tensile/Source/client/include/DataInitialization.hpp
@@ -44,7 +44,7 @@ namespace Tensile
 {
     namespace Client
     {
-        // Problem-indept. from 0~7, and 16 (fixed values for every problem)
+        // Problem-indept. from 0~7, and 16, and 23~26 (fixed values for every problem)
         // And problem-dept. from 8~15 (values depend on problem)
         // RandomNegPosLimited: integer -128~128. fp -1.0~1.0
         enum class InitMode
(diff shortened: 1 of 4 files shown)


tensilelite/Tensile/Source/client/include/DataInitialization.hpp
```diff
@@ -517,6 +525,18 @@ namespace Tensile
                 case InitMode::TrigCos:
                 case InitMode::TrigAbsSin:
                 case InitMode::TrigAbsCos:
+                case InitMode::TrigIndSin:
```
jichangjichang, line 528: case InitMode::Free: ~ case InitMode::TrigAbsCos:  would return runtime error originally.

tensilelite/Tensile/Source/client/include/DataInitialization.hpp
```diff
@@ -591,6 +611,18 @@ namespace Tensile
                 case InitMode::TrigAbsCos:
                     initArrayTrig<T, true, true>(array, tensor);
                     break;
+                case InitMode::TrigIndSin:
```
jichangjichang, line 614: is TrigSin equal to TrigIndSin?

tensilelite/Tensile/Source/client/include/DataInitialization.hpp
jichangjichang: Can we just support TrigSin ~ TrigAbsCos in getValue() without adding TrigIndXXX?

tensilelite/Tensile/Source/client/include/DataInitialization.hpp
aazz44ss: TrigInd is problem independent trig initialization, so that we don't need to re-initialize every time we change a problem during tuning.
TrigIndSign goes to initArray(InitMode mode, T* array, size_t elements), but we still need these codes, otherwise, compiler will report error.

tensilelite/Tensile/Source/client/include/DataInitialization.hpp
aazz44ss: TrigXXX we need to pass tensor sizes to get MNK
TrigIndXXX we need to pass total elements.

AndySu12 (approved):

Jinp800125 (approved):

aazz44ss:
[==========] 48206 tests from 13 test suites ran. (425493 ms total)
[  PASSED  ] 48206 tests.
hipBLASLt version: 1100

command line: ./hipblaslt-test

PR #774, opened 2024-05-17
Refactor tensilehost predicate and Remove pure gemm fallback mechanism

jichangjichang:

PR #969, opened 2024-07-30
Fixed out-of-bounds access of makeArgument in tensile_host.cpp

Serge45:
## Brief ##
Fixed potential segfault caused by out-of-bound access of `makeArgument` in tensile_host.cpp.

## Implementation
The original intention of `std::vector<int> useScaleAlphaVec` seems that for state saving/restoring, but it was not implemented properly. I made a fix for it.

## Tests
Local tests on gfx942 are all passed: [test_result_pr969.txt](https://github.com/user-attachments/files/16428008/test_result_pr969.txt)


KKyang (approved):

jichangjichang (approved):
lgtm

jichangjichang:
This code would not exist once #774 is merged

PR #976, opened 2024-08-02
Fix gpu arch regex in extop test

Serge45:
## Brief ##
Fixed GPU arch regex in extop test, it should cover both 11xx and 12xx.


cmingch (approved):

jichangjichang (approved):
lgtm

PR #980, opened 2024-08-05
gfx11/12 - correct v_wmma operand for wave64 to align compiler/doc

cmingch:


cmingch:
already did local verification with npi and mainline

hcman2:
Does this impact wave32?

cmingch:
> Does this impact wave32?

No, width of operand is calculated in code gen. the only hard code is here. we did _tryAssembler with wave64. due to compiler correct their behavior for wave64 in the latest build, so wrong operand result to capability detect incorrectly.

hcman2 (approved):

jichangjichang (approved):

PR #985, opened 2024-08-07
support i8 GRVW=1 in non-EccHalf platform

TonyYHsieh:
support i8 GRVW=1 in non-EccHalf platform


hcman2:
Should we add a test yaml to cover this kinds of issue?

TonyYHsieh:
gfx942 test pass
gfx1201 test same as before except i8 is passed.

jichangjichang (approved):
lgtm

PR #990, opened 2024-08-08
Update setuptools inside virtualenv

yoichiyoshida:
fix build issues on some systems with newer setuptools

### DIFF
--- a/cmake/virtualenv.cmake
+++ b/cmake/virtualenv.cmake
@@ -26,6 +26,7 @@ function(virtualenv_install)
     virtualenv_create()
     execute_process(
       COMMAND ${VIRTUALENV_BIN_DIR}/${VIRTUALENV_PYTHON_EXENAME} -m pip install --upgrade pip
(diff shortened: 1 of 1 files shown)


babakpst (approved):
We need this change to avoid setuptool version upgrade.

nakajee:
I wanted to confirm if this works with gfx942 CI node, but CI test issue is not resolved yet....
I encountered an issue with this change on my gfx942 node (it happens on my node only... Not sure why).
In my case, I need to specify setuptools version to avoid getting an error with newer verision (72).
   -q -U "setuptools<71.0"

KKyang:
@nakajee  Can you build now? Or is the error still existing?

nakajee:
> @nakajee Can you build now? Or is the error still existing?

The same error I saw before seems not happening with the latest mainline docker.
I think this is OK as long as the build pass with current CI environment.

KKyang (approved):

PR #1005, opened 2024-08-15
gridbased batch experiment

aazz44ss:
Environment variable to speedup wide range of batch gemm.
It multiply batch size to M or N depending on which one is larger.

### DIFF
--- a/tensilelite/Tensile/Source/lib/include/Tensile/PropertyMatching.hpp
+++ b/tensilelite/Tensile/Source/lib/include/Tensile/PropertyMatching.hpp
@@ -961,7 +961,7 @@ namespace Tensile
             }
 
             template <bool T_Debug>
-            std::vector<ReturnValue> findBestKeyMatch_GridBased(Key const& key,
+            std::vector<ReturnValue> findBestKeyMatch_GridBased(Key const& key_orig,
                                                                 Transform  transform,
                                                                 int        numSolutions) const
             {
@@ -974,6 +974,19 @@ namespace Tensile
                 ptrdiff_t count = 0;
                 bool      Debug = T_Debug;
                 std::cout << std::setprecision(2) << std::fixed;
+
+                Key key = key_orig;
+                if(!Debug::Instance().gridBasedBatchExp())
+                {
+                    if(key.size() > 3)
+                    {
+                        if(key[0] > key[1])
+                            key[0] = key[0] * key[2];
+                        else
+                            key[1] = key[1] * key[2];
+                        key[2] = 1;
+                    }
+                }
 
                 if(Debug::Instance().gridBasedKDTree())
                 {
--- a/tensilelite/Tensile/Source/lib/include/Tensile/Debug.hpp
+++ b/tensilelite/Tensile/Source/lib/include/Tensile/Debug.hpp
@@ -82,6 +82,8 @@ namespace Tensile
 
         bool gridBasedKDTree() const;
 
+        bool gridBasedBatchExp() const;
+
     private:
         friend LazySingleton<Debug>;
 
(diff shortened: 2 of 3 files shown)


yenong-amd (dismissed):
LGTM

tensilelite/Tensile/Source/lib/include/Tensile/PropertyMatching.hpp
```diff
@@ -975,6 +975,19 @@ namespace Tensile
                 bool      Debug = T_Debug;
                 std::cout << std::setprecision(2) << std::fixed;
 
+                Key key = key_orig;
+                if(!Debug::Instance().gridBasedBatchExp())
```
nakajee, line 979: Looks like this is enabled by default ( "!" in if condition)
If you would like to enable this by default, probably it is better not to use "Debug".

tensilelite/Tensile/Source/lib/include/Tensile/PropertyMatching.hpp
```diff
@@ -975,6 +975,19 @@ namespace Tensile
                 bool      Debug = T_Debug;
                 std::cout << std::setprecision(2) << std::fixed;
 
+                Key key = key_orig;
```
nakajee, line 978: Do we need to keep key_orig?

tensilelite/Tensile/Source/lib/include/Tensile/PropertyMatching.hpp
aazz44ss: @yenong-amd ask me to leave env variable to use previous behavior, in case there is regression or you want to do more test on gridbased tuning with batch.
But if we decide to abandon gridbased tuning with batch, we can remove it.

tensilelite/Tensile/Source/lib/include/Tensile/PropertyMatching.hpp
aazz44ss: We don't want original key to be modified, in case there are more hierarchical search in the future, like equality -> gridbased -> stream-k -> freesize

nakajee: It is not a good idea to change Key const & to Key at line 964.
We should be able to use copy constructor here without changing the type at line 964.

tensilelite/Tensile/Source/lib/include/Tensile/PropertyMatching.hpp
nakajee: I mean, this name is not best describing what this does.
You use the same name but changed the meaning (opposite behavior as default).

tensilelite/Tensile/Source/lib/include/Tensile/PropertyMatching.hpp
aazz44ss: done

tensilelite/Tensile/Source/lib/include/Tensile/PropertyMatching.hpp
aazz44ss: That means if you want to do gridbased tuning with batch experiment, you should use this env.

jichangjichang (approved):
lgtm

tensilelite/Tensile/Source/lib/include/Tensile/PropertyMatching.hpp
nakajee: OK. I was talking about the naming, not about the usage.
It is not a big deal. We can keep this name.

nakajee (approved):
I do not have further comment.

aazz44ss:
[==========] 35372 tests from 13 test suites ran. (1663461 ms total)
[  PASSED  ] 35372 tests.
hipBLASLt version: 800

PR #1045, opened 2024-08-27
perf: add argument to select cpp bench

solaslin:
Added an argument -e to **specify the folder of hipblaslt-bench**. This is for PTS mainly since it will run two benchmarks from two different commits. 
For a local bench you can simply ignore this argument and use the default value which is your release build.

### DIFF
--- a/clients/scripts/performance/hipblaslt-perf
+++ b/clients/scripts/performance/hipblaslt-perf
@@ -78,13 +78,19 @@ def runBenchmark(benchType, args, executable_folder, probYamlFolder, out_csv_Fil
         print(header)
         print(content)
 
-def command_perf(arguments, executable_folder, probYaml_foler):
+def command_perf(arguments, probYaml_foler):
     """Run bench"""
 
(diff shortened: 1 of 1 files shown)


clients/scripts/performance/hipblaslt-perf
```diff
@@ -147,7 +154,14 @@ def main():
...
                         help='workspace folder keeping the perf data',
-                        default=os.path.join(dir_of_this_repo, "hipBlasLt_benchmark/"))
+                        default=os.path.join(dir_of_this_repo, "hipBLASLt_benchmark/"))
+
+    # this argument is mainly used in jenkins/pts. For local dev, just use the default
+    parser.add_argument('-e',
+                        '--execFolder',
+                        type=str,
+                        help='folder where the cpp bench executables are located',
+                        default=executable_folder)
```
solaslin, line 165: The main change for this PR

KKyang (approved):

hcman2 (approved):

PR #1056, opened 2024-09-03
Revised output logic YAML to use efficiency instead of gflops.

waynehuangntu:
* Main modifications:
 1. Replacement of winnerGflops with effLike
 2. Addition of Frequency Monitor in TensileLite

* Notice: If Frequency is Nan or 0, original WinnerGflops will be used as the performance metric. The console will output warning to notify users WinnerGflops is in used instead of efficiency. 

image

Please refer to the comments in [link](https://ontrack-internal.amd.com/browse/SWDEV-474557) for further details.

### DIFF
--- a/tensilelite/Tensile/LibraryLogic.py
+++ b/tensilelite/Tensile/LibraryLogic.py
@@ -35,6 +35,7 @@ import array
 import csv
 import os
 import time
+import math
 
 ################################################################################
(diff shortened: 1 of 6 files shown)


jichangjichang (approved):
lgtm

vin-huang (approved):

PR #1068, opened 2024-09-05
Restore use-cache as default

davidd-amd:
@AlexBrownAMD observed that tensilelt was significantly slower than tensile when running as follows:

```bash
./Tensile/Bin/Tensile test.yaml tensile-out
```

Profiling revealed that `writeBenchmarkFiles` was called by tensilelt leading to the additional runtime. The reason for this is because `--use-cache` was changed to false in #410. It isn't clear if this was the intention because `store_true` has the effect of setting the arg to `False` if `--use-cache` is not passed.

In this PR, we restore the default to `True` and correct issues that arose now that we are following a different code path.


AlexBrownAMD (dismissed):
Tested these changes out with a sample test file and it significantly reduces cached build time.

AlexBrownAMD (approved):

davidd-amd:
@AlexBrownAMD @jichangjichang do we want to get this merged? If so, I will update to develop and retrigger tests.

davidd-amd:
@AlexBrownAMD @KKyang @jichangjichang this reduces the tox test time on the 90a node by 25%. Would like to get this reviewed and merged

tensilelite/Tensile/Tensile.py
```diff
@@ -207,7 +207,7 @@ def Tensile(userArgs):
     argParser.add_argument("--alternate-format", dest="AlternateFormat", action="store_true",
             help="Alternate format for config_file(s): first file is alternate config "
             "and optional second file is size list")
-    argParser.add_argument("--use-cache", dest="useCache", action="store_true",
+    argParser.add_argument("--use-cache", dest="useCache", action="store_false",
```
davidd-amd, line 210:
```suggestion
    argParser.add_argument("--no-cache", dest="useCache", action="store_false",
```

jichangjichang:
> @AlexBrownAMD @KKyang @jichangjichang this reduces the tox test time on the 90a node by 25%. Would like to get this reviewed and merged

We want the disable cache by default. In our daily work, we often change kernel generator.  The asm code could be different in each run of tensile benchmarking. We think cache should be enabled only when necessary but not enabling it by default.

PR #1007, opened 2024-08-15
Library logic filter for developer

Serge45:
## Brief ##
Added ~~`--dev-logic-filter`~~ `--logic-yaml-filter` to `install.sh` for developer. It's equivalent to the following glob pattern for logic searching:

```bash
<logic-root>/**/<logic-yaml-filter>.yaml
```
E.g.
```bash
bash ./install.sh -a gfx90a --logic-yaml-filter *BBS*
```
then only the following logics will be built

```bash   library/src/amd_detail/rocblaslt/src/Tensile/Logic/asm_full/aldebaran/110CU/GridBased/aldebaran_Cijk_Ailk_Bljk_BBS_BH_Bias_AH_SAV.yaml
library/src/amd_detail/rocblaslt/src/Tensile/Logic/asm_full/aldebaran/110CU/GridBased/aldebaran_Cijk_Ailk_Bjlk_BBS_BH_Bias_AH_SAV.yaml
library/src/amd_detail/rocblaslt/src/Tensile/Logic/asm_full/aldebaran/110CU/GridBased/aldebaran_Cijk_Alik_Bjlk_BBS_BH_Bias_AH_SAV.yaml
library/src/amd_detail/rocblaslt/src/Tensile/Logic/asm_full/aldebaran/110CU/GridBased/aldebaran_Cijk_Alik_Bljk_BBS_BH_Bias_AH_SAV.yaml
library/src/amd_detail/rocblaslt/src/Tensile/Logic/asm_full/aldebaran/104CU/Equality/aldebaran_Cijk_Alik_Bljk_BBS_BH_Bias_AH_SAV.yaml
library/src/amd_detail/rocblaslt/src/Tensile/Logic/asm_full/aldebaran/104CU/GridBased/aldebaran_Cijk_Ailk_Bljk_BBS_BH_Bias_AH_SAV.yaml
library/src/amd_detail/rocblaslt/src/Tensile/Logic/asm_full/aldebaran/104CU/GridBased/aldebaran_Cijk_Ailk_Bjlk_BBS_BH_Bias_AH_SAV.yaml
library/src/amd_detail/rocblaslt/src/Tensile/Logic/asm_full/aldebaran/104CU/GridBased/aldebaran_Cijk_Alik_Bjlk_BBS_BH_Bias_AH_SAV.yaml
[...]

PR #1074, opened 2024-09-08
Add option to build logics in sub-folder

solaslin:
## Descriptions:
In order to shorten our build time when working in our local side, we always add **-a** to build for specific arch. 
But we can do a further filter to pick a sub-path from arch and build only from the sub-path. 

This PR adds an new option `--local-sub-folder` as a handy option so that we don't need to manually remove those unwanted folder. ( Such as now we still need to remove gfx940, gfx941...even we already put -a gfx942. )

## A few examples:
- **`./install.sh -c -a gfx942 --logic-sub-folder gfx941`**
  * -a gfx942 will build from aquavanjaram folder, and `--logic-sub-folder gfx941` filters logics only if the path contains "gfx941"

- **`./install.sh -c -a gfx942 --logic-sub-folder gfx942`**
  * -a gfx942 will build from aquavanjaram folder, and `--logic-sub-folder gfx942` filters logics only if the path contains "gfx942"
  * This will still build gfx942_cuXX
  
- **`./install.sh -c -a gfx942 --logic-sub-folder gfx942/`**
  * -a gfx942 will build from aquavanjaram folder, and `--logic-sub-folder gfx942/` filters logics only if the path contains "gfx942/"
  * This can filter the folder gfx942_cuXX, only gfx942/ itself will be built
  
- **`./install.sh -c -a gfx942 --logic-sub-folder gfx942/Equality`**
  * -a gfx942 will build from aquavanjaram folder, and `--logic-sub-folder gfx942/Equality` filters logics only if the path contains "gfx942/Equality"
  * You don't even manually remove FreeSize, GridBased folder. Which might be a good way for us when developing.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -171,6 +171,7 @@ else()
     set( Tensile_COMPILER "amdclang++" CACHE STRING "Tensile compiler")
     set( Tensile_LIBRARY_FORMAT "msgpack" CACHE STRING "Tensile library format")
     set( Tensile_CPU_THREADS "" CACHE STRING "Number of threads for Tensile parallel build")
+    set( Tensile_BUILD_SUB_FOLDER "" CACHE STRING "Build only the logic files in the sub-folder" )
 
     option( Tensile_MERGE_FILES "Tensile to merge kernels and solutions files?" ON )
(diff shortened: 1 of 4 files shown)


solaslin:
See the discussion in PR 1007. PR1007 can also achieve the same thing :)
https://github.com/ROCm/hipBLASLt/pull/1007#issuecomment-2345470606

PR #1078, opened 2024-09-09
Fix stream-k tests for gfx90a

AlexBrownAMD:
The new stream-k tests include some test cases setting NonTemporalC/D. Gfx90a does not support NonTemporal=7, and was causing a build error for those kernels.

This change updates the test files to test NonTemporal=3 instead of 7 so it can run on gfx90a. This will fix the test cases for now, and I'll work on an additional change to detect the invalid parameter combination for the given architecture and either reject or update the settings as needed.


AlexBrownAMD:
Closing this PR in favor of PR 1080, which is a better fix for this issue.

PR #1080, opened 2024-09-09
Update NonTemporal setting according to AsmCaps

AlexBrownAMD:
During SolutionStructs, check NonTemporal settings against AsmCaps for valid values. If the setting is not supported (ie: "nt" bit) on the current architecture, then update the setting. This change ensures correct kernel name is applied, and avoids duplicate kernels due to NonTemporal settings.

### DIFF
--- a/tensilelite/Tensile/SolutionStructs.py
+++ b/tensilelite/Tensile/SolutionStructs.py
@@ -2009,6 +2009,12 @@ class Solution(collections.abc.Mapping):
       state["1LDSBuffer"] = 1
       print2("\nSet SIA=2, force PrefetchLocalRead=1, ExpandPointerSwap=1, 1LDSBuffer=1")
 
+    if not globalParameters["AsmCaps"][isa]["HasNTModifier"]:
+      # force to disable nt flag if it is not supported by arch
+      for ch in ["", "A", "B", "C", "D", "E", "WS", "Metadata"]:
+        if state["NonTemporal%s"%ch] >= 4:
+          state["NonTemporal%s"%ch] -= 4
+
     if state["WavefrontSize"] == 32 and not globalParameters["ArchCaps"][isa]["HasWave32"]:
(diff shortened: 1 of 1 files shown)


AlexBrownAMD:
tensilelite tox and hipblaslt-test tests all pass on gfx90a. Still running on gfx942

tensilelite/Tensile/SolutionStructs.py
```diff
@@ -2009,6 +2009,12 @@ def assignDerivedParameters(state):
       state["1LDSBuffer"] = 1
       print2("\nSet SIA=2, force PrefetchLocalRead=1, ExpandPointerSwap=1, 1LDSBuffer=1")
 
+    if not globalParameters["AsmCaps"][isa]["HasNTModifier"]:
```
nakajee, line 2012: I think this needs to be done after resolving state["NonTemporal"] != -1.
Otherwise, it can be overwritten by state["NonTemporal"].

nakajee: Sorry, I missed "".

nakajee (approved):
Looks good.

tensilelite/Tensile/SolutionStructs.py
AlexBrownAMD: I think it works as written as well. "NonTemporal" is included in the loop with the empty string "", so if "NonTemporal" is set to an invalid values, it will get updated before it gets copied to the other settings.

KKyang (approved):

AlexBrownAMD:
Thanks for the reviews. All 90a tests passed. tensilelite tox tests completed on 942, hipblaslt-test is still running. If you need the changes in sooner please merge. Otherwise I will merge them in the morning after the last of the tests have finished.

bethune-bryant (approved):

AlexBrownAMD:
942 tests finished overnight, all passed.

PR #1124, opened 2024-09-18
Fixed unsupported syntax in Python 3.6

Serge45:
## Brief
Similar to #1076, another fix is for built-in container type annotation, which is supported since Python 3.9.

## Reference
https://docs.python.org/3/library/typing.html#aliases-to-built-in-types


KKyang (approved):

jichangjichang (approved):
lgtm

PR #1129, opened 2024-09-18
Fix default value for tensile logic build parameter

AlexBrownAMD:
The install script allows you to specify which logic directory to build by setting the "-l" or "--logic" option. The default in the CMake file tries to set the value to "asm_full", but if the user doesn't specify a logic directory then the install script overrides the default with the base Logic directory causing all libraries to be built.

This change fixes the default logic directory by checking if the user set a value for the "-l" flag before overriding CMake's default value.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -380,6 +380,7 @@ matrices_dir=
 matrices_dir_install=
 gpu_architecture=all
 cpu_ref_lib=blis
+tensile_logic=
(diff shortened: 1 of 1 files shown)


nakajee (approved):
Looks good.

davidd-amd (approved):

KKyang (approved):

PR #1130, opened 2024-09-18
 Use the AOCL BLIS 4.2

NaveenElumalaiAMD:
resolves: [LWPTENSILE-1363](https://ontrack-internal.amd.com/browse/LWPTENSILE-1363)

Summary of proposed changes:
- Switched hipBLASLt to use BLIS-4.2 and the overall test (hipBLASLt-test) time reduced from 45 minutes to 24 minutes (Almost 2 times speed-up).
- The hipBLASLt use case, which was reported to have 250 times slower CPU-GFLOPs in this ticket  [LWPTENSILE-1363](https://ontrack-internal.amd.com/browse/LWPTENSILE-1363), had a performance boost from **0.3** to **788** GFLOPs when using BLIS-4.2.


clients/CMakeLists.txt
```diff
@@ -79,22 +79,44 @@ if ( NOT hip_FOUND )
...
+        set( BLIS_INCLUDE_DIR ${BUILD_DIR}/deps/blis/include/blis )
+      elseif(EXISTS      "/usr/local/lib/libblis.a")
+        set( BLAS_LIBRARY /usr/local/lib/libblis.a )
+        set( BLIS_INCLUDE_DIR /usr/local/include/blis )
+      else()
+        message( WARNING "Could not find libblis" )
+      endif()
+    else()
+      set( BLAS_LIBRARY "blas" )
+    endif()
```
TorreZuk: Not sure if you really need the full history of blis version support as this will be new dependency requirement on hipblasLt so could consider just one old and gcc-4.2.0 but probabably fine.

cmake/findBLIS.cmake
```diff
@@ -1,46 +0,0 @@
...
-# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
-# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
-# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
-# THE SOFTWARE.
-#
-# ########################################################################
-
-file(GLOB BLIS_AOCL_LIBS /opt/AMD/aocl/aocl-linux-aocc-*/aocc/lib_ILP64/libblis-mt.a)  # Possible location 1
-file(GLOB BLIS_LOCAL_LIB /usr/local/lib/libblis.a)  # Possible location 2
-file(GLOB BLIS_DEPS_LIB ${CMAKE_SOURCE_DIR}/build/deps/blis/lib/libblis.a)  # wget location
```
TorreZuk, line 26: So this is the fallback search history you might want to preserve not the rocblas one.

install.sh
```diff
@@ -312,27 +312,34 @@ install_msgpack_from_source( )
...
-      tar -xvf blis.tar.gz
-      rm -rf blis/amd-blis-mt
-      mv amd-blis-mt blis
-      rm blis.tar.gz
-      cd blis/lib
-      ln -sf libblis-mt.a libblis.a
+    if [[ ! -e "/opt/AMD/aocl/aocl-linux-gcc-4.2.0/gcc/lib_ILP64/libblis-mt.a" ]] &&
+        [[ ! -e "/opt/AMD/aocl/aocl-linux-aocc-4.1.0/aocc/lib_ILP64/libblis-mt.a" ]] &&
+        [[ ! -e "/opt/AMD/aocl/aocl-linux-aocc-4.0/lib_ILP64/libblis-mt.a"  ]] &&
+        [[ ! -e "/usr/local/lib/libblis.a" ]]; then
```
TorreZuk, line 319: Okay this should match whatever you decide in the cmake fallbacks,  so fine to keep both if others agree.

TorreZuk (dismissed):
LGTM with some considerations to get hipblasLt to confirm backward compatibility requirements.   
A required change should be the documentation update needs to describe the new dependency AOCL download and installation.

cmake/findBLIS.cmake
NaveenElumalaiAMD: Removed this file as it might be redundant after the changes as in the clients/CMakeLists.txt we directly set the path to BLAS_LIBRARY and in install.sh we check if there is other latest BLIS versions before installing the version 2.0.

clients/CMakeLists.txt
KKyang: Please modify the findBLIS.cmake instead of expanding everything here.

clients/common/cblas_interface.cpp
```diff
@@ -685,6 +685,187 @@ void cast_mul_with_Tci(customVector<TcCast>& dst,
...
+                    {
+                        C[j * ldc + i] = alpha * temp + beta * C[j * ldc + i];
+                    }
+                }
+            }
+        }
+    }
+}
+
 template <typename Tc>
```
KKyang, line 869: @mengzcai Does this meet your "reduce build time" design?

cmake/findBLIS.cmake
KKyang: install.sh is different from cmake, some people build hipblaslt directly from cmake commands.

install.sh
KKyang: Same reason here.

clients/common/cblas_interface.cpp
mengzcai: @KKyang  Yes, it's appropriate.

cmake/findBLIS.cmake
NaveenElumalaiAMD: I agree if we already have pre-installed dependencies, if not it is always better to go through install.sh.

cmake/findBLIS.cmake
```diff
@@ -21,26 +21,35 @@
...
-list (GET BLIS_LIBS 0 BLIS_LIB)
+
+set(BLIS_LIBS ${BLIS_AOCL_LIB_1} ${BLIS_AOCL_LIB_2} ${BLIS_DEPS_LIB} ${BLIS_LOCAL_LIB})
+
+# Remove empty strings
+foreach(item IN LISTS BLIS_LIBS)
+    if(NOT item STREQUAL "")
+        list(APPEND BLIS_LIBS_NON_EMPTY_LIST "${item}")
+    endif()
+endforeach()
```
TorreZuk: I don't understand the value of this pattern over a simple if /else if with precedence in the order of checking.  Am I missing something here.

NaveenElumalaiAMD: I did not want to change their pattern of finding the path to the AOCL BLIS, the only changes I did is adding the path to find AOCL BLIS 4.2 and removed the reverse and instead added remove empty strings.

install.sh
```diff
@@ -313,27 +313,34 @@ install_msgpack_from_source( )
...
-
-      tar -xvf blis.tar.gz
-      rm -rf blis/amd-blis-mt
-      mv amd-blis-mt blis
-      rm blis.tar.gz
-      cd blis/lib
-      ln -sf libblis-mt.a libblis.a
+    if [[ ! -e "/opt/AMD/aocl/aocl-linux-gcc-4.2.0/gcc/lib_ILP64/libblis-mt.a" ]] &&
+        [[ ! -e "/opt/AMD/aocl/aocl-linux-aocc-4.1.0/aocc/lib_ILP64/libblis-mt.a" ]] &&
+        [[ ! -e "/opt/AMD/aocl/aocl-linux-aocc-4.0/lib_ILP64/libblis-mt.a"  ]] &&
```
TorreZuk, line 318: For 4.0 did the include tree also change does it match BLIS_INCLUDE_DIR?   Not sure if you tested with that install.  I am not sure if you even need to support older than 4.2 as you are adding new dependency.   Usually can just what is advises in the documentation.

NaveenElumalaiAMD: If BLIS_LIB is `/opt/AMD/aocl/aocl-linux-aocc-4.0/lib_ILP64/libblis-mt.a` 
BLIS_INCLUDE_DIR will be `/opt/AMD/aocl/aocl-linux-aocc-4.0/include_ILP64/`
which seems to be right, i haven't tested it though. I am okay with changing it to just use 4.2.

NaveenElumalaiAMD:
This PR is ready to be merged, unless if there is any further comments. Thanks

babakpst (dismissed):

cmake/findBLIS.cmake
KKyang: You can change to simple if/else if you want. Although it's simple but it makes code longer if you want to find multiple versions and multiple locations.

KKyang (dismissed):

cmake/findBLIS.cmake
NaveenElumalaiAMD: Changed to simple if/else. Thanks!

babakpst (approved):

KKyang (approved):

PR #1143, opened 2024-09-20
Add F32/F64 Equality Logic Yamls for gfx12

wenchuanchen:
These two logic yamls are for [m=4096, n=4096, k=4096] benchmarking

PR #1157, opened 2024-09-24
Add F32/F64 Equality Logic Yamls for gfx12

wenchuanchen:
same as https://github.com/ROCm/hipBLASLt/pull/1143


jichangjichang (changes requested):
please add your commit on top of release-staging/rocm-rel-6.3 branch

PR #1160, opened 2024-09-24
fortran used by bench and test clients

TorreZuk:
* clients link and so use Fortran
* gfortran is one pulled as dependency in install.sh so assume it should be default

### DIFF
--- a/clients/benchmarks/CMakeLists.txt
+++ b/clients/benchmarks/CMakeLists.txt
@@ -82,7 +82,7 @@ if (NOT WIN32)
   endif()
   list( APPEND COMMON_LINK_LIBS "-lm -lstdc++fs")
 
-  if (CMAKE_Fortran_COMPILER MATCHES "gfortran")
+  if (CMAKE_Fortran_COMPILER_ID MATCHES "GNU")
     list( APPEND COMMON_LINK_LIBS "-lgfortran") # for lapack
   else()
     list( APPEND COMMON_LINK_LIBS "-lflang -lflangrti") # for lapack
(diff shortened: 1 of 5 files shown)


eidenyoshida:
the RHEL9 installation failure can be ignored. We have fixed this in math-ci

TorreZuk:
Well @eidenyoshida can you approve, need 2 approvals. But the status for tests seems delayed, runs are done but it still shows "beginning installation stage...."

eidenyoshida (dismissed):

clients/benchmarks/CMakeLists.txt
```diff
@@ -81,8 +81,11 @@ if (NOT WIN32)
...
   endif()
   list( APPEND COMMON_LINK_LIBS "-lm -lstdc++fs")
-  if (NOT BUILD_FORTRAN_CLIENTS)
-    list( APPEND COMMON_LINK_LIBS "-lgfortran -lflang -lflangrti") # for lapack
+
+  if (CMAKE_Fortran_COMPILER_ID MATCHES "GNU")
+    list( APPEND COMMON_LINK_LIBS "-lgfortran") # for lapack
+  else()
+    list( APPEND COMMON_LINK_LIBS "-lflang -lflangrti") # for lapack
   endif()
```
davidd-amd, line 89: Will the following not work?

```suggestion
   find_package(LAPACK Required)
   target_link_libraries(hipblaslt-bench PRIVATE LAPACK::LAPACK)
```

TorreZuk: The cmake config varies so much it wasn't reliable in the past, nor do we want to add an entire file of config rules just to find it.  This was meant as a quick fix but is dragging on due to other CI failures.   Could try to move to a find_package fallback (as AOCL will become default) but I think we will ask @NaveenElumalaiAMD to remove the lapack dependency entirely like done in rocblas.

TorreZuk: I think Eiden you should become a code owner, and so only 1 team member review required?

KKyang (approved):

TorreZuk:
@KKyang can you reapprove as Eiden accidentally requested you review

eidenyoshida:
Merging as it has passed the sles compilation stage that this PR is targeting

PR #1165, opened 2024-09-25
perf-script: added function field in csv data

solaslin:
Revised the perf script according to feedback from pts team:
- [x] Added a function field in csv data (matmul, amax...etc) For pts to do data query.
- [x] Fixed a argument parsing bug in amax client

### DIFF
--- a/clients/benchmarks/client_extop_amax.cpp
+++ b/clients/benchmarks/client_extop_amax.cpp
@@ -106,7 +106,7 @@ int parseArgs(int                       argc,
             }
             else if(arg == "-m" || arg == "--m")
             {
-                n = std::stoul(argv[++i]);
+                m = std::stoul(argv[++i]);
             }
             else if(arg == "-n" || arg == "--n")
             {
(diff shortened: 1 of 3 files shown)


clients/benchmarks/client_extop_amax.cpp
```diff
@@ -106,7 +106,7 @@ int parseArgs(int                       argc,
             }
             else if(arg == "-m" || arg == "--m")
             {
-                n = std::stoul(argv[++i]);
+                m = std::stoul(argv[++i]);
             }
```
solaslin, line 110: bug fixing.

clients/scripts/performance/suites.py
```diff
@@ -82,7 +82,7 @@ def amax_set_1():
 def matmul_set_1():
     """gemm benchset 1"""
 
-    problemlist = [Problem(args={"--yaml" : "matmul_probset1_bench.yaml"})]
+    problemlist = [Problem(args={"--log_function_name" : "" , "--yaml" : "matmul_probset1_bench.yaml"})]
     yield ProblemSet(benchType="matmul", name="benchset_1", problems=problemlist)
```
solaslin, line 86: thus csv data contains the function field.

jichangjichang (approved):
lgtm

Serge45 (approved):

PR #1169, opened 2024-09-26
Fix short circuit logic for alpha=0 with stream-k

AlexBrownAMD:
This update fixes the case when alpha=0 by ensuring that A/B matrices are not read and main loop does not run. Also added a new small test case with stream-k and alpha=0 to validate output.

### DIFF
--- a/tensilelite/Tensile/Components/StreamK.py
+++ b/tensilelite/Tensile/Components/StreamK.py
@@ -24,7 +24,7 @@ from ..TensileInstructions import Module, Label, SAddU32, RegisterPoolResource, 
     SCmpLtU32, SCSelectB32, sMagicDivAlg2, SMulI32, SSubU32, SMinU32, SMovB32, SCBranchSCC1, SCmpLeU32, VMovB32, vgpr, \
     SAddCU32, SCmpGtU32, SCMovB32, SAddI32, SCmpEQU32, SCBranchSCC0, SLShiftLeftB32, SLoadB32, SWaitCnt, SMEMModifiers, \
     log2, SBarrier, SStoreB32, SLongBranchPositive, SBranch, ceilDivide, replaceHolder, SNop, staticMultiply, SSleep, \
-    VAddF32, VAddF64, SAndB32, SLShiftRightB32, VReadfirstlaneB32
+    VAddF32, VAddF64, SAndB32, SLShiftRightB32, VReadfirstlaneB32, SBranchIfNotZero
 from ..Common import print2
 # from ..TensileInstructions.Containers import SMEMModifiers
 from ..Component import Component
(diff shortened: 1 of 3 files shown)


tensilelite/Tensile/Components/StreamK.py
```diff
@@ -1910,8 +1916,22 @@ def graWorkGroup(self, writer, kernel, tPA, tPB):
...
+        # Map SK index to WG
         module.add(self.skIndexToWG(writer, kernel, sTmp))
 
+        # Short circuit if alpha==0 (skip main loop and reading A/B, only do beta * C)
+        # To skip main loop in stream-k, we check if this WG is responsible for writing results (ie: WG starts tile)
+        # If WG starts tile then set LocalEnd=ItersPerTile to skip fixup step, and set loopCounter to 0 to skip main loop
+        # If WG does not start tile, skip to end of persistent loop to check for other SK tile
+        alphaLabel = Label("SKAlphaCheck", "")
+        module.add(SBranchIfNotZero("Alpha", kernel["ProblemType"]["ComputeDataType"], alphaLabel))
+        # Skip to end if not doing the global write
```
nakajee, line 1928: What does this mean?
We need global write even if alpha==0, right?

AlexBrownAMD: We need global write of beta * C (unless beta is 0, then it skips that part too)

tensilelite/Tensile/Tests/common/streamk/sk_sgemm_alpha0.yaml
```diff
@@ -0,0 +1,119 @@
+TestParameters:
+  marks: [skip-gfx900, skip-gfx906, skip-gfx1010, skip-gfx1011, skip-gfx1012, skip-gfx1030, skip-gfx1100, skip-gfx1101, skip-gfx1102, skip-gfx1200, skip-gfx1201] # not supported by arch
+
+GlobalParameters:
+  NumElementsToValidate: -1
+  BoundsCheck: False
+  KernelTime: False
+  DataInitTypeAlpha: 0
+  DataInitTypeBeta: 1
```
nakajee: Maybe better to use Beta !=1 to make sure beta * C is properly stored.

tensilelite/Tensile/Components/StreamK.py
nakajee: OK. I understand the logic.
StreamKLocalStart=0 WG does global write.
All other WGs with StreamKLocalStart>1, no global write, right?

AlexBrownAMD: Yep that's right

KKyang:
https://github.com/ROCm/hipBLASLt/blob/c961e7dbd621ddddd95eb6853e4a992d13dbeedd/tensilelite/Tensile/KernelWriterAssembly.py#L1882

Suggest adding a comment or implement this path to ensure stream-K works with backward pass in the future.

tensilelite/Tensile/Tests/common/streamk/sk_sgemm_alpha0.yaml
AlexBrownAMD: Updated

AlexBrownAMD:
> Suggest adding a comment or implement this path to ensure stream-K works with backward pass in the future.

I don't see any code that needs to be changed in this part right now for stream-k so I added a comment to this section to say how the short-circuit works for the stream-k case

nakajee (approved):
Looks good

KKyang (approved):

neoblizz (approved):
Looks good to me

PR #1170, opened 2024-09-26
Dependency on hipblas-common-dev

awhittle3:
resolves #SWDEV-487853

The dependency on [hipblas-common](https://github.com/ROCm/hipBLAS-common) actually needs to point to the hipblas-common-dev package.

With this change, hipblaslt-dev package will now list hipblas-common-dev as a dependency.

Thanks to @lawruble13 for the fix!


bnemanich (approved):

jichangjichang (approved):
lgtm, please also provide hotfix PR for release-staging/rocm-rel-6.3 branch

PR #1171, opened 2024-09-26
Update dependency list with the actual package name

raramakr:
hipblas-common-devel/dev is the actual package name and it provides hipbals-common. Use the actual package name in dependency.


awhittle3:
Please close in favour of #1170 . Only the devel packages require the `hipblas-common-dev` package.

raramakr:
Handled in https://github.com/ROCm/hipBLASLt/pull/1170, so closing the PR

PR #1180, opened 2024-09-30
split local write instructions.

hcman2:
ds_store_b128 takes more cycles then 1 mfma latency. 
Split b128 into b32's and schedule into different mfma if ppossible.

### DIFF
--- a/tensilelite/Tensile/Components/SIA.py
+++ b/tensilelite/Tensile/Components/SIA.py
@@ -23,9 +23,11 @@
 from ..TensileInstructions import Item, Module, HolderContainer, Instruction, \
                                 GlobalReadInstruction, LocalReadInstruction, \
                                 LocalWriteInstruction, SSetPrior, SWaitCnt, \
-                                replaceHolder, fastdeepcopy, VMovB32
+                                replaceHolder, fastdeepcopy, VMovB32, \
+                                DSStoreB128, DSStoreB64, DSStoreB32
 from ..Common import roundUp
 from ..Component import SIA
+from ..TensileInstructions.Containers import DSModifiers
 
 import copy
 from math import ceil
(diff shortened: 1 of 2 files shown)


jichangjichang:
would this cause LDS bank conflict?

hcman2:
> would this cause LDS bank conflict?
I don't think this patch will change the LDS bank behavior.
**ds_store_b128 addr, vgpr[0:3], offset:0**
is changed into
**ds_store_b32 addr, vgpr[0], offset:0
ds_store_b32 addr, vgpr[1], offset:4
ds_store_b32 addr, vgpr[2], offset:8
ds_store_b32 addr, vgpr[3], offset:12**
if ds_store_b128 has no bank conflict, ds_store_b32 should has no bank conflict either.

hcman2:
[==========] 48206 tests from 13 test suites ran. (1332135 ms total)
[  PASSED  ] 48206 tests.

jichangjichang:
> > would this cause LDS bank conflict?
> > I don't think this patch will change the LDS bank behavior.
> > **ds_store_b128 addr, vgpr[0:3], offset:0**
> > is changed into
> > **ds_store_b32 addr, vgpr[0], offset:0
> > ds_store_b32 addr, vgpr[1], offset:4
> > ds_store_b32 addr, vgpr[2], offset:8
> > ds_store_b32 addr, vgpr[3], offset:12**
> > if ds_store_b128 has no bank conflict, ds_store_b32 should has no bank conflict either.

ds_store_b32 operations are issued by 32 threads at the same cycle.
ds_store_b128 operations are issued by 8 threads only at the same cycle.
This PR should change the access pattern of LDS banks for each cycle.

hcman2:
gfx90a passed
[==========] 13061 tests from 13 test suites ran. (480980 ms total)
[  PASSED  ] 13061 tests.

hcman2:
========== 76 passed, 32 skipped, 1258 warnings in 8858.51s (2:27:38) ==========
gfx94x tox passed.

hcman2:
If we split a b128 into 4 x b32. We will have default 4-way bank conflict.
The extra cycles for 1 4-way bank conflict b32 instruction are 2, 1 for addressing and 1 for data.
The basic ds_store_b32 is about 56 cycles so we have no instruction stall (less then 4 x mfma).
Considering the barrier, we have to make sure that the final LW has more then 56+2 cycles to be waited.

KKyang (approved):

jichangjichang (commented):
Is there any perf test result proving the most of cases would get uplift?

hcman2:
On real test , the LW b32 has 108 cycle latency with 4-way bank conflict.

hcman2:
> Is there any perf test result proving the most of cases would get uplift?

Yes.
My test example is TN MT256x192 with problem size [8192,7296,1,8192]. 
For this example we have 192 mfma's = 192x16 = 3072 cycles
And we have 14 ds_write_b128 = 14 x (20 - 12) = 112 cycles, which we can save with this PR.
The best eff% gain will be 112 / (3072+112) = ~ 3.5% eff
The performance difference in 2 different machine is:
Machine 1: 1506.08us  650198GFLOPS  ------> 1501.99us  651971GFLOPS    ~0.3% gain
Machine 2: 1525.66us  641856GFLOPS  ------> 1493.4us  655721GFLOPS      ~2.2% gain
In this example the higher eff leads the higher power consumption, so that the lower performance.
The performance may not have 3.x% gain as the eff%.

jichangjichang (approved):

PR #1183, opened 2024-09-30
Bug fix in stream-k tile assignment

AlexBrownAMD:
Fixes a bug in tile assignment when running dynamic grid mode.


nakajee:
Can we add a test case to trigger this issue?

AlexBrownAMD:
> Can we add a test case to trigger this issue?

I'm looking into how I could add a test case for it. It would require setting an environment variable to enable dynamic grid mode, which I'm not sure if we can do currently in the python tests.

nakajee (approved):
Thanks for your comment.
It is probably difficult to set some specific env variable for unit test.
OK for now.

neoblizz (approved):

nakajee (approved):
Looks good.

KKyang (approved):

PR #1191, opened 2024-10-01
Remove duplicate fields in some logic files

b-shi:
Removed duplicate fields in some logic files, the value kept corresponds to the value used in the global `problemType` field.


nakajee (commented):
Looks good.

babakpst (approved):

Jinp800125 (approved):
LGTM

PR #1207, opened 2024-10-04
add fp16 configs for gfx942_20cu

fsx950223:


library/src/amd_detail/rocblaslt/src/Tensile/Logic/asm_full/aquavanjaram/gfx942_20cu/Equality/aquavanjaram_Cijk_Ailk_Bjlk_HHS_BH_UserArgs.yaml
```diff
@@ -0,0 +1,85 @@
...
+  UseScaleAB: ''
+  UseScaleAlphaVec: 0
+  UseScaleCD: false
+- []
+- [2, 3, 0, 1]
+- null
+- null
+- null
+- DeviceEfficiency
+- GridBased
```
nakajee: This part should be same as other existing Equality logic.

fsx950223: Thanks. Done.

library/src/amd_detail/rocblaslt/src/Tensile/Logic/asm_full/aquavanjaram/gfx942_20cu/Equality/aquavanjaram_Cijk_Ailk_Bjlk_HHS_BH_UserArgs.yaml
```diff
@@ -82,4 +82,4 @@
 - null
 - null
 - DeviceEfficiency
-- GridBased
+- Equality
```
nakajee, line 85: I am not sure how you generated this yaml, but only this file has 3 "- null".
It should be 2.
I just realized this does not have any size information.
Then, we do not need this file.

PR #1216, opened 2024-10-08
findBLIS should check local build/dep

m-kim:
install.sh will install blis in build/dep directory if it's not found, so findBLIS.cmake should look there as well.


cmake/findBLIS.cmake
```diff
@@ -39,6 +39,9 @@ elseif(EXISTS      "${BUILD_DIR}/deps/blis/lib/libblis.a")
 elseif(EXISTS      "/usr/local/lib/libblis.a")
     set( BLIS_LIB /usr/local/lib/libblis.a )
     set( BLIS_INCLUDE_DIR /usr/local/include/blis )
+elseif(EXISTS "${CMAKE_CURRENT_BINARY_DIR}/../deps/blis")
```
NaveenElumalaiAMD: Does Line-36 (elseif(EXISTS "${BUILD_DIR}/deps/blis/lib/libblis.a")) not do this?

m-kim: It doesn't. Where is BUILD_DIR set?

babakpst (dismissed):

cmake/findBLIS.cmake
NaveenElumalaiAMD: If the `${BUILD_DIR}` is not set then we could replace all `${BUILD_DIR}`  with `${CMAKE_CURRENT_BINARY_DIR}/..` that should fix the issue as well.

cmake/findBLIS.cmake
```diff
@@ -30,12 +30,12 @@ elseif(EXISTS      "/opt/AMD/aocl/aocl-linux-aocc-4.1.0/aocc/lib_ILP64/libblis-m
...
-elseif(EXISTS      "${BUILD_DIR}/deps/amd-blis/lib/ILP64/libblis-mt.a") # 4.0 and 4.1.0
-    set( BLIS_LIB ${BUILD_DIR}/deps/amd-blis/lib/ILP64/libblis-mt.a )
-    set( BLIS_INCLUDE_DIR ${BUILD_DIR}/deps/amd-blis/include/ILP64 )
-elseif(EXISTS      "${BUILD_DIR}/deps/blis/lib/libblis.a")
-    set( BLIS_LIB ${BUILD_DIR}/deps/blis/lib/libblis.a )
-    set( BLIS_INCLUDE_DIR ${BUILD_DIR}/deps/blis/include/blis )
+elseif(EXISTS "${CMAKE_CURRENT_BINARY_DIR}/../deps/amd-blis/lib/ILP64/libblis-mt.a") # 4.0 and 4.1.0
+    set( BLIS_LIB ${CMAKE_CURRENT_BINARY_DIR}/../deps/amd-blis/lib/ILP64/libblis-mt.a )
+    set( BLIS_INCLUDE_DIR ${CMAKE_CURRENT_BINARY_DIR}/../deps/amd-blis/include/ILP64 )
+elseif(EXISTS "${CMAKE_CURRENT_BINARY_DIR}/../deps/blis")
```
NaveenElumalaiAMD: `/lib/libblis.a` is missing in the search path. This is to check before setting the path to `BLIS_LIB`

m-kim: Fixed.

NaveenElumalaiAMD (approved):
LGTM!

jichangjichang (approved):
lgtm

PR #1222, opened 2024-10-10
Enable variant builds via device ID and cu count

bstefanuk:
This PR adds functionality in TensileCreateLibrary to filter out logic files that don't match an architecture variant, and thereby omit those solutions from being built into libraries. This reduces disk space for the built libraries and reduces peak memory and build time because there are less libraries to build. A variant is specified by a set of device IDs and CU counts and is added the to the pre-existing `--architecture` CLI argument. In this context, device IDs are the PCI device IDs and can be found with `lspci`.

**Notable changes**
String syntax for specifying a variant:
```
# Building for multiple GFX architectures remains the same
$ ./TensileCreateLibrary --architecture "gfx942;gfx901"

# To build libraries for only gfx942s that have 80 CUs
$ ./TensileCreateLibrary --architecture "gfx942[cu=80]"

# To build libraries for only gfx942s that have *either* 80 CUs or the device ID 74a5
$ ./TensileCreateLibrary --architecture "gfx942[cu=80, id=74a5]"
```

**Implementation**
The string argument between `[...]` is parsed into a mapping of device IDs and CU counts. A quick loop over the logic files extracts metadata from each file and conditionally keeps it if the metadata matches the requested variant(s).

The technique implemented only parses the first 4 lines of each logic file when pulling metadata, saving time and memory since we don't have to import the entire file. Measurements show that this method is 80x faster than the naive implementation.

**Caveats**
- There are some logic files in the Experimental/ directory that have an atypical structure and thus can't be parsed by the custom metadata parser. To resolve this, the update simply ignores any files in the Experimental/ directory.
- Most logic files don't contain device ID for the devices they support, and typically only contain the emulation IDs. These will have to be updated to make full use of the device ID predicate.

**Testing**
Unit tests have been added for all new functions.

### DIFF
--- a/tensilelite/Tensile/Com/ArchVariant.py
+++ b/tensilelite/Tensile/Com/ArchVariant.py
@@ -2,6 +2,8 @@ import re
 from pathlib import Path
 from typing import NamedTuple, Optional, Union, Tuple, Set, Dict, List
 
+from ..Common import printWarning
+
 class ArchVariant(NamedTuple):
     Name: str
     Gfx: str
@@ -36,7 +38,7 @@ def _extractArchVariant(file: Union[str, Path]) -> ArchVariant:
 
     def l0(line: str):
         if not re.match(r"- \{MinimumRequiredVersion", line):
-            raise LogicFileError(f"Expected minimum required version: line: {line}")
+            raise LogicFileError(f"Expected minimum required version:\n  line: {line}  file: {file}")
 
     def l1(line: str):
         return line[2:].strip()
@@ -57,10 +59,9 @@ def _extractArchVariant(file: Union[str, Path]) -> ArchVariant:
         if re.match(r"- \[Device", line):
             devIds = re.findall(r"Device (\w+)", line)
             
-            # Temporary, until we add the correct IDs
             if any(id in emulationIds for id in devIds):
-                # printWarning("Emulation device ID found, interpreting as fallback device...")
-                return None
+                printWarning("Emulation device ID(s) found, ignoring...")
+                devIds = filter(lambda id: id not in emulationIds, devIds)
             return set(f"id={id}" for id in devIds)
         if re.match(r"-\[alldevices", line.lower().replace(" ", "")):
             return None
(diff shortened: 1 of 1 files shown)


tensilelite/Tensile/TensileCreateLibrary.py
```diff
@@ -1401,6 +1405,15 @@ def validLogicFile(p: Path):
   print1(f"# LogicFilter: {globPattern}")
   logicFiles = (os.path.join(logicPath, file) for file in glob.iglob(globPattern, recursive=True))
   logicFiles = [file for file in logicFiles if validLogicFile(Path(file))]
+  logicFiles = [file for file in logicFiles if not "Experimental/" in file]
```
bstefanuk: Remove this and add in a separate PR.

tensilelite/Tensile/Tests/unit/test_ArchVariant.py
```diff
@@ -0,0 +1,213 @@
...
+
+    # Device ID doesn't match
+    mock_extract.return_value = MockArchVariant("foo", "gfx942", {"1234", "5678"}, 64)
+    assert matchArchVariant({"gfx942"}, {"9999"}, {64}, "valid_file.txt") == False
+
+    # CU count doesn't match
+    mock_extract.return_value = MockArchVariant("foo", "gfx942", {"1234", "5678"}, 64)
+    assert matchArchVariant({"gfx942"}, {"1234"}, {32}, "valid_file.txt") == False
+
+    assert matchArchVariant({"gfx942"}, {"1234"}, {64}, "experimental/file.txt") == False
```
bstefanuk: Add newline at EOF.

jichangjichang:
https://github.com/ROCm/hipBLASLt/pull/1007
Could you check this PR which provide similar functionality with this one?
Do we really need another feature to filter logic yaml?

bstefanuk:
@jichangjichang This approach adds a more structured approach to variant filtering and allows for intelligent selection of solutions from the metadata of logic files, as opposed to relying solely on details in the filename.

bstefanuk:
Closing as stale with the option to re-open.

PR #1234, opened 2024-10-14
Conditionally populate _format9 array in scoped function

davidd-amd:
The validMFMA and validSMFMA arrays are built in the global scope which means they are replicated across all processes, growing the memory consumption of TensileCreateLibrary. We moved the creation of the `_format9` entry in these dictionaries to assignGlobalParameters and made populating this entry conditional. For `TensileCreateLibrary` by default we will not create the `_format9` entry but the behavior is unchanged for `Tensile`. 

This change reduces the peak memory consumption from ~78 GB to ~70 GB when building for gfx942 with 64 jobs.


tensilelite/Tensile/Common.py
```diff
@@ -1609,6 +1585,39 @@ def assignGlobalParameters( config ):
 
   global globalParameters
 
+  if config["ValidateMatrixInstructions"]:
```
bstefanuk, line 1589: A comment here clarifying that this is set to True by default when executing ./Tensile but set to False when executing ./TensileCreateLibrary would be helpful for future developers to understand the context of this code.

tensilelite/Tensile/Common.py
```diff
@@ -1609,6 +1585,39 @@ def assignGlobalParameters( config ):
...
+    for MFMA in [validMFMA["H"], validMFMA["S"], validMFMA["B"], validMFMA["D"], validMFMA["X"], validMFMA["F8"], validWMMA]:
+      for MI in MFMA:
+        for bm in range(int(math.log(MI[3],2))+1):
+          for tt0 in range(1,validTT+1):
+            for tt1 in range(1,validTT+1):
+              for wave_m in range (3):
+                for wave_n in range(3):
+                  validMFMA["_format9"].append([MI[0],MI[1],MI[2],MI[3],2**bm,tt0,tt1,2**wave_m, 2**wave_n])
+    validMatrixInstructions = [[], [-1]] + validMFMA["H"] + validMFMA["S"] + validMFMA["B"] + validMFMA["D"] + validMFMA["B1k"] + validMFMA["X"]
+    validMatrixInstructions = validMatrixInstructions + validMFMA["_format9"]
```
bstefanuk, line 1605: Create a function `populateValidMatrixInstructions`

tensilelite/Tensile/Common.py
```diff
@@ -1609,6 +1585,39 @@ def assignGlobalParameters( config ):
...
+        for bm in range(int(math.log(MI[3],2))+1):
+          for tt0 in range(1,validTT+1):
+            for tt1 in range(1,validTT+1):
+              for wave_m in range (3):
+                for wave_n in range(3):
+                  validSMFMA["_format9"].append([MI[0],MI[1],MI[2],MI[3],2**bm,tt0,tt1,2**wave_m, 2**wave_n])
+    validSparseMatrixInstructions = validSMFMA["H"] + validSMFMA["B"] + validSMFMA["4xi8"]
+    validMatrixInstructions = validMatrixInstructions + validSparseMatrixInstructions + validSMFMA["_format9"]
+
+    validParameters["MatrixInstruction"] = validMatrixInstructions
```
bstefanuk, line 1620: Create a function `populateValidMatrixInstructions` for better encapsulation.

tensilelite/Tensile/TensileCreateLibrary.py
```diff
@@ -1268,6 +1268,8 @@ def splitExtraParameters(par):
   argParser.add_argument("--package-library",        dest="PackageLibrary",    action="store_true", default=False)
   argParser.add_argument("--embed-library",          dest="EmbedLibrary",
                          help="Embed (new) library files into static variables.  Specify the name of the library.")
+  argParser.add_argument("--validate-matrix-instructions", dest="ValidateMatrixInstructions", action="store_true",
```
bstefanuk, line 1270: Shouldn't this be set to `action="store_false"` because we want to skip the validation for TensileCreateLibrary to save memory?

tensilelite/Tensile/Common.py
KKyang: Is there a way to pass the same object to all working process instead of turning off? If this is turned off, there is no way to check the correctness of the logic yaml files.

bstefanuk: @KKyang There are options in joblib to allow for the use of shared memory, but we haven't found success when implementing it, even on a prototype level.

The intention here is that the validation should already be conducted prior to saving this as a long-term logic file, e.g., in Tensile. Thus, the step shouldn't be required in TensileCreateLibrary. While keeping this validation does provide some peace of mind, it's also one of the largest contributors to memory consumption, so we feel the trade off is worth it.

davidd-amd: @KKyang this is off by default in the context of0` TensileCreateLibrary` because I couldn't find where the check was applied. However, it is on by default for Tensile which in that context the check is applied.

> If this is turned off, there is no way to check the correctness of the logic yaml files.

We shouldn't validate logic files as a part of invoking `TensileCreateLibrary`. It should be assumed that the logic files are already validated. Otherwise, we do redundant work every time we build. Seems like validation could occur at the time of creating the Logic files or we can have a separate module specifically for logic file validation.

tensilelite/Tensile/TensileCreateLibrary.py
davidd-amd: Please refer to the docs here:

https://docs.python.org/3/library/argparse.html

the semantics of store_<true/false> is what to do if the optional argument is encountered. In this case, if we encounter `--validate-matrix-instructions` we want to store the value true if not encountered it defaults to the opposite which is false. For the case of TCL we will default to false because we don't actually use this state for TCL.

tensilelite/Tensile/Common.py
davidd-amd: Good catch!

davidd-amd:
Will include this work future PR

PR #1249, opened 2024-10-17
Enable test option for stream-k full tile or remainder tile

AlexBrownAMD:
Re-enable environment variable that can be used to test the number of stream-k tiles that get run in a stream-k kernel.

The default (in 2-tile algorithm) is to run all remainder tiles + 1 extra tile per workgroup. This ensures cases like remainder=1 are not split across a large number of workgroups. But for cases when remainder tiles is large (ie: remainder = WGs - 1), it is generally faster to run only the remainder tiles rather than remainder + #WGs. This way it will run more tiles in data-parallel and improve memory alignment.

This change adds an environment variable that allows the user to override the number of stream-k tiles selected. I developed a prediction model to select the best setting automatically, but that still needs to be ported from Tensile in a future change (this PR is just a first step to allow some experiments to run).

TENSILE_STREAMK_FULL_TILES=0 (runs remainder only)
TENSILE_STREAMK_FULL_TILES=1 (default, runs remainder + #WGs)
TENSILE_STREAMK_FULL_TILES=[large number] (can be used to make a 2-tile kernel double as basic algorithm, ie StreamK=1)

### DIFF
--- a/tensilelite/Tensile/Common.py
+++ b/tensilelite/Tensile/Common.py
@@ -942,9 +942,9 @@ validParameters = {
     # TENSILE_STREAMK_FIXED_GRID lets you override the default grid size with a specific number
     #   0 = override disabled (default)
     # TENSILE_STREAMK_FULL_TILES sets the number of full tiles to be included in stream-k work
-    #   -1 = use prediction model for best performance (default)
+    #   -1 = use prediction model for best performance (not yet implemented)
     #   0 = only remainder tiles run in stream-k
-    #   1+ = remainder + 1 (or more) full grids of tiles run in stream-k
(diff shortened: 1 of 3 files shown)


nakajee (dismissed):
Looks good.

jichangjichang (commented):
should we modify the description of TENSILE_STREAMK_FULL_TILES in Common.py?

AlexBrownAMD:
> should we modify the description of TENSILE_STREAMK_FULL_TILES in Common.py?

Good call, just posted an update with how it currently works in hipBLASLt repo.

nakajee (approved):
Thanks for your update.
Looks good.

KKyang (approved):

PR #1256, opened 2024-10-18
Fix temp register alignment in stream-k alpha check

AlexBrownAMD:
Fix alignment of temp register for divides in stream-k code. This bug was causing errors generating kernels for some gfx942 hardware.

### DIFF
--- a/tensilelite/Tensile/Components/StreamK.py
+++ b/tensilelite/Tensile/Components/StreamK.py
@@ -64,7 +64,7 @@ class XCCMappingOn(XCCMapping):
             sqTmp = writer.sgprPool.checkOut(1, "sqTmp", preventOverflow=False)
             divisor = kernel["StreamKXCCMapping"]
             if ((divisor & (divisor - 1)) != 0): # Need temp registers if not power of 2
-                sTmp = writer.sgprPool.checkOut(2, "sTmp", preventOverflow=False)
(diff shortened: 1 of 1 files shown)


tensilelite/Tensile/Components/StreamK.py
```diff
@@ -64,7 +64,7 @@ def __call__(self, writer, kernel):
             sqTmp = writer.sgprPool.checkOut(1, "sqTmp", preventOverflow=False)
             divisor = kernel["StreamKXCCMapping"]
             if ((divisor & (divisor - 1)) != 0): # Need temp registers if not power of 2
-                sTmp = writer.sgprPool.checkOut(2, "sTmp", preventOverflow=False)
+                sTmp = writer.sgprPool.checkOutAligned(2, 2, "sTmp", preventOverflow=False)
```
nakajee, line 67: The change itself looks OK, but I have a question.
This line is inside "with writer.allocTmpSgpr(4) as tmpSgprRes".
Is it really safe to call another sgprPool.checkOut here?
Can we allocate 6 at line 57(with align 2) in case divisor is not power of 2?

AlexBrownAMD: It is generating a good range of correct kernels so far for testing so I don't think it is affecting the sgpr cap. But I do also have a ticket filed to look at sgpr reductions as a future change.

bethune-bryant (approved):
LGTM.

KKyang (approved):

PR #1277, opened 2024-10-25
Resolve RPM failure

davidd-amd:
When using RPM to package the tensilelite artifacts, /usr/lib/rpm/strip fails. This change set modifies the RPM spec file such that llvm-strip is used rather then the system strip.


TorreZuk (approved):
Looks good.  Should be cherry picked next week so a quick review desired

KKyang (approved):

PR #1284, opened 2024-10-28
Move from namespace Tensile to TensileLite

davidd-amd:
Use TensileLite to disambiguate from Tensile project when building statically.


TorreZuk (dismissed):
LGTM if all tests pass and team is okay with it.

davidd-amd:
Should we namespace this file:

[hipBLASLt/library/src/amd_detail/hipblaslt-ext-op-internal.hpp](https://github.com/ROCm/hipBLASLt/blob/4f485824b2a4c2b96a638d728e0ebc9ed3ce5001/library/src/amd_detail/hipblaslt-ext-op-internal.hpp)

e.g. amd_detail?

TorreZuk:
> Should we namespace this file:
> 
> [hipBLASLt/library/src/amd_detail/hipblaslt-ext-op-internal.hpp](https://github.com/ROCm/hipBLASLt/blob/4f485824b2a4c2b96a638d728e0ebc9ed3ce5001/library/src/amd_detail/hipblaslt-ext-op-internal.hpp)
> 
> e.g. amd_detail?

Suggest: hipblaslt , TensileLite, or anonymous
Should indicate ownership, look at other math libraries which do use namespaces. rocSPARSE I think does

davidd-amd:
@jichangjichang @KKyang  need approval for this. Please let me know if you are aware of any downstream projects that will be impacted by this change.

jichangjichang (approved):

TorreZuk:
@ellosel assume you built rocblas against this PR without issue.   Frameworks you could ping but if these weren't in the public API no one can really complain too much.

davidd-amd:
> @ellosel assume you built rocblas against this PR without issue. Frameworks you could ping but if these weren't in the public API no one can really complain too much.

Correct - that is how I found the duplicate symbols.

bstefanuk (approved):

PR #1302, opened 2024-11-01
GFX942 equality tuning for F8HS and F8B8HS for TN,NT,NN

smalekta:
Adding new kernels/equality libraries for F8HS+F8B8HS for TN,NT,NN with bias


smalekta:
(Supporting files will be uploaded) Tests are passing 

[==========] 48206 tests from 13 test suites ran. (642284 ms total)
[  PASSED  ] 48206 tests.
hipBLASLt version: 1100

smalekta:
reduced the number of added/changed libraries to only with bias. They can handle both with and without bias case with minimal perf difference according to our data

smalekta:
(On the latest commit in the PR)

[==========] 48206 tests from 13 test suites ran. (453950 ms total)
[  PASSED  ] 48206 tests.
hipBLASLt version: 1100

babakpst (approved):

jichangjichang:
> reduced the number of added/changed libraries to only with bias. They can handle both with and without bias case with minimal perf difference according to our data

If so, please revise PR title accordingly

jichangjichang (approved):

PR #1304, opened 2024-11-04
Modify the condition to check resource.

hcman2:
The number of resource in different platform are different. Use different condition.

### DIFF
--- a/tensilelite/Tensile/KernelWriterAssembly.py
+++ b/tensilelite/Tensile/KernelWriterAssembly.py
@@ -11487,9 +11487,28 @@ class KernelWriterAssembly(KernelWriter):
         self.getLdsSize(kernel), self.agprPool.size(), self.states.doubleVgpr) < 2:
       self.states.overflowedResources = 6
 
-    vgprPerCU = 65536
+    if self.states.version[0] == 10:
+      vgprPerCU = 1024 * 32
+    elif self.states.version[0] == 11:
+      if self.states.version[2] == 2:
+        vgprPerCU = 1024 * 32
+      else:
+        vgprPerCU = 1536 * 32
+    elif self.states.version[0] == 12:
+      vgprPerCU = 1536 * 32
+    elif self.states.version[0] == 9:
+      if self.states.archCaps["ArchAccUnifiedRegs"]:
+        vgprPerCU = 2048 * 64
+      else:
+        vgprPerCU = 1024 * 64
+    else:
+      assert 0, "No valid VGPR value for this platform"
+    if self.states.archCaps["ArchAccUnifiedRegs"]:
+      totalVgprUsed = self.vgprPool.size() + self.agprPool.size()
+    else:
+      totalVgprUsed = max(self.vgprPool.size(), self.agprPool.size())
     vgprPerThreadPerOccupancy = vgprPerCU // kernel["NumThreads"]
-    numWorkGroupsPerCU = vgprPerThreadPerOccupancy // max(self.vgprPool.size(), self.agprPool.size())
+    numWorkGroupsPerCU = vgprPerThreadPerOccupancy // totalVgprUsed
     if numWorkGroupsPerCU < 1:
       self.states.overflowedResources = 4
 


tensilelite/Tensile/KernelWriterAssembly.py
```diff
@@ -11487,9 +11487,17 @@ def functionSuffix(self, kernel):
...
-    vgprPerCU = 65536
-    vgprPerThreadPerOccupancy = vgprPerCU // kernel["NumThreads"]
-    numWorkGroupsPerCU = vgprPerThreadPerOccupancy // max(self.vgprPool.size(), self.agprPool.size())
+    if self.states.version[0] is 10:
+      vgprPerCU = 1024
+    elif self.states.version[0] is 11:
+      vgprPerCU = 1536
+    elif self.states.version[0] is 12:
+      vgprPerCU = 1536
+    else:   # gfx9 serise
```
AndySu12: serise -> series

tensilelite/Tensile/KernelWriterAssembly.py
```diff
@@ -11487,9 +11487,20 @@ def functionSuffix(self, kernel):
...
+      vgprPerCU = 1024
+    elif self.states.version[0] is 11:
+      if self.states.version[2] is 2:
+        vgprPerCU = 1024
+      else:
+        vgprPerCU = 1536
+    elif self.states.version[0] is 12:
+      vgprPerCU = 1536
+    else:   # gfx9 series
+      vgprPerCU = 2048
```
jichangjichang: consider self.archCaps["ArchAccUnifiedRegs"]

hcman2: Why ArchAccUnifiedRegs? This is the physical VGPR size. We only care about the total VGPR usage. If a WG uses VGPRs more than the physical sizes, the HIP API will return error. This check is to avoid the error.

hcman2:
navi3 tox pass
----- generated xml file: /menghung/hipBLASLt/tensilelite/python_tests.xml -----
================== 9 passed, 101 skipped in 293.92s (0:04:53) ==================

hcman2:
gfx940 CI + tox pass

tensilelite/Tensile/KernelWriterAssembly.py
jichangjichang: If no ArchAccUnifiedRegs, the actual allocation of accVgpr (or archVgpr) would be equal to max(self.vgprPool.size(),self.agprPool.size()).
For gfx908, if NumThreads is 512 and vgprPool.size() is 256, the new formula would calculate numWorkGroupsPerCU as 1, which wouldn't trigger overflowResources. But the actual case is that it wouldn't have enough vgpr.

hcman2:
Cannot use wavefroutsize because navi platform can set either wave32 or wave64.
Use the native wavefront size instead.

AndySu12 (dismissed):

AndySu12 (approved):

hcman2:
Add assert for the unsupported platform. 
New developer for new platform needs to add the VGPR number.

tensilelite/Tensile/KernelWriterAssembly.py
```diff
@@ -11487,9 +11487,28 @@ def functionSuffix(self, kernel):
         self.getLdsSize(kernel), self.agprPool.size(), self.states.doubleVgpr) < 2:
       self.states.overflowedResources = 6
 
-    vgprPerCU = 65536
+    if self.states.version[0] == 10:
```
jichangjichang, line 11490: should we use WaveFrontSize to calculate vgprPerCU for all cases?

hcman2: WavefrontSize is a SW configuration in the code gen, not the native HW wavefront size. For example, navi3 and navi4 support both wave32 and wave64 configuration. When navi3+wave64, the VGPRs used are doubled. Hence, we have to use the HW default wavefront setting instead of the SW configuration.

hcman2:
gfx12 passed
===================================================================================== 14 passed, 96 skipped, 4 warnings in 244.86s (0:04:04) ======================================================================================
py3: exit 0 (245.02 seconds) /workspace/source/hipBLASLt/tensilelite> py.test -v --basetemp=/workspace/source/hipBLASLt/tensilelite/tensile-tox/py3/tmp --junit-xml=/workspace/source/hipBLASLt/tensilelite/python_tests.xml --junit-prefix=py3 --color=yes -n 4 --prebuilt-client=/workspace/source/hipBLASLt/tensilelite/tensile-tox/py3/client/0_Build/client/tensile_client Tensile/Tests -m common pid=1053525
  py3: OK (380.06=setup[10.46]+cmd[0.82,123.76,245.02] seconds)
  congratulations :) (380.10 seconds)

jichangjichang (approved):

hcman2:
[==========] 48206 tests from 13 test suites ran. (858571 ms total)
[  PASSED  ] 48206 tests.
hipBLASLt version: 1100
gfx940 passed

PR #1309, opened 2024-11-04
Fix occupancy does not calculate correctly

KKyang:


jichangjichang:
Please describe the issue/root cause/ solution in the PR's description

tensilelite/Tensile/KernelWriterAssembly.py
```diff
@@ -74,7 +74,7 @@ def getVgprOccupancy(self, numThreads, vgprs, doubleVgpr=False):
     maxOccupancy = self.consts.maxOccupancy//multiplier
 
     vgprAllocateAligned = 4    if not doubleVgpr else 8
-    totalVgprs = self.consts.maxVgprs if not doubleVgpr else self.consts.maxVgprs*2
```
jichangjichang, line 77: But vgprs could be 512, why did you remove the 2*maxVgrps?

KKyang:
Close due to this is not the root cause of the issue

PR #1311, opened 2024-11-04
reduce default gfx targets

TorreZuk:


jichangjichang (commented):
Have we confirmed that we can get rid of gfx940 and gfx941entirely?

jichangjichang (approved):

hcman2 (approved):

PR #1322, opened 2024-11-06
Use find_package Python

davidd-amd:
The python version used significantly impacts TensileCreateLibrary performance as the python version will also impact pip installed dependencies as well such as joblib. This change enables more idiomatic find usage and find_package will find the most recent version of python installed in the event that there are two such is the case for rhel8 images where we install both the default python release (3.6.8) and a more recent version available through the system package manager like 3.8 or 3.9.

Initial testing indicates that this change coupled with a more recent python install reduces build times by ~30%.

### DIFF
--- a/cmake/virtualenv.cmake
+++ b/cmake/virtualenv.cmake
@@ -1,18 +1,16 @@
-# find_package(PythonInterp)
-# # TODO: Check PYTHON_VERSION_MAJOR
 
-find_program(VIRTUALENV_PYTHON_EXE python3)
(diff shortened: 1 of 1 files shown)


davidd-amd:
@jichangjichang @KKyang this PR needs to go in ASAP to help accelerate builds. The failing tests have nothing to do with this CMake change. We should look into why those failures are occurring and why we appear to hang for so long in those cases.

jichangjichang (approved):

bstefanuk (approved):

PR #1341, opened 2024-11-14
Fix dependency check in unrolled loop with numItersPLR == 0

Serge45:
- Fixed dependency check via register range in some cases
 - Added missing dependency check for SMFMAInstruction
 - Removed type hints that only supported in Python 3.10+.

PR #1346, opened 2024-11-15
Change syntax of Union for earlier python versions

daineAMD:
Union `|` syntax works for python 3.10+, so changing to Union[] for earlier version of python


daineAMD:
Fixed in https://github.com/ROCm/hipBLASLt/pull/1341

PR #1349, opened 2024-11-15
Add sgpr occupancy

KKyang:


tensilelite/Tensile/Common.py
```diff
@@ -1115,6 +1115,8 @@ def supportedCompiler(compiler: str) -> bool:
     # Intended for use with custom kernels which have confirmed to be correct
     "NoReject":                    [False, True],
 
+    "TotalSgprNumber":              list(range(0,801)),
```
jichangjichang: Don't add new kernel parameter for occupancy calculation

KKyang: I add this because of TotalVgprNumber, should I remove this too?

aazz44ss (approved):

vin-huang (approved):

PR #1355, opened 2024-11-18
Remove Min/Max/TotalVgprNumber in Common.py

KKyang:


aazz44ss (approved):

jichangjichang:
Are all changes in logic yaml? If there is some change in python code, please separate logic yaml change and python code change as 2 commits

KKyang:
> Are all changes in logic yaml? If there is some change in python code, please separate logic yaml change and python code change as 2 commits

Done

jichangjichang (approved):

PR #1365, opened 2024-11-19
Library Logic Format Simplification

b-shi:
This PR makes a few changes to the library logic format and related for simplification purposes:

1. Fields in the library logic are now structured as key-value pairs (previously a list)
2. Removed `ProblemType` field in each solution - the global one is used instead
3. Add a `DefaultSolution` field in each logic file.
4. For each solution, parameters that are set to default their values are omitted in the logic file.
    * During library load, we reference `DefaultSolution` in each logic file to fill in the missing parameter values
5. Add a new field `LibraryLogicVersion` to track any logic file format changes in the future.
6. Add a new merge script `mergeV2.py` for merging logic files in new format .

With the format changes the total disk usage of the logic files is reduced ~40%. Sample logic file using new format [here](https://github.com/b-shi/hipBLASLt/blob/liblogic1/library/src/amd_detail/rocblaslt/src/Tensile/Logic/asm_full/aquavanjaram/gfx942/GridBased/aquavanjaram_Cijk_Ailk_Bjlk_B8BS_BH_BiasSB_HAS_SAB_SAV_UserArgs.yaml). 

[Tensilite changes](https://github.com/ROCm/hipBLASLt/commit/863eda2d01aebb772ab81a31c5a0407387233e6d) are in a separate commit for ease of review.

### DIFF
--- a/tensilelite/Tensile/LibraryIO.py
+++ b/tensilelite/Tensile/LibraryIO.py
@@ -91,6 +91,8 @@ def writeYAML(filename, data, **kwargs):
         kwargs["explicit_end"] = True
     if "default_flow_style" not in kwargs:
         kwargs["default_flow_style"] = None
+    if "sort_keys" not in kwargs:
+        kwargs["sort_keys"] = False
 
     with open(filename, "w") as f:
         yaml.dump(data, f, **kwargs)
@@ -251,7 +253,20 @@ def parseLibraryLogicFile(filename, cxxCompiler, archs=None):
 def parseLibraryLogicData(data, srcFile, cxxCompiler, archs=None):
     """Parses the data of a library logic file."""
     if isinstance(data, List):
+        # TODO: this can be removed when all logic files have dict format
         data = parseLibraryLogicList(data, srcFile)
+    elif isinstance(data, dict) and "LibraryLogicVersion" in data:
+        libraryType = data["LibraryType"]
+        if libraryType == "FreeSize":
+            data["Library"] = {}
+            data["Library"]["indexOrder"] = None
+            data["Library"]["table"] = [0, len(data["Solutions"])]
+            data["Library"]["distance"] = None
+        elif libraryType == "Matching":
+            data["Library"] = {}
+            data["Library"]["indexOrder"] = data["IndexOrder"]
+            data["Library"]["table"] = data["ExactLogic"]
+            data["Library"]["distance"] = libraryType
 
     is_arch_valid = lambda cArch, tArch : (cArch == tArch or cArch == "all")
     if not (archs is None) and "ArchitectureName" in data:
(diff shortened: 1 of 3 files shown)


jichangjichang:
why do we need to keep merge.py? Should we just update merge.py for new format?

tensilelite/Tensile/LibraryIO.py
```diff
@@ -320,11 +343,10 @@ def parseLibraryLogicList(data, srcFile="?"):
                 .format(srcFile, len(data)))
 
     rv = {}
+
+    rv["LibraryLogicVersion"] = "0.0.0"
```
bstefanuk, line 356: We hard code the library logic version to 0.0.0? Should this instead be extracted from the library logic directly?

**edit:** I see, the original list-style logic files won't have a version; disregard, or perhaps a comment could be useful.

tensilelite/Tensile/LibraryIO.py
```diff
@@ -251,7 +253,20 @@ def parseLibraryLogicFile(filename, archs=None):
 def parseLibraryLogicData(data, srcFile="?", archs=None):
     """Parses the data of a library logic file."""
     if isinstance(data, List):
+        # TODO: this can be removed when all logic files have dict format
         data = parseLibraryLogicList(data, srcFile)
+    else:
+        libraryType = data["LibraryType"]
```
bstefanuk: I wonder if here on the `else` branch, we should convert it into an `elif` and check against the expected type (Dict) and the library logic version, then add an else block that raises if nothing matches?

KKyang:
I don't think we need a mergeV2.py, we should let merge.py to support the new format and make sure it also supports transferring legacy yamls to new format.

b-shi:
> why do we need to keep merge.py? Should we just update merge.py for new format?

@babakpst, can we replace the old `merge.py` with the new version instead?

b-shi:
> I don't think we need a mergeV2.py, we should let merge.py to support the new format and make sure it also supports transferring legacy yamls to new format.

Just to check, if we convert all the existing logic files to the newer format, would we still have a need to support the legacy format?  One of the motivations for converting all the logic files to the new format in this PR is to avoid adding support for the legacy format (main issue is there a several places where the fields are hard coded).

tensilelite/Tensile/LibraryIO.py
b-shi: Yeah, `parseLibraryLogicList` can be removed once all the logic files are in the new format. Currently its only being used to help convert the legacy format to new format.

KKyang:
The reason why we still need a converter is that customers are using our tuning guides and having their own logic yamls kept in their hands. This change without a converter will break the compatibility.

tensilelite/Tensile/LibraryIO.py
b-shi: Thanks, good point, I will have it check the `LibraryLogicVersion` and type

b-shi:
> The reason why we still need a converter is that customers are using our tuning guides and having their own logic yamls kept in their hands. This change without a converter will break the compatibility.

Ah.. thats a good point, thoughts on adding a convert option in `merge.py` and then having most of the logic in `merge.py` only support the dictionary format?

b-shi:
Closing this PR - created new one [here](https://github.com/ROCm/hipBLASLt/pull/1702).

PR #1366, opened 2024-11-19
Skip the very first cold iteration from gpu time measurement 

TomokoKurotobi:
I am new to contribution to the public repository.
I like the new feature to optionally skip running hot iteration for potentially slow solutions that @jichangjichang made.

Since the very first kernel launch may have a big overhead (I have experienced this), how about adding one additional cold iteration for this feature, and start time measurement from 2nd cold iteration?

### DIFF
--- a/clients/include/testing_matmul.hpp
+++ b/clients/include/testing_matmul.hpp
@@ -2974,11 +2974,14 @@ void testing_matmul_with_bias(const Arguments& arg,
                             gemmVec[b].initialize(heuristicResult[sol].algo,
                                                   tuningVec[heuristicTuningIndex[sol]],
                                                   *dWorkspace));
-                    if(arg.skip_slow_solution_ratio)
-                        pre_gpu_time(
-                            arg.use_gpu_timer, event_gpu_time_start, gpu_time_used, stream);
-                    for(int i = 0; i < number_cold_calls; i++)
+                                                  
+                    const int cold_itr_addition = arg.skip_slow_solution_ratio ? 1 : 0;
+                    for(int i = 0; i < number_cold_calls + cold_itr_addition; i++)
                     {
+                        if (i == 1 && arg.skip_slow_solution_ratio){
+                            pre_gpu_time(
+                                arg.use_gpu_timer, event_gpu_time_start, gpu_time_used, stream);
+                        }
                         CHECK_HIPBLASLT_ERROR(gemmVec[i % block_count].run(stream));
                         if(i == 0 && (arg.unit_check || arg.norm_check || arg.allclose_check))
                             copy_gemm_to_host(stream, gemm_count, hD_1, (*dDp));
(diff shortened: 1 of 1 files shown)


clients/include/testing_matmul.hpp
```diff
@@ -2974,11 +2974,14 @@ void testing_matmul_with_bias(const Arguments& arg,
...
                                                   *dWorkspace));
-                    if(arg.skip_slow_solution_ratio)
-                        pre_gpu_time(
-                            arg.use_gpu_timer, event_gpu_time_start, gpu_time_used, stream);
-                    for(int i = 0; i < number_cold_calls; i++)
+                                                  
+                    const int cold_itr_addition = arg.skip_slow_solution_ratio ? 1 : 0;
+                    for(int i = 0; i < number_cold_calls + cold_itr_addition; i++)
                     {
+                        if (i == 1 && arg.skip_slow_solution_ratio){
```
nakajee, line 2981: pre_gpu_time will not be called if number_cold_calls is 0.
First of all, number_cold_calls needs to be >=1 if arg.skip_slow_solution_ratio is true.
I suggest to set number_cold_calls = 1 if arg.cold_iters == 0 and arg.skip_slow_solution_ratio == true.

jichangjichang:
> I am new to contribution to the public repository. I like the new feature to optionally skip running hot iteration for potentially slow solutions that @jichangjichang made.
> 
> Since the very first kernel launch may have a big overhead (I have experienced this), how about adding one additional cold iteration for this feature, and start time measurement from 2nd cold iteration?

It is too complicated for using. I don't like this change. We won't use just 1 iter for warm-up, and the skip mechanism also can be adjusted tolerance by user. I don't think you need to care the first iter of cold iter.

clients/include/testing_matmul.hpp
TomokoKurotobi: number_cold_calls will be at least 1 according to source code, and if skip_slow_solution_ratio is true, I add another addition before for loop, so cold iteration loop will run at least twice.

TomokoKurotobi:
> > I am new to contribution to the public repository. I like the new feature to optionally skip running hot iteration for potentially slow solutions that @jichangjichang made.
> > Since the very first kernel launch may have a big overhead (I have experienced this), how about adding one additional cold iteration for this feature, and start time measurement from 2nd cold iteration?
> 
> It is too complicated for using. I don't like this change. We won't use just 1 iter for warm-up, and the skip mechanism also can be adjusted tolerance by user. I don't think you need to care the first iter of cold iter.

Thanks for your response.
I will close this PR, then

clients/include/testing_matmul.hpp
TomokoKurotobi: @nakajee Sorry, I was wrong. Depending on the argument, number_cold_calls can be zero or negative, I think.
image

PR #1382, opened 2024-11-22
Logic fix to exclude streamk by default

mahmoodw:
When trying to add streamk compatibility with tunable ops we noticed that the default behaviour was including streamk kernels in the solutions being considered. This change corrects the logic to only include them when the associated env variable is set to request them.

I've tested using the following combinations:
TENSILE_SOLUTION_SELECTION_METHOD=0 TENSILE_DB=0x8000 --algo_method heuristic
correctly **not** considering any sk3 kernels
TENSILE_SOLUTION_SELECTION_METHOD=2 TENSILE_DB=0x8000 --algo_method heuristic
correctly considering sk3 kernels
TENSILE_SOLUTION_SELECTION_METHOD=0 TENSILE_DB=0x8000 --algo_method all
correctly **not** considering any sk3 kernels
TENSILE_SOLUTION_SELECTION_METHOD=2 TENSILE_DB=0x8000 --algo_method all
correctly considering sk3 kernels

Note:
formatting changes are done by clang as suggested by hipBLASLt team in previous PR
.githooks/pre-commit

### DIFF
--- a/tensilelite/Tensile/Source/lib/include/Tensile/ExactLogicLibrary.hpp
+++ b/tensilelite/Tensile/Source/lib/include/Tensile/ExactLogicLibrary.hpp
@@ -2,7 +2,7 @@
  *
  * MIT License
  *
- * Copyright (C) 2022-2023 Advanced Micro Devices, Inc. All rights reserved.
+ * Copyright (C) 2022-2024 Advanced Micro Devices, Inc. All rights reserved.
  *
  * Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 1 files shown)


AlexBrownAMD (approved):
Tested this change with a local build and it is working as expected

yenong-amd (approved):
The current change only excludes streamK solutions from default getAllSolutions. I found that gridbased solutions are included in streamK getAllSolutions when TENSILE_SOLUTION_SELECTION_METHOD=2.

KKyang (approved):

PR #1228, opened 2024-10-11
re-enable DirectToLds

nakajee:

PR #1384, opened 2024-11-24
PrintSolutionRejectionReason = True doesn't work at 4_LibraryClient stage

Jay0521:


jichangjichang (commented):
stage4 should always be able to generate the winner solutions.
Which case you encountered still fails to be generated in stage4?

Jay0521:
> stage4 should always be able to generate the winner solutions. Which case you encountered still fails to be generated in stage4?

After this PR (https://github.com/ROCm/hipBLASLt/pull/1228), some sparse kernels cannot be generated. I want to find out the reason for kernel failure from stage 4, but currently, I can only set PrintSolutionRejectionReason = True in common.py. Therefore, I modified this part so that PrintSolutionRejectionReason can be set directly from the YAML file and executed correctly. Is this modification feasible?

jichangjichang:
> > stage4 should always be able to generate the winner solutions. Which case you encountered still fails to be generated in stage4?
> 
> After this PR (#1228), some sparse kernels cannot be generated. I want to find out the reason for kernel failure from stage 4, but currently, I can only set PrintSolutionRejectionReason = True in common.py. Therefore, I modified this part so that PrintSolutionRejectionReason can be set directly from the YAML file and executed correctly. Is this modification feasible?

If a kernel is generated failed in stage1, it shouldn't be target in stage4.

Jay0521:
> > > stage4 should always be able to generate the winner solutions. Which case you encountered still fails to be generated in stage4?
> > 
> > 
> > After this PR (#1228), some sparse kernels cannot be generated. I want to find out the reason for kernel failure from stage 4, but currently, I can only set PrintSolutionRejectionReason = True in common.py. Therefore, I modified this part so that PrintSolutionRejectionReason can be set directly from the YAML file and executed correctly. Is this modification feasible?
> 
> If a kernel is generated failed in stage1, it shouldn't be target in stage4.

Understand. Thanks.

PR #1389, opened 2024-11-25
Add extended profile logging along with flush and rotating size 

NaveenElumalaiAMD:
1. Added Extended profile logging which prints out extra information: Solution Index, Solution name and Kernel name.
2. Add flush and rotating size to bench, profile and Extended profile logging.

### DIFF
--- a/library/src/amd_detail/hipblaslt-ext-op.cpp
+++ b/library/src/amd_detail/hipblaslt-ext-op.cpp
@@ -563,3 +563,15 @@ void hipblasltSetRotatingBufferSizeValue(int newrotatingBufferSize)
     UserClientArguments clientArguments;
     clientArguments.SetRotatingBufferSizeValue(newrotatingBufferSize);
 }
+
+void hipblasltSetColdIterationsValue(int newColdIterations)
+{
+    UserClientArguments clientArguments;
+    clientArguments.SetColdIterationsValue(newColdIterations);
+}
+
+void hipblasltSetHotIterationsValue(int newHotIterations)
+{
+    UserClientArguments clientArguments;
+    clientArguments.SetHotIterationsValue(newHotIterations);
+}
--- a/library/src/amd_detail/hipblaslt-ext-op.cpp
+++ b/library/src/amd_detail/hipblaslt-ext-op.cpp
@@ -558,8 +558,8 @@ void hipblasltSetFlushValue(bool newFlush)
     UserClientArguments clientArguments;
     clientArguments.SetFlushValue(newFlush);
 }
-void hipblasltSetRotatingMemorySizeValue(int newrotatingMemorySize)
+void hipblasltSetRotatingBufferSizeValue(int newrotatingBufferSize)
 {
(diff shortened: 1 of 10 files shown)


library/src/amd_detail/hipblaslt-ext-op.cpp
```diff
@@ -543,3 +550,38 @@ hipblasStatus_t hipblasltAMaxWithScaleRun(const hipDataType datatype,
...
+// Getter and setter for the rotatingMemorySize member variable, only for internal usage purpose.
+int SingletonUserClientArguments::hipblasltInternalGetRotatingMemorySizeValue() const
+{
+    return rotatingMemorySize;
+}
+void SingletonUserClientArguments::hipblasltInternalSetRotatingMemorySizeValue(
+    int newrotatingMemorySize)
+{
+    rotatingMemorySize = newrotatingMemorySize;
+}
```
TorreZuk: It's not clear you need to export the class and selected methods.     It's name is it in namespace?
Why not just export a few functions and implement getting the singleton instance and calling the class methods. The only exported symbols could be the setters (do you need getters in client?), e.g.
hipblasltInternalSetRotatingMemorySizeValue(
    int newrotatingMemorySize);
  All exported symbols should have a unique prefix for the library, so hipblasltInternal.... should suffice for that.

KKyang:
Customer is complaining about the log format changes frequently, if there's any other changes, please submit them in this PR.

amd-jnovotny (dismissed):
Docs changes OK.

library/src/amd_detail/hipblaslt-ext-op.cpp
```diff
@@ -298,28 +298,29 @@ hipblasStatus_t hipblasltSoftmaxRun(hipDataType datatype,
     auto        gpu       = TensileLite::hip::GetCurrentDevice();
     const auto  archName  = trimArchName(gpu->archName());
     auto&       masterLib = getExtOpMasterLibrary();
-    const auto& lib
-        = masterLib
-              .getLibrary(archName, hipblaslt_ext::SoftmaxSolutionLibrary::opName, hipDataTypeo_char(datatype))
-              ->as<hipblaslt_ext::SoftmaxSolutionLibrary>();
-    auto sol
-        = lib.findBestSolution(hipblaslt_ext::SoftmaxProblem(m, n, hipDataType_to_tensile_type(datatype)), *gpu);
+    const auto& lib       = masterLib
```
smalekta, line 301: NBD, but these are just reformatting?  makes PR review harder

NaveenElumalaiAMD: Yes these are formatting changes. Maybe standardizing the clang-format could make it easy.

babakpst (dismissed):

smalekta (approved):

pkamd (approved):
Tested the output using HIPBLASLT_LOG_MASK=128 and print_kernel_info in hipblaslt-bench. The 2 sets of info match.

PR #1391, opened 2024-11-26
Fix: abnormal terminal output if the first solution is rejected when printWinnerOnly is enabled

jichangjichang:
If the first solution is rejected when printWinnerOnly is enabled, terminal output would show NO_CHECK but not DID_NOT_SATISFY_ASSERTS.
And the winner time would be recorded as NaN, which makes the following winner cannot be printed out anymore.

### DIFF
--- a/tensilelite/Tensile/Source/client/include/LogReporter.hpp
+++ b/tensilelite/Tensile/Source/client/include/LogReporter.hpp
@@ -204,7 +204,7 @@ namespace TensileLite
             {
                 if(value == "PASSED" || value == "NO_CHECK")
                     m_rowLevel = LogLevel::Normal;
-                else if(value == "FAILED" || value == "FAILED_CONV")
+                else if(value == "FAILED")
                     m_rowLevel = LogLevel::Error;
                 else if(value == "WRONG_HARDWARE")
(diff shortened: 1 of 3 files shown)


KKyang (approved):

aazz44ss (approved):

jichangjichang:
tox pass except the random errors of swizzleA, which is not related to this PR.
> FAILED``` Tensile/Tests/common/test_config.py::test_config[Tensile/Tests/common/gemm/dtvA_swizzleA.yaml] - SystemExit: 1

> ========================================================================================= 1 failed, 79 passed, 37 skipped, 7 warnings in 9084.87s (2:31:24) ==========================================================================================

> py3: exit 1 (9085.15 seconds) /workspace/hipBLASLt/tensilelite> py.test -v --basetemp=/tmp/.tensile-tox/py3/tmp --junit-xml=/workspace/hipBLASLt/tensilelite/python_tests.xml --junit-prefix=py3 --color=yes -n 4 --prebuilt-client=/tmp/.tensile-tox/                                        py3/client/0_Build/client/tensile_client Tensile/Tests -m common `pid=439178

PR #1399, opened 2024-11-27
Fix stream-k kernels when K==0, allow beta*C calculation

AlexBrownAMD:
Fixes a bug causing problems with K=0 to end early. With this change if K==0 only main loop is skipped, and beta*C will be calculated.


neoblizz (dismissed):
Looks good!

bethune-bryant (dismissed):
Looks good.

tensilelite/Tensile/Source/lib/source/ContractionSolution.cpp
```diff
@@ -3028,6 +3030,15 @@ namespace TensileLite
                                           Hardware const& hardware,
                                           size_t          tiles) const
     {
+        // If K==0, run kernel as DP with Alpha=0 to skip main loop and apply beta*c
+        size_t z = 1;
+        for(size_t i = 0; i < problem.boundIndices().size(); ++i)
+        {
+            z *= problem.boundSize(i);
+        }
```
davidd-amd, line 3051: There is probably a way to do this in one line using `std::reduce` assuming there is a way to iterate over whatever `boundSize` refers to.

KKyang (dismissed):

tensilelite/Tensile/Source/lib/source/ContractionSolution.cpp
AlexBrownAMD: Thanks for the note! This patch needs to go with release changes, but I'll look at possible code cleanup in the next patch. This particular loop happens in a bunch of places in the code.

neoblizz (approved):

KKyang (approved):

PR #1410, opened 2024-11-28
Stream-k batch predict

AlexBrownAMD:
Update grid size prediction code to consider the problem's batch count. Without this change, strided-batch problems will likely run on fewer CUs than they should, since the batch dimension adds more tiles that can be spread across workgroups.


tensilelite/Tensile/Source/lib/source/ContractionSolution.cpp
```diff
@@ -60,24 +60,35 @@ namespace TensileLite
...
-            return m_tiles * n_tiles;
+            return m_tiles * n_tiles * batch;
         }
 
-        constexpr size_t num_fixup_peers(size_t BLK_K, size_t k, size_t iters_per_cta)
+        constexpr size_t num_fixup_peers(size_t BLK_K, size_t k, size_t iters_total, size_t iters_per_tile, size_t iters_per_cta, size_t g)
         {
-            return math::safe_ceil_div(math::safe_ceil_div(k, BLK_K), iters_per_cta);
+            // If tiles don't evenly divide there are always at least 2 fixup peers, and more if iters_per_tile > iters_per_cta
+            // size_t hasFixup = (iters_total % g == 0 && // Check if some WGs have more iters than others
```
neoblizz, line 88: Not sure I understand the `hasFixup`, is this supposed to make a more accurate prediction on how many fixup peers there actually are? And if so, why do we default back to the ceil div of iters per tile and iters per cta?

tensilelite/Tensile/Source/lib/source/ContractionSolution.cpp
```diff
@@ -720,11 +736,11 @@ namespace TensileLite
...
             // For now grouped gemm is not supported and passes nullptr
-            TENSILE_ASSERT_EXC(hardware != nullptr);
-            size_t cuCount = 0;
+            // TENSILE_ASSERT_EXC(hardware != nullptr);
+            // size_t cuCount = 0;
 
-            auto   tiles  = problem.getNumTiles(sizeMapping);
-            size_t skGrid = getSKGrid(problem, *hardware, tiles);
+            // auto   tiles   = problem.getNumTiles(sizeMapping);
+            // size_t skGrid  = getSKGrid(problem, *hardware, tiles);
```
neoblizz: Comments can be removed.

tensilelite/Tensile/Source/lib/source/ContractionSolution.cpp
```diff
@@ -60,24 +60,35 @@ namespace TensileLite
...
+        constexpr size_t number_of_output_tiles(size_t BLK_M, size_t BLK_N, size_t m, size_t n, size_t batch)
         {
             size_t m_tiles = math::safe_ceil_div(m, BLK_M);
             size_t n_tiles = math::safe_ceil_div(n, BLK_N);
-            return m_tiles * n_tiles;
+            return m_tiles * n_tiles * batch;
         }
 
-        constexpr size_t num_fixup_peers(size_t BLK_K, size_t k, size_t iters_per_cta)
+        constexpr size_t num_fixup_peers(size_t BLK_K, size_t k, size_t iters_total, size_t iters_per_tile, size_t iters_per_cta, size_t g)
```
neoblizz: Based on the comment above, do we need to update the signature of this function if its only relying on iters_per_tile and  iters_per_cta?

neoblizz (commented):
Looks good overall, some minor questions. Adds batch as a dimension for calculating output tiles, which should fix the calculations for total iters, etc. Makes grid prediction more accurate for batched compute.

bethune-bryant (commented):
I think this looks good. Once @neoblizz 's comments are addressed.

tensilelite/Tensile/Source/lib/source/ContractionSolution.cpp
AlexBrownAMD: Prior to this change, num_fixup_peers would return ceil(itersPerTile / itersPerCta). In the average case this is always 1 because an average workgroup does a bit more work than 1 tile (only case where it does less is a split-k type of problem size). If fixupPeers=1, that basically means no partials write or fixup step is happening, which is not actually the case for most grid sizes. FixupPeers is only 1 when the grid evenly divides the problem in to whole tiles. This logic checks if any tiles are split across workgroups by checking if all WGs have an even number of full tiles.
With this change, #FixupPeers becomes 2 in the average case, and is only 1 when there is no fixup/partials required. The changes are currently commented out, because the weights for b and d coefficients in the predictor may need to be adjusted since the change actually lowered average performance in a benchmarks.

tensilelite/Tensile/Source/lib/source/ContractionSolution.cpp
AlexBrownAMD: Fixed

tensilelite/Tensile/Source/lib/source/ContractionSolution.cpp
AlexBrownAMD: Good call, posted an update

neoblizz (approved):

KKyang (approved):

PR #1415, opened 2024-12-02
tensilelite:  executing asm-new.sh can be in different folder

jichangjichang:
When we handwrite kernel, we need to switch the same folder of asm-new.sh and execute asm-new.sh to recompile the handwritten kernel.
With this PR, we can execute asm-new.sh in arbitrary folder.


jichangjichang:
the failed case is in dtvA_swizzleA, which is random error and not related to this PR. 
FAILED Tensile/Tests/common/test_config.py::test_config[Tensile/Tests/common/gemm/dtvA_swizzleA.yaml]

hcman2 (approved):

KKyang (approved):

PR #1425, opened 2024-12-04
Add reject states for failing streamk params

daineAMD:


AlexBrownAMD (dismissed):
Looks good!

hcman2:
Do we have any solutions in the existing logic yaml which will be rejected by this condition? If yes, we should remove them in the same change.

jichangjichang (dismissed):

daineAMD:
> Do we have any solutions in the existing logic yaml which will be rejected by this condition? If yes, we should remove them in the same change.

After adding the last commit, I've checked and don't see any solutions which will be rejected by the condition. Will merge once CI passes.

KKyang (approved):

jichangjichang (approved):

PR #1428, opened 2024-12-05
Fix invalid string printed when running hipblaslt-test

NaveenElumalaiAMD:
bug fix

### DIFF
--- a/clients/benchmarks/client.cpp
+++ b/clients/benchmarks/client.cpp
@@ -447,7 +447,7 @@ try
 
         ("activation_type",
          value<std::string>(&activation_type)->default_value("none"),
-         "Options: None, gelu, relu")
+         "Options: none, gelu, relu")
 
         ("activation_arg1",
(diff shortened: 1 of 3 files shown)


clients/benchmarks/client.cpp
```diff
@@ -813,7 +813,7 @@ try
         throw std::invalid_argument("Invalid value for --initialization " + initialization);
 
     arg.activation_type = string_to_hipblaslt_activation_type(activation_type);
-    if(arg.activation_type == static_cast<hipblaslt_activation_type>(0))
+    if(arg.activation_type == static_cast<hipblaslt_activation_type>(-1))
```
jichangjichang, line 816: string_to_hipblaslt_activation_type would never return -1

NaveenElumalaiAMD: Thanks for catching that, fixed it in the latest commit.

babakpst (approved):

KKyang (approved):

PR #1429, opened 2024-12-05
Support conjugate-transpose as equivalent to transpose

daineAMD:
For real types, cublas/cublaslt/rocblas all support real-types with conjugate-transpose. In this case, it is treated as just transpose. This PR allows hipblasLT to treated conj-transpose as supported and is converted to transpose before passing off to Tensile. This makes integration with rocBLAS easier, and helps move toward equivalency with cublasLT.


AlexBrownAMD (dismissed):

hcman2:
Do we have any solutions in the existing logic yaml which will be rejected by this condition? If yes, we should remove them in the same change.

jichangjichang (approved):

KKyang (approved):

PR #1456, opened 2024-12-16
Refactor compile and link functions

bstefanuk:
Tensile and TensileCreateLibrary both invoke the function writeSolutionsAndKernels. This function is responsible for executing the "build recipe". That is, nearly all code related to code generation, compiling, linking, and bundling is contained in writeSolutionsAndKernels. This PR is a part of a larger parallelization change to TensileLt. Its objective is to improve the composability and testability of this important function.

**Outcomes:**
- Toolchains are created at the beginning of Tensile and TensileCreateLibrary, unifying structure and access to toolchain components.
- Toolchains have composable methods for general compile and link commands.
- processKernelSource is the code generation entry point for TensileLt. It now returns a named tuple, KernelCodeGenResult, which contains the generated file string and metadata on the code.
- processKernelSource is no longer responsible for writing assembly files or assembling/linking them in any way.
- KernelWriters are now only responsible for *generating source code* in the form of strings, no more hidden assembler invocations.

**Objectives:**
- [x] Remove getCompileArgs from KernelWriter
- [x] Remove \_getAssembledKernelObjectFile from KernelWriter
- [x] Remove getLinkCodeObjectArgs from KernelWriter
- [x] Remove \_getSingleCodeObjectFile from KernelWriter
- [x] Decouple the kernel writer from the link and compile commands altogether
- [x] Move compile and link flag setup currently in TensileInstructions.Utils to the Toolchain module
- [x] Remove the assemble step from KernelWriterAssembly and add it to AssemblyToolchain


tensilelite/Tensile/Toolchain/Assembly.py
```diff
@@ -0,0 +1,202 @@
+import collections
```
bstefanuk, line 25: Add copyright header comment

tensilelite/Tensile/Toolchain/Source.py
```diff
@@ -0,0 +1,191 @@
+import itertools
```
bstefanuk, line 25: Add copyright header comment

tensilelite/Tensile/Toolchain/Validators.py
```diff
@@ -1,5 +1,6 @@
 import os
```
bstefanuk, line 25: Add copyright header comment

tensilelite/Tensile/Utilities/Printing.py
bstefanuk, line 1: Mistaken new file, remove

tensilelite/Tensile/Toolchain/Assembly.py
```diff
@@ -0,0 +1,226 @@
...
+          "--target=amdgcn-amd-amdhsa", 
+          f"-mcode-object-version={coVersion}", 
+          f"-mcpu={cpu}",  
+          wavefrontFlag, 
+          "-g" if debug else "",
+          "-c", 
+          "-o", destPath, srcPath
+      ]
+      print2(f"Assembling object files: {' '.join(args)}")
+      subprocess.check_call(args)
```
davidd-amd: This pattern shows up in each function. Can we have one function that handles passing a string to subprocess, printing diagnostics and error handling? Then make the other functions responsible for generating the string representing the command?

tensilelite/Tensile/Toolchain/Assembly.py
```diff
@@ -0,0 +1,226 @@
...
+          self.assembler, 
+          "-x", "assembler", 
+          "--target=amdgcn-amd-amdhsa", 
+          f"-mcode-object-version={coVersion}", 
+          f"-mcpu={cpu}",  
+          wavefrontFlag, 
+          "-g" if debug else "",
+          "-c", 
+          "-o", destPath, srcPath
+      ]
```
davidd-amd: much of this does not change between calls to the function. Does is make sense to separate out the stuff that can be set up front from the stuff that varies? This will reduce the number of parameters required to call the function.

tensilelite/Tensile/Toolchain/Assembly.py
```diff
@@ -0,0 +1,226 @@
...
+        args = [
+            self.bundler,
+            "--compress",
+            "--type=o",
+            "--bundle-align=4096",
+            f"--targets=host-x86_64-unknown-linux,hipv4-amdgcn-amd-amdhsa--{gfx}",
+            "--input=/dev/null",
+            f"--input={srcPath}",
+            f"--output={destPath}",
+        ]
```
davidd-amd, line 142: When we have these things buried in a function they will inevitably change which violates the open/close principle.

tensilelite/Tensile/Toolchain/Assembly.py
```diff
@@ -0,0 +1,226 @@
...
+      archKernelMap[tuple(k['ISA'])].append(k)
+
+    coFiles = []
+    for arch, archKernels in archKernelMap.items():
+      if len(archKernels) == 0:
+        continue
+
+      gfx = getGfxName(arch)
+
+      if globalParameters["MergeFiles"] or globalParameters["NumMergedFiles"] > 1 or globalParameters["LazyLibraryLoading"]:
```
davidd-amd: This will need to be updated when we remove merge files

tensilelite/Tensile/Toolchain/Assembly.py
```diff
@@ -0,0 +1,226 @@
...
+          if compress:
+            toolchain.compress(str(coFileRaw), str(coFile), gfx)
+          else:
+            shutil.move(coFileRaw, coFile)
+
+          coFiles.append(coFile)
+      else:
+        # no mergefiles
+        def newCoFileName(kName):
+          if globalParameters["PackageLibrary"]:
```
davidd-amd: `PackageLibrary` removal is already in develop so this probably needs to be removed.

tensilelite/Tensile/Toolchain/Assembly.py
```diff
@@ -0,0 +1,226 @@
...
+
+          coFile = destDir / coFileRaw.name.replace(extCoRaw, extCo)
+          if compress:
+            toolchain.compress(str(coFileRaw), str(coFile), gfx)
+          else:
+            shutil.move(coFileRaw, coFile)
+
+          coFiles.append(coFile)
+      else:
+        # no mergefiles
```
davidd-amd: Probably already here but do we need the comment? If so we should update it to "no mergefiles AND no lazy loading"

tensilelite/Tensile/Toolchain/Assembly.py
```diff
@@ -0,0 +1,226 @@
...
+        def newCoFileName(kName):
+          if globalParameters["PackageLibrary"]:
+            return os.path.join(destDir, gfx, kName + '.co')
+          else:
+            return os.path.join(destDir, kName + '_' + gfx + '.co')
+
+        def orgCoFileName(kName):
+          return os.path.join(asmDir, kName + '.co')
+
+        for src, dst in Utils.tqdm(((orgCoFileName(kName), newCoFileName(kName)) for kName in \
```
davidd-amd: If we are going to use tqdm we should use it conditionally so the log files are cleaner.

tensilelite/Tensile/Toolchain/Source.py
```diff
@@ -0,0 +1,217 @@
...
+            globalParameters: A dictionary of global parameters.
+
+        Raises:
+            RuntimeError: If the compilation command fails.
+        """
+        launcher = shlex.split(os.environ.get("Tensile_CXX_COMPILER_LAUNCHER", ""))
+
+        hipFlags = [
+            "-D__HIP_HCC_COMPAT_MODE__=1",
+            "--cuda-device-only",
```
davidd-amd: Just an observation but we can now use `--offload-device-only`

tensilelite/Tensile/Toolchain/Source.py
```diff
@@ -0,0 +1,217 @@
...
+            "-I", includePath,
+            "-Xoffload-linker", f"--build-id={self.buildIdKind}",
+            "-std=c++17",
+        ]
+        if self.asanBuild:
+            hipFlags.extend(["-fsanitize=address", "-shared-libasan", "-fuse-ld=lld"])
+        if self.saveTemps:
+            hipFlags.append("--save-temps")
+        if os.name == "nt":
+            hipFlags.extend(["-fms-extensions", "-fms-compatibility", "-fPIC", "-Wno-deprecated-declarations"])
```
davidd-amd, line 94: Seems like we should know much of this stuff up front and can lift it out of this function.

tensilelite/Tensile/Toolchain/Validators.py
```diff
@@ -1,5 +1,30 @@
...
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+#
+################################################################################
+
 import os
 import re
+import shlex
```
davidd-amd: what are we using from this module?

tensilelite/Tensile/TensileCreateLibrary.py
```diff
@@ -69,38 +71,44 @@ def wrapper(*args, **kwargs):
...
+    asmFilename = kernelWriter.getKernelFileBase(kernel)
+    err, src = kernelWriter.getSourceFileString(kernel)
+    header = kernelWriter.getHeaderFileString(kernel)
+    # will be put in Kernels.h/cpp if None
+    objFilename = kernel._state.get("codeObjectFile", None)
+    isa = kernelWriter.isa
+    wavefrontSize = kernelWriter.wavefrontSize
 
-    return (err, src, header, kernelName, filename)
+    return KernelCodeGenResult(err, src, header, asmFilename, objFilename, isa, wavefrontSize)
```
davidd-amd: Does the wavefrontsize change? Do we have one kernelWriter or many?

tensilelite/Tensile/Toolchain/Assembly.py
bstefanuk: coVersion can be moved from this function into the toolchain ctor since it's specified at the program-level. The other two are more complicated since they are extracted from state on the kernelWriter. The cpu variable (which I will rename, but is actually the gfx name of an ISA) is needed to be configurable at assemble time to support multiple archs. Which would be similar for wavefront size.

tensilelite/Tensile/Toolchain/Assembly.py
bstefanuk: Okay. Question is then, where does the buck stop? Should I make a parameter for `--bundle-align`, `--type`, do I make every parameter configurable? 

By doing this we're pre-emptively guessing about possible future states of our code. If something changes then we'll have to change the code anyway, either here or at the call site. Just my two cents.

tensilelite/Tensile/Toolchain/Source.py
bstefanuk: I'm not sure I understand what your vision of these functions are. Eventually, we must translate a boolean to a string literal. Why not leave compiler-specific setup right where we create the compiler string? If we lift them out of this function, where do we put them?

tensilelite/Tensile/Toolchain/Validators.py
bstefanuk: This should be a static analysis check.

tensilelite/Tensile/TensileCreateLibrary.py
bstefanuk: Given the existing code, yes, wavefront size will change. Whether it does in practice, I don't know. See comment above about wavefront size.

tensilelite/Tensile/Toolchain/Assembly.py
davidd-amd: yeah it becomes a question of what is application code and what is library code and where do we want stability

tensilelite/Tensile/Toolchain/Source.py
davidd-amd: Every time we call this function we are doing redundant work. I would like to only do it once.

PR #1464, opened 2024-12-18
gfx12 - enlarging tolerance is not needed

cmingch:


jichangjichang (approved):
Is that possible to remove the special tolerance for gfx11 in the future?

cmingch:
> Is that possible to remove the special tolerance for gfx11 in the future?

it's hw limitation. in order to save space and power, 2's compliment design is not standard in gfx11, when a negation is needed, this accuracy loss will happen.

Serge45 (approved):

PR #1469, opened 2024-12-19
Provide unsurprising behavior for tensile threads

davidd-amd:
On two occasions now I've helped users untangle surprising behavior when attempting to pass parallelization settings to TensileCreateLibrary. This PR corrects some of the issues that cause the surprising behavior such as a cap at 64 threads. We have users that want to request 128 and 256 parallel jobs. 

Having made these changes, we need to confirm that the CI configuration is still throttling to avoid OOM issues (see https://github.com/ROCm/rocJenkins/pull/919). The rocJenkins PR was merged so we should be ok.

FYI @loggrado @amd-aakash


jichangjichang:
should we notify build team to limit the number of parallel jobs at their end?

davidd-amd:
> should we notify build team to limit the number of parallel jobs at their end?

I don't think this will be a problem because we are using cpu nodes for builds.

PR #1474, opened 2024-12-19
Gfx942 80cu grid based and equality tuning for HHS NN

smalekta:
- Made HHS_NN grid denser for certain regions of interest by (1) adding new tuned points (2) adding additional points (pointing to existing solution indices such that any point is mapped to its previous solution index to prevent regressions due to the newly added sizes)
- Add an equality tuning for an important size for HHS_NN


aferoz21 (approved):
Looks good.

babakpst (approved):

PR #1487, opened 2024-12-25
Using i32*u32->i64 when removing staggerU offset

briannwu:
Fix: memory access fault in tail loop with extreme large size
sgprGlobalReadIncs may be over int32 positive range for extreme large size,
it may cause the wrong MSB if using int32*int32 calculation.

### DIFF
--- a/tensilelite/Tensile/KernelWriterAssembly.py
+++ b/tensilelite/Tensile/KernelWriterAssembly.py
@@ -4032,7 +4032,7 @@ class KernelWriterAssembly(KernelWriter):
       # might be able to refactor this to eliminate signed math
       imod.add(SSubI32(dst=sgpr(tmp), src0=3 if kernel["PrefetchGlobalRead"] else 2, \
               src1=sgpr("StaggerUIter")))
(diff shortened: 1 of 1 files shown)


KKyang (approved):

jichangjichang (approved):

hcman2:
What if the i32 is -2,147,483,648? (the smallest)

briannwu:
@hcman2 
I think it gets the correct result if smallest int32 * largest uint32.
if s27(int32) * s64(uint32) -> s[28:29](int64)
(gdb) p/x $s27
$5 = 0x80000000
(gdb) p/d $s27
$6 = -2147483648 (int32 smallest)
(gdb) p/x $s64
$7 = 0xffffffff
(gdb) p/u $s64
$8 = 4294967295 (uint32 largest)
(gdb) p/x $s28
$9 = 0x80000000
(gdb) p/x $s29
$10 = 0x80000000
{9BB438A6-AD87-448F-AE93-F6B7EDC81BFD}

briannwu:
{A0E850CD-03A8-4FBF-B9D8-C5D5870DFA16}
{D341D474-925A-4EBF-AD53-34C7737BBCD9}

PR #1490, opened 2024-12-25
Add ENABLE_MAKER to the option set defined at Tensilelite's cmake configuration.

tsenwang:
If the ENABLE_MARKER option is not enabled, Maker functionality cannot be activated through TensileCreateLibraryFiles().


vin-huang (approved):
LGTM

jichangjichang (approved):

PR #1494, opened 2024-12-26
Remove some Equality problems.

hcman2:
Remove the sizes that gridbased is better.


jichangjichang (commented):
you can use merge.py to remove the solutions that don't has any size mapping.

jichangjichang (approved):

AndySu12 (approved):

PR #1498, opened 2024-12-27
Update NT/NN BBS Equality yaml.

hcman2:
Update NT/NN BBS Equality yaml for some Large sizes.


jichangjichang (dismissed):

AndySu12 (approved):

AndySu12 (dismissed):
The config should be SAV instead of SAB.

AndySu12 (approved):

jichangjichang (approved):

PR #1507, opened 2025-01-03
Feature/refactor compile and link fxns

davidd-amd:
As a side effect of doing this work, we discovered that for gfx942 builds we were spawning processes to build 35k duplicate kernels. By refactoring the code we got the 942 builds down from 100+ minutes to about 75 minutes.

Tensile and TensileCreateLibrary both invoke the function writeSolutionsAndKernels. This function is responsible for executing the "build recipe". That is, nearly all code related to code generation, compiling, linking, and bundling is contained in writeSolutionsAndKernels. This PR is a part of a larger parallelization change to TensileLt. Its objective is to improve the composability and testability of this important function.

**Outcomes:**
- Toolchains are created at the beginning of Tensile and TensileCreateLibrary, unifying structure and access to toolchain components.
- Toolchains have composable methods for general compile and link commands.
- processKernelSource is the code generation entry point for TensileLt. It now returns a named tuple, KernelCodeGenResult, which contains the generated file string and metadata on the code.
- processKernelSource is no longer responsible for writing assembly files or assembling/linking them in any way.
- KernelWriters are now only responsible for *generating source code* in the form of strings, no more hidden assembler invocations.

**Objectives:**
- [x] Remove getCompileArgs from KernelWriter
- [x] Remove \_getAssembledKernelObjectFile from KernelWriter
- [x] Remove getLinkCodeObjectArgs from KernelWriter
- [x] Remove \_getSingleCodeObjectFile from KernelWriter
- [x] Decouple the kernel writer from the link and compile commands altogether
- [x] Move compile and link flag setup currently in TensileInstructions.Utils to the Toolchain module
- [x] Remove the assemble step from KernelWriterAssembly and add it to AssemblyToolchain

Original changes were made by bstefanuk, but the final merge conflicts were resolved by ellosel.


KKyang (dismissed):

jichangjichang (dismissed):

tensilelite/Tensile/TensileInstructions/Code.py
```diff
@@ -770,9 +770,9 @@ def __str__(self):
         kStr += "    KernArgsVersion: %d\n"%self.kernArgsVersion
         kStr += "amdhsa.version:\n"
         kStr += "  - 1\n"
-        if self.codeObjectVersion == 4:
+        if self.codeObjectVersion == "4" or self.codeObjectVersion == "default":
```
bstefanuk, line 773: "default" isn't a support argument via the command line. Why are we adding it here as an option?

davidd-amd: Because codeObjectVersion can be set via a yaml file.

jichangjichang (dismissed):

KKyang (approved):

bstefanuk (approved):

davidd-amd:
The tests that are failing were previously commented out in rocJenkins. When tests were restored, they added everything back. So they just patched rocJenkins again to prevent these tests from running.

PR #1528, opened 2025-01-09
fix frequency monitor

wenchuanchen:
rsmi_dev_metrics_xcd_counter_get gets xcd_counter=0 by chance which leads to segmentation fault in the following instruction.
image


hcman2:
Is this a workaround or fix?

wenchuanchen:
this is a fix to protect xcd_count=0 case

Serge45 (approved):

jichangjichang (approved):

PR #1536, opened 2025-01-10
Set debug flag as default

cliffxzx:
When profiling a kernel that was recompiled without the debug flag, it shows the old source file because the debug symbols haven't been updated.

image

Expected:
image


cliffxzx:
Found `ASM_ARGS` for adding `-g` flag.

PR #1563, opened 2025-01-16
Deprecate AMDGPU_TARGETS variable

lawruble13:
Summary of proposed changes:

- Change the CMake to set the GPU_TARGETS variable from AMDGPU_TARGETS if not set by the user
- Change usages of AMDGPU_TARGETS to GPU_TARGETS

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -170,6 +170,10 @@ else()
       )
     endif()
 
+    if(AMDGPU_TARGETS AND NOT GPU_TARGETS)
+      message( DEPRECATION "AMDGPU_TARGETS use is deprecated. Use GPU_TARGETS." )
+    endif()
+
     if (AMDGPU_TARGETS)
         set(TMPAMDGPU_TARGETS "${AMDGPU_TARGETS}")
         if(TMPAMDGPU_TARGETS STREQUAL "all" )


CMakeLists.txt
```diff
@@ -170,6 +170,10 @@ else()
       )
     endif()
 
+    if(AMDGPU_TARGETS AND NOT GPU_TARGETS)
+      message( DEPRECATION "AMDGPU_TARGETS use is deprecated. Use GPU_TARGETS." )
+    endif()
+
     if (AMDGPU_TARGETS)
```
jichangjichang, line 177: Should you modify this section as well?

jichangjichang (commented):
AMDGPU_TARGETS also appears in install.sh and cmake/Verbose.cmake

CMakeLists.txt
lawruble13: The intention is that AMDGPU_TARGETS should still work, it's just marked as deprecated, so I don't want to change this block. Additionally, we're trying to move away from "all" as an accepted target (as it doesn't convey the same thing across all components, and what "all" means is not clearly defined), so I don't want to update that processing.

jichangjichang:
Is this change required for 6.4?

jichangjichang:
@lawruble13  close PR since there is no more staging build for release-staging/rocm-rel-6.4 branch. If this is required for 6.4, please go cherry-pick request for 6.4 RC branch.

PR #1570, opened 2025-01-20
hipblaslt-bench: throw error if c_type is not equal to d_type

jichangjichang:
####  Command:
hipblaslt-bench   --a_type bf16_r --b_type bf16_r --c_type bf16_r --transA T --transB N -j 1 -i 1

#### Original output: (no solution found without any reason)
error: NO solution found! at /workspace/hipBLASLt/clients/benchmarks/../include/testing_matmul.hpp:2534
#### New output: (notify user that configuration is incorrect)
Invalid: --c_type bf16_r is not equal to **--d_type f16_r**

### DIFF
--- a/clients/benchmarks/client.cpp
+++ b/clients/benchmarks/client.cpp
@@ -780,6 +780,11 @@ try
     if(arg.d_type == HIPBLASLT_DATATYPE_INVALID)
         throw std::invalid_argument("Invalid value for --d_type " + d_type);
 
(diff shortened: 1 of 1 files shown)


Serge45 (approved):

ssuyuanchang (approved):

KKyang (approved):

jichangjichang:
Change doesn't affect hipblaslt-test and hipblaslt library and tenslielite.

PR #1590, opened 2025-01-23
Pin toolchain environment variables from configure->build time.

stellaraccident:
Tensile uses a number of environment variables for controlling how it finds/uses its toolchain. However, there is no reliable way to pass these to hipBLASLt configure and have them stick at build time. This leads to all kinds of fallback paths on /opt/rocm and other issues.

This patch:

* Fixes the Tensile add_custom_commands to launch commands in an `env` with critical variables set.
* Uses project-consistent environment variables to find clang/assembler vs in two scripts vs hardcoding to /opt/rocm or requiring a ROCM_PATH.
* Adds CMake cache variables `Tensile_ROCM_OFFLOAD_BUNDLER_PATH` and `Tensile_ROCM_ASSEMBLER_PATH` to explicitly allow controlling these paths via configuration.
* Adds a `Tensile_TOOLCHAIN_FLAGS` env var and CMake setting. These flags will be prepended to any C/C++ compiler invocations and are used for explicit control of hip and bitcode library toolchain options (i.e. avoids spurious fallback to search heuristics that are less precise).


stellaraccident:
Hold for an update. Found an issue I need to roll in a fix for.

stellaraccident:
Rebased to mainline and reworked the patch substantially. Closing this one in favor of https://github.com/ROCm/hipBLASLt/pull/1615

PR #1596, opened 2025-01-24
reinterpreted structs to match

TorreZuk:
* test matching struct for algo API


daineAMD (approved):
Looks good to have needed parity between rocblaslt_matmul_algo and hipblasLtMatmulAlgo_t

TorreZuk:
rocblaslt_matmul_heuristic_result contains this rocblaslt_matmul_algo struct (and it's hipblaslt interface structs do the same) so that is the real problem if they don't match as reinterpret cast is used.

jichangjichang (approved):

PR #1597, opened 2025-01-24
Build system improvements

davidd-amd:
The hipblaslt build system was inherited from rocblas. However, the physical design of the project is different. In particular, tensile is embedded in hipblalt, rather than being a separate project. Nevertheless, all of the machinery to install tensile into a python virtual environment is still in the hipblaslt build system. This has several consequences:

- Changes to Tensile source files are not observable by the build system so when making a change and attempting to rebuild `make` is unable to detect that anything changes.
- The Tensile source files consumed by the hipblaslt build system are those installed into the virtual environment so after making a change to Tensile one must force a reconfigure and build which currently requires completely rebuilding the tensile gemm libraries and the cpp interface which is very time consuming.
- The behavior of the build system is not conventional so users with idomatic CMake experience are confused or surprised.

This PR reconciles these issues by taking initial steps to properly include tensile into the hipblaslt build system. To begin, Tensile is no longer pip installed through the `execute_process` API. Rather, we use and `add_subdirectory` call which ensures that source file changes in Tensile are observable and executing `make` in the build directory will retrigger a rebuild of the necessary targets. Further, `ALL` is removed from the custom target for generating the gemm libraries with `TensileCrateLibrary` so that they are no longer unconditionally rebuilt every time `make` is executed.  These changes will significantly reduce build times for developers that are only working on the cpp interfaces in hipblaslt.

In addition, we made some changes to source files to resolve compiler warnings.

We will include documentation updates for building hipblaslt.


cmake/virtualenv.cmake
```diff
@@ -26,7 +26,7 @@ function(virtualenv_create)
 endfunction()
 
 function(virtualenv_install)
-    virtualenv_create()
+virtualenv_create()
```
davidd-amd, line 29: restore whitespace

CMakeLists.txt
```diff
@@ -82,7 +82,7 @@ include(cmake/Dependencies.cmake)
 
 # Append our library helper cmake path and the cmake path for hip (for convenience)
 # Users may override HIP path by specifying their own in CMAKE_MODULE_PATH
-list( APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake  ${ROCM_PATH}/lib/cmake/hip /opt/rocm/lib/cmake/hip ${HIP_PATH}/cmake )
+list( APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake  ${ROCM_PATH}/lib/cmake/hip /opt/rocm/lib/cmake/hip ${HIP_PATH}/cmake ${CMAKE_CURRENT_SOURCE_DIR}/tensilelite/Tensile/cmake)
```
powderluv, line 85: lets not include /opt/rocm anywhere. It will cause hard fails but it is easy to fix.

CMakeLists.txt
```diff
@@ -153,12 +153,11 @@ if (LINK_BLIS)
 endif()
 
 if (USE_CUDA)
-	find_package( CUDA REQUIRED )
+	  find_package( CUDA REQUIRED )
```
powderluv, line 156: whitespace change ?

CMakeLists.txt
davidd-amd: In what way does it cause a hard failure?

PR #1599, opened 2025-01-27
Make global soln indices deterministic

b-shi:
Currently, global solution indices can change between different builds. This PR sorts the global solution indices to make them deterministic.

### DIFF
--- a/tensilelite/Tensile/TensileCreateLibrary/Run.py
+++ b/tensilelite/Tensile/TensileCreateLibrary/Run.py
@@ -135,7 +135,7 @@ def writeAssembly(asmPath: Union[Path, str], result: KernelCodeGenResult):
     with open(path, "w", encoding="utf-8") as f:
       f.write(result.src)
       del result # result.src is very large so let gc know to clean up asap
- 
+
     return path, isa, wfsize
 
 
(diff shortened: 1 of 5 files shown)


KKyang:
What happens when a solution is removed in the middle?

b-shi:
> What happens when a solution is removed in the middle?

Do you mean when we delete duplicate solutions? In that case it should fine, since, afaik, that step is done in serial. The non-determinism comes from creating the master library in parallel.

KKyang:
> > What happens when a solution is removed in the middle?
> 
> Do you mean when we delete duplicate solutions? In that case it should fine, since, afaik, that step is done in serial. The non-determinism comes from creating the master library in parallel.

No, I mean if a solution is replaced by a faster solution, or it is removed due to new optimizations that makes another existing kernel faster. You noticed that indices change between builds and you want to make them deterministic.

tensilelite/Tensile/TensileCreateLibrary/Run.py
```diff
@@ -341,11 +341,31 @@ def libraryIter(lib: MasterSolutionLibrary):
...
-        matchTable[s.index] = [srcFile, localIdx]
+
+  if args["GenSolTable"]:
+    matchTable = {}
+    # Match yaml file solutions to solution index
+    for _,masterLibrary in masterLibraries.items():
+      for localIdx, _, s in libraryIter(masterLibrary):
+        matchTable[s.index] = [s.srcName, localIdx]
+    LibraryIO.write("MatchTable", matchTable)
+
```
yenong-amd: Should the MatchTable be written out after solution sorting?

b-shi: Ah, good catch, thanks!

b-shi:
> > > What happens when a solution is removed in the middle?
> > 
> > 
> > Do you mean when we delete duplicate solutions? In that case it should fine, since, afaik, that step is done in serial. The non-determinism comes from creating the master library in parallel.
> 
> No, I mean if a solution is replaced by a faster solution, or it is removed due to new optimizations that makes another existing kernel faster. You noticed that indices change between builds and you want to make them deterministic.

Oh, I meant for the same commit (no changes to library logic files) if we rebuild multiple times the global solution indices may be different. We can see this by checking the contents of `MatchTable.yaml` it will change whenever we rebuild.

KKyang (approved):

yenong-amd (approved):

PR #1605, opened 2025-01-30
clear normal operation hip error code

TorreZuk:
* this pattern should be backward and forward hip compatible


daineAMD (approved):
Looks good to me to clear the error code here.

jichangjichang (approved):

PR #1615, opened 2025-02-04
Add explicit CMake flags for specifying Tensile toolchain components.

stellaraccident:
The legacy tensile settings are a mess of things using bare tool names and trying to use heuristics to find them on a variety of paths. All of this code needs to go.

However, as a starting point, I introduce new CMake flags:

* Tensile_HIP_CONFIG
* Tensile_CXX_COMPILER
* Tensile_C_COMPILER
* Tensile_OFFLOAD_BUNDLER
* Tensile_ASSEMBLER
* Tensile_TOOLCHAIN_FLAGS

Those that can be passed as arguments to TensileCreateLibrary are done that way. Unlike prior revisions, if any of the component names are absolute paths, it short circuits the auto-detection heuristics and uses exactly what it was told. This is an essential characteristic for hermetic builds (I am building in an environment where the compiler is built hermetically and the build system needs to specify where the pieces are).

Note that there is some coupling here between gen_assembly.sh and TensileCreateLibrary which is unfortunate. The former must have its configuration passed by environment, so the CMake custom commands now do this properly. The latter would require a lot of plumbing to get access to Tensile_TOOLCHAIN_FLAGS and Tensile_CXX_COMPILER_LAUNCHER, so for the moment, those are pulled from the environment. A future refactor should add them as TensileCreateLibrary arguments explicitly and pass them through. There are just so many layers of loose arguments to get to _compileSourceObjectFile that I opted for doing this in two steps (it would be a substantial improvement if the toolchain options were passed around as an object vs as half a dozen independent variables).


stellaraccident:
No longer needed with next-cmake

PR #1617, opened 2025-02-04
Add log to roctx

KKyang:


library/src/amd_detail/rocblaslt/include/Debug.hpp
```diff
@@ -82,6 +104,7 @@ namespace rocblaslt
         int         m_value;
         int         m_value2;
         bool        m_printMarker       = false;
+        bool        m_printLog          = false;
```
jichangjichang: could you rename it as  printBenchMarker?
printLog cannot reveal its purpose.

jichangjichang (approved):

hcman2 (approved):

PR #1629, opened 2025-02-06
MBSK prefetch optimization

boringmorning:
Add tuning parameter "MbskPrefetchOpt" to control the MBSK reduction order:

-1: Select between 0/1 based on # store elements.

0:  Fetch from workgroup dim -> elements dim. This is the original design. (default)

1: Fetch from elements dim -> workgroup dim. Has better prefetch pattern when # store elements is large.

### DIFF
--- a/tensilelite/Tensile/Common/GlobalParameters.py
+++ b/tensilelite/Tensile/Common/GlobalParameters.py
@@ -1264,6 +1264,11 @@ validParameters = {
     # True:  Use WaveSplitK. Number of WorkGroup[2] threads in the same wave compute the same output elements (matrix D) along different unroll indices.
     #        The local sum from those threads are reduced through suffling using VALU instructions. Currently only support dot2 kernel.
     "WaveSplitK": [False, True],
+    # Control mbsk reduction prefetch order
+    # -1 : Select between 0/1 based on # store elements.
+    # 0  : Fetch from workgroup dim -> elements dim. (default)
+    # 1  : Fetch from elements dim -> workgroup dim. Has better prefetch pattern when # store elements is large.
+    "MbskPrefetchOpt": [-1, 0, 1],
 }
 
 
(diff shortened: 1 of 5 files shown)


boringmorning:
image
image
In these 2 benchmarks, I tuned ~2000 kernels each for **MbskPrefetchOpt** = 0 and **MbskPrefetchOpt** = 1 and compare their winners' performance (all other parameters settings are the same). For the benchmark with smaller size, the result are almost the same. In contrast, **MbskPrefetchOpt** = 1 resulted in noticable improvement with the larger sizes.

KKyang:
why another new parameter? Can't we make this auto calculated? (Calculated in runtime and use a flag to control the branch)

boringmorning:
> why another new parameter? Can't we make this auto calculated? (Calculated in runtime and use a flag to control the branch)

@KKyang This is an option, but how to decide the branching is an important problem. Also will result in larger code size (hundreds to thousand lines of code). 
Note that this is just the draft version, and one reason to use a new parameter is for easier benchmarking. I will have a presentation & discussion about this optimization (maybe next week).

boringmorning:
Did another benchmark to test MbskPrefetchOpt=-1. I choose 500 sizes whose solution is with MBSK from gfx942/GridBased/aquavanjaram_Cijk_Alik_Bljk_HHS_BH_Bias_HAS_SAV_UserArgs.yaml and benchmark them with MbskPrefetchOpt = 0 & -1. With -1, 269 sizes' solutions use MbskPrefetchOpt = 1 and have 103.11% Gflops improvement in average.

KKyang:
A new problem arise that we have too much kernel because of these fork parameters and cause the library getting bigger and bigger (also increases the getHeuristic search time). Comparing to each kernel itself already contains thousands or even ten thousand lines of code, it's always better than duplicating them.

hcman2:
Is it possible to set MbskPrefetchOpt=-1 forever. Then the number of kernels will be the same as before.

boringmorning:
> Is it possible to set MbskPrefetchOpt=-1 forever. Then the number of kernels will be the same as before.

@KKyang do you mean something like this?  In this way we can remove this fork parameter. I'm just worried about that there might be some performance regression issue, since I can't benchmark all kernels & problems.

jichangjichang (dismissed):
Please wait until gfx950 init PR gets merged.

hcman2 (approved):

geotseng-amd (approved):

PR #1650, opened 2025-02-11
Add swish activation

chengchingwen:
This PR add Swish activation in tensilelite and enable SiLU (Swish with beta=1) in hipblaslt.

The [swish paper v1](https://arxiv.org/abs/1710.05941v1) refer `x * sigmoid(x)` (SiLU) as Swish and `x * sigmoid(beta * x)` as Swish_beta, but [swish paper v2](https://arxiv.org/abs/1710.05941v2) directly call the one with beta as Swish activation. [Jax](https://flax.readthedocs.io/en/v0.5.3/_autosummary/flax.linen.swish.html) is also using the v1 definition.

This PR only enable SiLU in hipblastlt but the activation type is named `swish`. If the Swish with beta is needed in the future, it can be done by taking the activation argument and calling the swish activation in tensilelite, so we can avoid having both `EPILOGUE_SWISH` and `EPILOGUE_SILU`.

ticket: https://ontrack-internal.amd.com/browse/SWDEV-512620


clients/samples/27_gemm_swish_bias/sample_hipblaslt_gemm_swish_bias.cpp
```diff
@@ -0,0 +1,209 @@
...
+                                                          matC,
+                                                          matD,
+                                                          pref,
+                                                          request_solutions,
+                                                          heuristicResult,
+                                                          &returnedAlgoCount));
+
+    if(returnedAlgoCount == 0)
+    {
+        std::cerr << "No valid solution found!" << std::endl;
```
jichangjichang, line 173: free allocated resource before return

chengchingwen: This is copied from 04_gemm_bias and also happened in other samples like 08_gemm_gelu_aux_bias. Should I also fix them in this PR?

amd-jnovotny (dismissed):
Docs changes ok.

KKyang:
You can just set beta = 1 in tensile_host.cpp when using silu.

chengchingwen:
@KKyang Can you elaborate? I'm not sure what you mean.

library/src/amd_detail/rocblaslt/include/rocblaslt-types.h
```diff
@@ -167,7 +167,9 @@ typedef enum rocblaslt_epilogue_
     ROCBLASLT_EPILOGUE_DGELU         = 192,
     ROCBLASLT_EPILOGUE_DGELU_BGRAD   = 208,
     ROCBLASLT_EPILOGUE_BGRADA        = 256,
-    ROCBLASLT_EPILOGUE_BGRADB        = 512
+    ROCBLASLT_EPILOGUE_BGRADB        = 512,
+    ROCBLASLT_EPILOGUE_SWISH         = 1024,
```
jichangjichang, line 171: Please shift the enum number to a larger one in case that has confliction with cublaslt enum in the future.
Besides, we use _EXT postfix for hipblaslt specific extension.
ex: ROCBLASLT_EPILOGUE_SWISH_EXT   = 2^16
ROCBLASLT_EPILOGUE_SWISH_BIAS_EXT   = 2^16 + 4

tensilelite/Tensile/Source/lib/include/Tensile/ContractionProblemPredicates.hpp
```diff
@@ -2080,7 +2080,9 @@ namespace TensileLite
                     if(value == ActivationType::Hipblaslt_all
                        && (problem.activationType() == ActivationType::DGelu
                            || problem.activationType() == ActivationType::Gelu
-                           || problem.activationType() == ActivationType::Relu))
+                           || problem.activationType() == ActivationType::Relu
+                           || problem.activationType() == ActivationType::Silu
+                           || problem.activationType() == ActivationType::Swish))
```
jichangjichang: hipblaslt_all should only build in Silu bases on your description of the requriement.

clients/samples/27_gemm_swish_bias/sample_hipblaslt_gemm_swish_bias.cpp
jichangjichang: it would be better if you can fix them together.

library/include/hipblaslt.h
```diff
@@ -79,7 +79,9 @@ typedef enum {
   HIPBLASLT_EPILOGUE_DGELU = 192,         /**<Apply gradient GELU transform. Requires additional aux input. */
   HIPBLASLT_EPILOGUE_DGELU_BGRAD = 208,   /**<Apply gradient GELU transform and bias gradient to the results. Requires additional aux input. */
   HIPBLASLT_EPILOGUE_BGRADA = 256,        /**<Apply bias gradient to A and output gemm result. */
-  HIPBLASLT_EPILOGUE_BGRADB = 512         /**<Apply bias gradient to B and output gemm result. */
+  HIPBLASLT_EPILOGUE_BGRADB = 512,        /**<Apply bias gradient to B and output gemm result. */
+  HIPBLASLT_EPILOGUE_SWISH = 1024,        /**<Apply Swish point-wise transform to the results (x:=Swish(x, 1)).*/
+  HIPBLASLT_EPILOGUE_SWISH_BIAS = 1028,   /**<Apply Bias and then Swish transform.*/
```
jichangjichang: Please shift the enum number to a larger one in case that has confliction with cublaslt enum in the future.
Besides, we use _EXT postfix for hipblaslt specific extension.
ex: HIPBLASLT_EPILOGUE_SWISH_EXT= 2^16
HIPBLASLT_EPILOGUE_SWISH_BIAS_EXT= 2^16 + 4

jichangjichang:
> You can just set beta = 1 in tensile_host.cpp when using silu.

silu doesn't have beta.

library/include/hipblaslt.h
chengchingwen: Does this also apply to ROCBLASLT_EPILOGUE_SWISH?

KKyang:
> > You can just set beta = 1 in tensile_host.cpp when using silu.
> 
> silu doesn't have beta.

We can share the same code by setting beta = 1 as silu, so we don't have to change the code again.

chengchingwen:
> We can share the same code by setting beta = 1 as silu, so we don't have to change the code again.

On the hipblaslt side it won't need much change, just taking the activation argument into account. On the tensilelite side, I'm not sure if we want to remove the existing Silu assembly. Having two assembly would help applying different optimization if needed, but that will depend on future requests.

KKyang:
> > We can share the same code by setting beta = 1 as silu, so we don't have to change the code again.
> 
> On the hipblaslt side it won't need much change, just taking the activation argument into account. On the tensilelite side, I'm not sure if we want to remove the existing Silu assembly. Having two assembly would help applying different optimization if needed, but that will depend on future requests.

Assembly code is long enough, so better to share the same code if possible.

library/include/hipblaslt.h
jichangjichang: rocblaslt enum value should be the same with hipblaslt.
For the postfix _EXT, I don't think we need to add that in ROCBLASLT enum.

jichangjichang:
> > > We can share the same code by setting beta = 1 as silu, so we don't have to change the code again.
> > 
> > 
> > On the hipblaslt side it won't need much change, just taking the activation argument into account. On the tensilelite side, I'm not sure if we want to remove the existing Silu assembly. Having two assembly would help applying different optimization if needed, but that will depend on future requests.
> 
> Assembly code is long enough, so better to share the same code if possible.

SWISH asm code is not in enabled in hipblaslt lib. I don't think we need to go SWISH asm code with beta =1.  Silu asm code should be enough for current requirement.

library/include/hipblaslt.h
chengchingwen: https://github.com/ROCm/hipBLASLt/blob/0cb1d8dd70935e904bc16cc671104b1227982342/library/src/amd_detail/rocblaslt/src/utility.cpp#L211

It seems some other enums in rocblaslt do have _EXT?

tensilelite/Tensile/Tests/common/gemm/fp16_tn.yaml
```diff
@@ -216,6 +216,8 @@ BenchmarkProblems:
         - BiasTypeArgs: ['h']
         - ActivationArgs:
           - [Enum: relu]
+          - [Enum: silu]
```
jichangjichang, line 219: Should we add "silu" test on ActivationType: hipblaslt_all?

chengchingwen: Where should I add it?

chengchingwen: I added silu to gemm_bias_strided_batched.yaml

jichangjichang (approved):

vin-huang (approved):

PR #1657, opened 2025-02-11
gfx942 80cu equality tuning for HHS NN/TN

minsukim-amd:
58 GEMM sizes tuned and merged with average ~40% improvement on reliable system. 7 GEMM we cannot improve sufficient performance.


minsukim-amd:
Close as duplicated

PR #1679, opened 2025-02-17
Add rocisa

KKyang:
This is a initial commit. Creating a module where we can put slow Python code here as C code to speed up the execution time.

Most of the code in the current Tensile generator will remain unchanged until roadmap 2.

Currently supports

1. TensileInstruction structures fully implemented

[Roadmap]

1. Adding optimization algorithm functions that accept instruction modules in rocisa
2. Porting generator stuff to rocisa

[Changes]

1. Entry point for tensilelite has changed from Python to shell script
2. Tensile has renamed to TensilePy


KKyang:
[==========] 53567 tests from 13 test suites ran. (2029563 ms total)
[  PASSED  ] 53567 tests.
hipBLASLt version: 1300

KKyang:
--------------------------------------------------- generated xml file: /data0/yangwen/hipBLASLt/tensilelite/python_tests.xml ----------------------------------------------------
============================================================ 87 passed, 36 skipped, 2 warnings in 4316.21s (1:11:56) =============================================================
py3: exit 0 (4316.41 seconds) /data0/yangwen/hipBLASLt/tensilelite> py.test -v --basetemp=/tmp/CI_out/py3/tmp --junit-xml=/data0/yangwen/hipBLASLt/tensilelite/python_tests.xml --junit-prefix=py3 --color=yes -n 4 --prebuilt-client=/tmp/CI_out/py3/client/0_Build/client/tensile_client Tensile/Tests -m common pid=823519
  py3: OK (4516.75=setup[10.01]+cmd[0.48,189.86,4316.41] seconds)
  congratulations :) (4516.82 seconds)

aazz44ss (dismissed):

vin-huang (dismissed):

aazz44ss (dismissed):

vin-huang (dismissed):

bstefanuk (dismissed):
Please contact @ellosel and myself to discuss these changes prior to merging. This code change will cause issues with the new build algorithm.

davidd-amd (changes requested):
Please do not merge this. This will introduce major conflicts in existing branches. There must be another way to solve the problem without reverting.

davidd-amd:
@KKyang what slow python code are you referring to? Is the goal to port all of the python code to c++? The solution to improving the amount of time spent in try assembler is to only run it on the architecture that you are targeting. We've already done this work in another feature branch and it is much faster without having to port to c++.

KKyang:
> @KKyang what slow python code are you referring to? Is the goal to port all of the python code to c++? The solution to improving the amount of time spent in try assembler is to only run it on the architecture that you are targeting. We've already done this work in another feature branch and it is much faster without having to port to c++.

No, the ultimate goal is to make the generator ported to C.

KKyang:
> Please contact @ellosel and myself to discuss these changes prior to merging. This code change will cause issues with the new build algorithm.

I remove the changes to the Common submodule.

KKyang:
============================================================ 87 passed, 36 skipped, 3 warnings in 4348.86s (1:12:28) =============================================================
py3: exit 0 (4349.13 seconds) /data0/yangwen/hipBLASLt/tensilelite> py.test -v --basetemp=/tmp/CI_out/py3/tmp --junit-xml=/data0/yangwen/hipBLASLt/tensilelite/python_tests.xml --junit-prefix=py3 --color=yes -n 4 --prebuilt-client=/tmp/CI_out/py3/client/0_Build/client/tensile_client Tensile/Tests -m common pid=2773076
  py3: OK (4512.14=setup[12.81]+cmd[0.39,149.82,4349.13] seconds)
  congratulations :) (4512.19 seconds)

aazz44ss (dismissed):

vin-huang (dismissed):

KKyang:
[==========] 53567 tests from 13 test suites ran. (1947265 ms total)
[  PASSED  ] 53567 tests.
hipBLASLt version: 1300

CMakeLists.txt
```diff
@@ -235,6 +235,44 @@ else()
...
+      RESULT_VARIABLE result
+    )
+    if(NOT ${result} EQUAL 0)
+      message(STATUS "Output: ${output}")
+      message(STATUS "Error: ${error}")
+      message(FATAL_ERROR "Result: ${result}")
+    else()
+      message(STATUS "rocIsa building done.")
+      set(ENV{PYTHONPATH} "${VIRTUALENV_HOME_DIR}/rocisa/build/lib:$ENV{PYTHONPATH}")
+    endif()
```
davidd-amd: Is this cmake code invoking python to build c? Why not do this natively in cmake?

tensilelite/Tensile/ClientWriter.py
```diff
@@ -97,8 +100,17 @@ def main(config, cxxCompiler: str, cCompiler: str, outputPath: Path):
...
 
+  # Get rocIsa path
+  module_path = os.path.dirname(inspect.getfile(rocisa))
+  env = os.environ.copy()
+  if 'PYTHONPATH' in env:
+    if not module_path in env['PYTHONPATH']:
+        env["PYTHONPATH"] = module_path + ":" + env["PYTHONPATH"]
+  else:
+    env["PYTHONPATH"] = module_path
+
```
davidd-amd: If we are going to move to this model of having c++/python library with bindings we should deal with these things on the cmake side using out of the box tools. Presumably nanobind has cmake support. I know that pybind11 does and nanobind claims to be a drop in replacement.

tensilelite/Tensile/Hardware.py
```diff
@@ -29,12 +29,12 @@
...
-        gfxArch = isaToGfx(isa)
+        gfxArch = isaToGfx(tuple(isa))
         return cls("AMDGPU", value=cls("Processor", value=gfxArch))
 
     @classmethod
     def FromHardware(cls, isa, cuCount=None):
-        gfxArch = isaToGfx(isa)
+        gfxArch = isaToGfx(tuple(isa))
         if cuCount == None:
             return cls("AMDGPU", value=cls("Processor", value=gfxArch))
```
ellosel, line 39:
```suggestion
        gfxArch = isaToGfx(tuple(isa))
        if cuCount == None:
            return HardwarePredicate.FromISA(cls, isa)
```

tensilelite/Tensile/Hardware.py
```diff
@@ -29,12 +29,12 @@
...
         else:
```
davidd-amd, line 40:
```suggestion
        if cuCount == None:
            return HardwarePredicate.FromISA(cls, isa)
        else:
               gfxArch = isaToGfx(tuple(isa))
```

tensilelite/Tensile/Tests/common/test_config.py
```diff
@@ -25,8 +25,12 @@
 import os
 import pytest
 import subprocess
+import sys
 import yaml
 
+module_build_dir = os.path.normpath(os.path.join(os.path.dirname(os.path.realpath(__file__)), "..", "..", "..", "..", "rocisa", "build"))
+sys.path.append(module_build_dir + "/lib")
+
```
davidd-amd: We should iterate on adding more robust cmake support to the project in general which should preclude the need of code like this.

tensilelite/Tensile/bin/BuildRocIsa.py
davidd-amd, line 1: Lets add this to the list of things that can be done in cmake rather than writing python code to invcoke cmake.

davidd-amd (dismissed):
Overall I think this aligns with what we think could be a viable solution to improving build times in the code. However, migrating from python to C/C++ is a big endeavor and we don't have any idea what to expect for performance improvements. This is something we have thought about and have several ideas we should discuss.

In the meantime, I think it would behoove us all to make improvements to the project cmake support, add testing to the existing code and then develop and plan to make this migration before we get too far along.

CMakeLists.txt
KKyang: ~~This is step is to copy the folder to virtual env and run subprocess to force rocisa build in configuration stage. If you use cmake, it'll build in the next stage and cause failure when running TensileCreateLibrary.~~

I misunderstood what you're asking, this is actually cmake, the python above is to get the python version in order to build the right module version for current python.

tensilelite/Tensile/ClientWriter.py
KKyang: This is because the entry point is from tensilelite's Tensile/bin/Tensile instead of cmake. At this point, the rocisa is built, in order to pass the path to the subprocess, we have the code here to get the module's folder path.

tensilelite/Tensile/Tests/common/test_config.py
KKyang: Same reason above.

tensilelite/Tensile/bin/BuildRocIsa.py
KKyang: The entry point is Python, you have to call subprocess anyway. Unless we change how we use tensilelite with config yaml.

KKyang:
> Overall I think this aligns with what we think could be a viable solution to improving build times in the code. However, migrating from python to C/C++ is a big endeavor and we don't have any idea what to expect for performance improvements. This is something we have thought about and have several ideas we should discuss.
> 
> In the meantime, I think it would behoove us all to make improvements to the project cmake support, add testing to the existing code and then develop and plan to make this migration before we get too far along.

Moving from Python to C is because everytime we'd like to write some mlir-like optimizations such as iterating through objects will take much longer than C code, that's why we are trying to get things fix instead of sticking on Python.

KKyang:
Anyway, most of the changes above is to prevent the change of the existing usage. Unless you have a script on top of rocisa and Tensile to perform the two steps cmake, build, and run Tensile, you have to run subprocess inside Python to get the module build.

tensilelite/Tensile/ClientWriter.py
```diff
@@ -97,8 +100,17 @@ def main(config, cxxCompiler: str, cCompiler: str, outputPath: Path):
...
 
+  # Get rocIsa path, remove this when subprocess is removed
+  module_path = os.path.dirname(inspect.getfile(rocisa))
+  env = os.environ.copy()
+  if 'PYTHONPATH' in env:
+    if not module_path in env['PYTHONPATH']:
+        env["PYTHONPATH"] = module_path + ":" + env["PYTHONPATH"]
+  else:
+    env["PYTHONPATH"] = module_path
+
```
KKyang, line 111: We cannot remove this since it's calling a subprocess.

tensilelite/Tensile/Hardware.py
KKyang: Will skip this suggestion as this is not a part of the commit.

tensilelite/Tensile/Tests/common/test_config.py
KKyang: Removed in updated commit

tensilelite/Tensile/bin/BuildRocIsa.py
KKyang: Fixed

tensilelite/Tensile/bin/Tensile
```diff
@@ -1,5 +1,4 @@
-#!/usr/bin/env python3
-
+#!/bin/bash
```
KKyang, line 1: Changed from Python to shell script. CMake and Python is seperated and the module path is pased to the Python though PYTHONPATH.

b-shi:
> > Overall I think this aligns with what we think could be a viable solution to improving build times in the code. However, migrating from python to C/C++ is a big endeavor and we don't have any idea what to expect for performance improvements. This is something we have thought about and have several ideas we should discuss.
> > In the meantime, I think it would behoove us all to make improvements to the project cmake support, add testing to the existing code and then develop and plan to make this migration before we get too far along.
> 
> Moving from Python to C is because everytime we'd like to write some mlir-like optimizations such as iterating through objects will take much longer than C code, that's why we are trying to get things fix instead of sticking on Python.

With this transition to C, will code gen still be using asm?

KKyang:
> > > Overall I think this aligns with what we think could be a viable solution to improving build times in the code. However, migrating from python to C/C++ is a big endeavor and we don't have any idea what to expect for performance improvements. This is something we have thought about and have several ideas we should discuss.
> > > In the meantime, I think it would behoove us all to make improvements to the project cmake support, add testing to the existing code and then develop and plan to make this migration before we get too far along.
> > 
> > 
> > Moving from Python to C is because everytime we'd like to write some mlir-like optimizations such as iterating through objects will take much longer than C code, that's why we are trying to get things fix instead of sticking on Python.
> 
> With this transition to C, will code gen still be using asm?

Yes, we're just replacing functions and classes used to generate the asm files.

tensilelite/Tensile/bin/Tensile
```diff
@@ -24,16 +23,25 @@
...
 
-    from Tensile import Tensile
+WORKING_DIR=$2
+ROCISA_LIB_DIR=$(relative2absolute $WORKING_DIR)/rocisa/lib
+SCRIPT_DIR=$(dirname "$(readlink -f "$0")")
+ROCISA_DIR=$SCRIPT_DIR/../../rocisa
+### Build rocisa
+# Currently the path to rocisa is fixed
+mkdir -p $WORKING_DIR/rocisa
+cmake -DCMAKE_CXX_COMPILER=/opt/rocm/bin/amdclang++ -DCMAKE_PREFIX_PATH=$(python3 -m nanobind --cmake_dir) -DCMAKE_BUILD_TYPE=Debug -DPYTHON_VERSION=$(python3 --version 2>&1 | cut -d ' ' -f 2) -S $ROCISA_DIR -B $WORKING_DIR/rocisa
```
jichangjichang: rocm may not be in /opt/rocm/
Please refer to ROCM_PATH

KKyang: fixed

tensilelite/tox.ini
```diff
@@ -8,17 +8,26 @@ labels =
...
 setenv =
     TENSILE_CLIENT_STATIC = {env:TENSILE_CLIENT_STATIC}
+    PYTHONPATH = {envdir}/rocisa/lib
 commands =
+    mkdir -p {envdir}/rocisa
+    sh -c "cmake -S rocisa -B {envdir}/rocisa -DCMAKE_CXX_COMPILER=/opt/rocm/bin/amdclang++ -DCMAKE_PREFIX_PATH=$(python3 -m nanobind --cmake_dir)"
+    cmake --build {envdir}/rocisa -j 8
     pip install --upgrade pip
-    python3 ./Tensile/bin/Tensile Tensile/Tests/build_client.yaml {envdir}/client
+    python3 ./Tensile/bin/TensilePy Tensile/Tests/build_client.yaml {envdir}/client
```
jichangjichang, line 25: what is TensilePy?

tensilelite/rocisa/tox.ini
```diff
@@ -0,0 +1,21 @@
...
+deps =
+    cmake
+    -r{toxinidir}/requirements.txt
+    nanobind
+    pytest>=5.4.1
+setenv =
+    PYTHONPATH = {envdir}/rocisa/lib
+commands =
+    mkdir -p {envdir}/rocisa
+    sh -c "cmake -S . -B {envdir}/rocisa -DCMAKE_CXX_COMPILER=/opt/rocm/bin/amdclang++ -DCMAKE_PREFIX_PATH=$(python3 -m nanobind --cmake_dir)"
```
jichangjichang: Don't hard code rocm path.

KKyang: this is a fallback in case the env is not found

KKyang: Oh sorry, wrong file

KKyang: Solved

vin-huang (approved):

aazz44ss (approved):

tensilelite/rocisa/include/base.hpp
```diff
@@ -0,0 +1,262 @@
...
+namespace nb = nanobind;
+
+namespace rocisa
+{
+    struct KernelInfo
+    {
+        IsaVersion isaVersion;
+        int        wavefront;
+        KernelInfo() = default;
+        KernelInfo(const IsaVersion& isaVersion, int wavefront)
```
davidd-amd, line 50: Why are we taking `IsaVersion` by const ref? We wind up making a copy. Is that reasonable? Can we move it in? Can we take a forwarding reference?

tensilelite/rocisa/include/base.hpp
```diff
@@ -0,0 +1,262 @@
...
+        {
+        }
+    };
+
+    struct IsaInfo
+    {
+        std::map<std::string, int>  asm_caps;
+        std::map<std::string, bool> arch_caps;
+        std::map<std::string, int>  reg_caps;
+        std::map<std::string, bool> asm_bugs;
```
davidd-amd, line 62: Do we need ordered maps here?

KKyang: These changes are just to prevent modifying too much code and having too much conflict in Python. Nanobind will convert this to the original IsaInfo format.

tensilelite/rocisa/include/base.hpp
```diff
@@ -0,0 +1,262 @@
...
+#include <typeinfo>
+
+#include "hardware_caps.hpp"
+#include "helper.hpp"
+
+namespace nb = nanobind;
+
+namespace rocisa
+{
+    struct KernelInfo
```
davidd-amd, line 45: Does this type preserve triviality? Should it?

tensilelite/rocisa/include/base.hpp
```diff
@@ -0,0 +1,262 @@
...
+
+    struct IsaInfo
+    {
+        std::map<std::string, int>  asm_caps;
+        std::map<std::string, bool> arch_caps;
+        std::map<std::string, int>  reg_caps;
+        std::map<std::string, bool> asm_bugs;
+    };
+
+    class rocIsa
```
davidd-amd, line 65: Will the instance of `rocIsa` ever get passed by const ref? If so, are we const correct?

KKyang: What const ref?

tensilelite/rocisa/include/base.hpp
```diff
@@ -0,0 +1,262 @@
...
+        void setData(const std::map<IsaVersion, IsaInfo>& data)
+        {
+            m_isainfo = data;
+        }
+
+    private:
+        rocIsa()
+        {
+            // trigger random seed for other components
+            srand(time(NULL));
```
davidd-amd, line 160: Do we care about reproducibility? If we are using the current time to seed a RNG then we've lost reproducibility.

Is rand a random number generator for use in production code? Please see the random header:

https://en.cppreference.com/w/cpp/header/random

KKyang: It is used to generate unique labels in asm kernels like in Python.

tensilelite/rocisa/include/base.hpp
KKyang: No. Maybe we should.

tensilelite/rocisa/include/base.hpp
KKyang: It cannot be pickable now.

tensilelite/rocisa/include/base.hpp
```diff
@@ -0,0 +1,262 @@
...
+        std::map<std::string, bool> asm_bugs;
+    };
+
+    class rocIsa
+    {
+    public:
+        // Delete copy constructor and assignment operator
+        rocIsa(const rocIsa&)            = delete;
+        rocIsa& operator=(const rocIsa&) = delete;
+
```
davidd-amd, line 71: What about move? Should we be able to move the instance? If not, we should explicitly delete move semantics as well. If we expect to move an instance then we've broken move by deleting copy sematics. See the rulle of 3/5/0:

https://en.cppreference.com/w/cpp/language/rule_of_three

tensilelite/rocisa/include/base.hpp
davidd-amd, line 45: Where are your tests for this class?

tensilelite/rocisa/include/base.hpp
davidd-amd, line 65: Tests?

KKyang: These passed the existing tox and hipblaslt-test, but no unit test yet.

tensilelite/rocisa/include/base.hpp
```diff
@@ -0,0 +1,262 @@
...
+    private:
+        rocIsa()
+        {
+            // trigger random seed for other components
+            srand(time(NULL));
+        }
+
+        static void construct()
+        {
+            instance = new rocIsa();
```
davidd-amd, line 165: Does this leak?

tensilelite/rocisa/include/base.hpp
```diff
@@ -0,0 +1,262 @@
...
+
+        std::mutex                            m_mutex;
+        std::map<std::thread::id, KernelInfo> m_threads;
+        std::map<IsaVersion, IsaInfo>         m_isainfo;
+
+        static rocIsa*        instance;
+        static std::once_flag initInstanceFlag;
+    };
+
+    struct Item
```
davidd-amd, line 176: Tests?

tensilelite/rocisa/include/base.hpp
KKyang: No, checked with Calgrind

KKyang: Nanobind will also report leaks as well.

tensilelite/rocisa/include/base.hpp
davidd-amd, line 65: RAII?

tensilelite/rocisa/include/base.hpp
KKyang: It will not be generated by default because the copy constructor is deleted.

tensilelite/rocisa/include/base.hpp
KKyang: Let's change to Meyer's then

tensilelite/rocisa/README.md
davidd-amd, line 1: This README needs work. There are no examples or usage patterns. How is the project consumed in Tensile build system? How do I import this module? How does one contribute? Also, do we have performance metrics?

tensilelite/rocisa/src/base.cpp
```diff
@@ -0,0 +1,212 @@
...
+    std::string isaToGfx(const nb::tuple& arch)
+    {
+        return getGfxNameTuple(
+            IsaVersion{nb::cast<int>(arch[0]), nb::cast<int>(arch[1]), nb::cast<int>(arch[2])});
+    }
+
+    std::string isaToGfx(const IsaVersion& arch)
+    {
+        return getGfxNameTuple(arch);
+    }
```
davidd-amd, line 42: Why do we need to do this on the C++ side? We have a simple function in python for this why not expose it to c++ using nanobind?

tensilelite/rocisa/include/base.hpp
davidd-amd: The question is should it and if not we should follow the rule of five where if you define any then define all of them.

tensilelite/rocisa/include/base.hpp
davidd-amd: Who is responsible for cleaning up instance? I don't see it getting deleted?

davidd-amd:
Nanobind is a nice tool. I am a big fan of pybind11 and nanobind is by the same author. However, the docs suggest that nanobind only works with python 3.8+. Have you tested with python 3.6?

CMakeLists.txt
```diff
@@ -235,6 +235,44 @@ else()
...
```
davidd-amd, line 275: This is not going to pass muster for users down stream of us that are responsible for building and deploying hipblaslt.

tensilelite/rocisa/include/base.hpp
davidd-amd: If TensileCreateLibrary and Tensile fails we leak and see numerous nanobind leak warnings.

CMakeLists.txt
```diff
@@ -235,6 +235,44 @@ else()
...
+    # else()
+      message(STATUS "rocIsa configuring done.")
+    # endif()
+    execute_process(
+      COMMAND make -j8
+      WORKING_DIRECTORY ${ROCISA_VIRTUAL_DIR}/build
+      OUTPUT_VARIABLE output
+      ERROR_VARIABLE error
+      RESULT_VARIABLE result
+    )
```
bstefanuk, line 267: This forces the user to use Make, this isn't necessarily correct if the user sets a different generator, like Ninja. It also breaks portability across different build environments. Perhaps instead we call `cmake --build <dir> -- -j8`.

Also, I would consider baking the build command directly into CMake to be non-idiomatic and would instead expect the build step to be separated from the configuration step.

tensilelite/Tensile/Activation.py
```diff
@@ -24,12 +24,15 @@
 import math
 import struct
 from collections import OrderedDict
-from enum import IntFlag
+from copy import deepcopy
+from enum import Enum, IntFlag
+from typing import List, Union
+
+from rocisa import rocIsa
+from rocisa.enum import *
```
bstefanuk, line 32: Wildcard imports are discouraged in Python, instead, we should spell out the items we need explicitly. Why? Namespace pollution, readability, obscured dependencies.

tensilelite/rocisa/include/code.hpp
```diff
@@ -0,0 +1,1613 @@
...
+          | | |--GlobalReadInst bogusGlobalReadInst                                // comments
+          |--Inst bogusInst                                          // comments
+          |--Module "Module-lvl2"
+          | |--Inst bogusInst                                          // comments
+          | |--Module "Module-lvl3"
+          | | |--TextBlock
+          | | | |--bogusTextBlock
+          | | | |--bogusTextBlock2
+          | | | |--bogusTextBlock3
+          | | |--GlobalReadInst bogusGlobalReadInst
```
davidd-amd, line 392: ?

tensilelite/requirements.txt
```diff
@@ -7,4 +7,4 @@ joblib>=1.1.1; python_version < '3.8'
 simplejson
 ujson
 orjson
-distro
```
davidd-amd, line 10: This will cause the static build tests to fail

tensilelite/rocisa/CMakeLists.txt
```diff
@@ -0,0 +1,65 @@
...
+    if(result)
+        message(FATAL_ERROR "Failed to install nanobind via pip: ${error}")
+    endif()
+    execute_process(
+        COMMAND python3 -m nanobind --cmake_dir
+        OUTPUT_VARIABLE nanobind_DIR
+        ERROR_VARIABLE error
+        RESULT_VARIABLE result
+    )
+endif()
```
davidd-amd, line 33: We should not install dependencies via cmake.

tensilelite/rocisa/CMakeLists.txt
```diff
@@ -0,0 +1,65 @@
+cmake_minimum_required(VERSION 3.15)
```
davidd-amd, line 1: This is not consistent with the required cmake version in hipblaslt - should it be?

tensilelite/rocisa/CMakeLists.txt
```diff
@@ -0,0 +1,65 @@
...
+project(rocisa LANGUAGES CXX)
+set(CMAKE_CXX_STANDARD 20)
```
davidd-amd, line 3: Why are we hard coding the standard? What C++ standards are we required to support?

tensilelite/rocisa/CMakeLists.txt
```diff
@@ -0,0 +1,65 @@
...
+
+set(ROCISAINST_SOURCE src/instruction/instruction.cpp
+                      src/instruction/common.cpp
+                      src/instruction/branch.cpp
+                      src/instruction/cmp.cpp
+                      src/instruction/cvt.cpp
+                      src/instruction/mem.cpp
+                      src/instruction/mfma.cpp)
+
+include_directories(${CMAKE_SOURCE_DIR}/include)
```
davidd-amd, line 57: use `target_include_directories`

tensilelite/rocisa/CMakeLists.txt
davidd-amd, line 1: No copyright info

tensilelite/rocisa/CMakeLists.txt
```diff
@@ -0,0 +1,65 @@
...
+string(REPLACE "\n" "" nanobind_DIR "${nanobind_DIR}")
+
+if(DEFINED PYTHON_VERSION AND PYTHON_VERSION)
+    message(STATUS "Manually set Python version to ${PYTHON_VERSION}")
+else()
+    set(PYTHON_VERSION 3)
+endif()
+
+# Use cmake -Dpybind11_DIR=$(python3 -m nanobind --cmakedir) .. to set pybind dir
+# Use cmake -DCMAKE_PREFIX_PATH=$(python3 -m nanobind --cmake_dir) .. to set nanobind
```
davidd-amd, line 44: So we can use pybind11 or nanobind?

tensilelite/rocisa/CMakeLists.txt
```diff
@@ -0,0 +1,65 @@
...
+    )
+endif()
+message(STATUS "Nanobind path: ${nanobind_DIR}")
+string(REPLACE "\n" "" nanobind_DIR "${nanobind_DIR}")
+
+if(DEFINED PYTHON_VERSION AND PYTHON_VERSION)
+    message(STATUS "Manually set Python version to ${PYTHON_VERSION}")
+else()
+    set(PYTHON_VERSION 3)
+endif()
```
davidd-amd, line 41: Why not make this a configurable cache variable with a default?

tensilelite/rocisa/include/base.hpp
bnemanich: We should be using a seed that can be reproduced each time. Either hardcode the seed or pass in as a value.

tensilelite/rocisa/CMakeLists.txt
davidd-amd: We shouldn't assume our users want us to set their environment for them.

tensilelite/rocisa/src/main.cpp
```diff
@@ -0,0 +1,45 @@
...
+void init_containers(nb::module_ m);
+void init_label(nb::module_ m);
+void init_enum(nb::module_ m);
+void init_inst(nb::module_ m);
+void init_code(nb::module_ m);
+void init_count(nb::module_ m);
+
+NB_MODULE(rocisa, m)
+{
+    m.doc() = "Module rocisa.";
```
bstefanuk, line 37: Improve doc-string?

tensilelite/rocisa/include/base.hpp
```diff
@@ -0,0 +1,262 @@
...
+        {
+            return 0;
+        }
+    };
+
+    std::string isaToGfx(const nb::tuple& arch);
+    std::string isaToGfx(const IsaVersion& arch);
+    std::string getGlcBitName(bool hasGLCModifier);
+    std::string getSlcBitName(bool hasGLCModifier);
+} // namespace base
```
bstefanuk, line 262: namespace rocisa

davidd-amd:
This project needs a setup.py file.

CMakeLists.txt
davidd-amd: We should provide a setup.py and `pip install rocisa/`

tensilelite/rocisa/include/base.hpp
```diff
@@ -0,0 +1,262 @@
...
+        }
+
+        void setKernel(const nb::tuple& arch, const int wavefrontSize)
+        {
+            std::thread::id id = std::this_thread::get_id();
+            IsaVersion      isaVersion
+                = {nb::cast<int>(arch[0]), nb::cast<int>(arch[1]), nb::cast<int>(arch[2])};
+            m_mutex.lock();
+            m_threads[id] = std::move(KernelInfo(isaVersion, wavefrontSize));
+            m_mutex.unlock();
```
bstefanuk, line 106: Use an RAII alternative like std::lock_guard?

tensilelite/rocisa/include/base.hpp
```diff
@@ -0,0 +1,262 @@
...
+        {
+        }
+
+        virtual ~Item() = default;
+
+        virtual std::shared_ptr<Item> clone() const
+        {
+            throw std::runtime_error("clone() not implemented");
+            return nullptr;
+        }
```
bstefanuk, line 190: What's the rational for not having this implemented? Should this have a base implementation?

tensilelite/rocisa/include/base.hpp
bstefanuk, line 176: Consider moving Item into its own file.

tensilelite/rocisa/src/base.cpp
```diff
@@ -0,0 +1,212 @@
...
+
+    std::string getSlcBitName(bool hasGLCModifier)
+    {
+        if(hasGLCModifier)
+            return "slc";
+        return "sc1";
+    }
+
+    // Force init the instance
+    auto isInit = rocIsa::getInstance().isInit();
```
bstefanuk, line 59: Why are we forcing the instantiation here instead of on-demand from the calling code?

tensilelite/rocisa/include/base.hpp
KKyang: Merged in https://github.com/ROCm/hipBLASLt/pull/1793

tensilelite/rocisa/include/base.hpp
KKyang: This is a port from Python, you shouldn't have porting and code modification at the same time. Anyway, I'll let it stay as default rather than setting it.

tensilelite/rocisa/include/base.hpp
KKyang: > Who is responsible for cleaning up instance? I don't see it getting deleted?

It will show still reachable and will be collected by the system. This is the pre-C++11 singleton. Anyway, I changed to Meyer's singlton instead. https://github.com/ROCm/hipBLASLt/pull/1793

tensilelite/rocisa/include/base.hpp
KKyang: https://github.com/ROCm/hipBLASLt/pull/1793

tensilelite/rocisa/README.md
KKyang: https://github.com/ROCm/hipBLASLt/pull/1811
https://github.com/ROCm/hipBLASLt/pull/1810

tensilelite/rocisa/src/base.cpp
KKyang: I thought I talked about this before, the goal is to move Python to C++, not importing symbols from Python to C++.

KKyang:
> Nanobind is a nice tool. I am a big fan of pybind11 and nanobind is by the same author. However, the docs suggest that nanobind only works with python 3.8+. Have you tested with python 3.6?

Currently none of the new images and conda is using Python 3.6.

tensilelite/rocisa/CMakeLists.txt
KKyang: Because the virtual env setup in hipBLASLt is not working properly.

tensilelite/rocisa/CMakeLists.txt
KKyang: we can use libraries from pip install.

tensilelite/rocisa/CMakeLists.txt
KKyang:
```
auto dst, dst2 = somefunction;
auto lamfunc = [&dst]()....
```
This is a C++20 extension

CMakeLists.txt
KKyang: Our project can't even run with ninja. I tried to change to ninja before.

CMakeLists.txt
KKyang: proof?

tensilelite/rocisa/include/code.hpp
KKyang: Don't just leave a question mark here and say nothing. Apparently, this is the comment from TensileInstruction in case if you don't know.

tensilelite/requirements.txt
KKyang: https://github.com/ROCm/hipBLASLt/pull/1812
Fixed here

tensilelite/rocisa/src/base.cpp
KKyang: It's a singleton.

tensilelite/rocisa/include/base.hpp
KKyang: Not now, this is only the port from TensileInstruction to rocisa. PR's should have it's goal and a clear topic instead of having new features and code refactoring at the same time. This will make your PR extremely large and easy to have conflict with others.

tensilelite/rocisa/include/base.hpp
KKyang: This is used to make the shared_ptr vector in module works properly when deepcopy is called from Python. Else it will become shallow copies.

tensilelite/rocisa/CMakeLists.txt
KKyang: Thanks

tensilelite/rocisa/CMakeLists.txt
KKyang: Not very necessary to change it. hipBLASLt is forced to use 3.25.2 is because a bug and has to use this version or above. Either we can keep this the minimum requirement or align with the hipBLASLt cmake file.

CMakeLists.txt
KKyang: Using cmake --build is a good idea
https://github.com/ROCm/hipBLASLt/pull/1814

tensilelite/rocisa/CMakeLists.txt
KKyang: https://github.com/ROCm/hipBLASLt/pull/1815

tensilelite/rocisa/include/base.hpp
bnemanich: So do the kernels currently have different symbol names each time you run the generator?

CMakeLists.txt
bstefanuk: Typically, how this works is that the onus is on the implementer to prove the reviewer wrong.

tensilelite/rocisa/src/base.cpp
bstefanuk: Yes... I understand that it's a singleton. That doesn't predicate the need for early initialization. What I'm asking is, why don't we leave it up to the calling code to decide when it's initialized and allow for lazy initialization? If there's a reasonable argument either direction, that's fine, I'm flexible, I would just like clarification on this decision.

CMakeLists.txt
stellaraccident: The correct thing to do here is to do a sub-cmake project and pass CMAKE_GENERATOR from the parent to child projet -- not just hand-code your own build system in cmake.

Please revert this patch. It will not work in various Linux and Windows build environments.

Please include me on future re-architecture patches to CMake that impact the build system. We spend significant time fixing the build for hipBLASLt, and we need a higher bar for this kind of development since you are pushing significant fragility and code quality issues to your users.

tensilelite/rocisa/include/base.hpp
stellaraccident: Please revert this patch. You cannot be generating instable sources.

tensilelite/rocisa/include/base.hpp
stellaraccident: Please revert and structure the submission to include unit tests.

CMakeLists.txt
stellaraccident: CMAKE_GENERATOR

tensilelite/Tensile/Activation.py
stellaraccident: +1

CMakeLists.txt
KKyang: @stellaraccident currently the Tensile client is built inside the TensileLite Python generator, and it is triggered in the middle of the cmake generation.
The first version of the build flow is that I also embedded rocisa into Tensile so no problem at all. However, David requested me to separate and build before the Python is launched instead.

Using a sub-cmake project here will make the project being built after the cmake configuration, and the whole flow has to be changed in order to make the project being built and add to the PYTHONPATH before TensileCreateLibrary starts. Correct me if I'm wrong.

stellaraccident: A couple of points:

* This kind of architecture refactoring is extremely hard to review and make progress on when done monolithically like this. Typically, if changing the build graph in such a dramatic way, a small patch is done to make the structural change and then a larger NFC patch is done for the logic porting. The months of latency and objections to this patch would have been much easier and quicker if the changes had been staged in a more incremental fashion. 
* You may not be wrong in the ultimate goal, but this patch is wrong on many details, and I guarantee you that next week, when the patch lands in my queue, this will cost me personally four hours undoing and fixing the many details that were glossed over in review and landed without the consent of the reviewers. Almost all of this pain could have been averted if the change was done in a more incremental sequence.
* Iiuc your comment, you are planning to move all of the tensilelite stuff to the build phase. That is the right thing to do, and it should have been done first. This patch is going in the wrong direction on things that will break for users.

I recommend reverting and reapplying in a more incremental sequence that breaks the build structure changes from the logic changes. And I would further recommend that a proper review flow is done: the reviewers seem to understand what needs to be done to the build but were largely ignored in the comments. This isn't how we land changes.

PR #1682, opened 2025-02-19
streamk - adding memory fallbacks

mahmoodw:
This PR adds in the fallback cases when not enough or null ptr workspace is provided. 

Notes
- correctly initialises inputs workspacesize
- updates requiredWorkspaceSize helper for data parallel and persistent kernel cases 
- logic in develop changes since my original implementation - to discuss validity of newer logic.
- using .githooks/pre-commit has introduced some additional formatting independent of my PR


tensilelite/Tensile/Source/lib/source/ContractionSolution.cpp
```diff
@@ -1190,15 +1167,32 @@ namespace TensileLite
...
             {
-                skGrid             = getSKGrid(problem, hardware, tiles);
+                if(!inputs.ws)
+                {
+                    std::cerr << "\nstreamk warning: workspace not initialized, falling back to "
+                                 "data parallel mode"
+                              << std::endl;
+                    skGrid = tiles;
+                }
+                else if(inputs.workspaceSize < requiredWorkspaceSize(problem, hardware))
```
bethune-bryant: It feels a little strange that we're using `requiredWorkspaceSize` here to determine what to set `skgrid` to, but `requiredWorkspaceSize` uses the skgrid to compute the required workspace.

It feels a little like a circular dependency.

skgrid depends on workspacesize which also depends on skgrid.

But maybe this is okay, since this is just handling an edge case when not enough workspace is passed in.

bethune-bryant (dismissed):
Looks good to me.

We can think about if there's any better way to handle the potential circular logic separately.

library/src/amd_detail/rocblaslt/src/tensile_host.cpp
```diff
@@ -537,7 +537,7 @@ namespace
                 : "",
             problem.tensor(TensileLite::ContractionProblemGemm::TENSOR::E).strides().size()
                 ? std::to_string(
-                    problem.tensor(TensileLite::ContractionProblemGemm::TENSOR::E).strides()[1])
+                      problem.tensor(TensileLite::ContractionProblemGemm::TENSOR::E).strides()[1])
```
KKyang: Is this trailing whitespace?

bethune-bryant: @mahmoodw Was this leading whitespace added automatically by the githook formatter?

mahmoodw: @bethune-bryant that's right

tensilelite/Tensile/Source/lib/source/ContractionSolution.cpp
mahmoodw: I understand the ambiguity. To clarify, requiredWorkspaceSize considers the expected output from getSKGrid to determine the amount of workspace needed. In this scenario, instead of blindly setting skgrid value to the getSKGrid result we consider safer options when the workspace is lacking. While the naming causes a double take, there is no actual circular dependency.

tensilelite/Tensile/Source/lib/source/ContractionSolution.cpp
```diff
@@ -1190,15 +1167,32 @@ namespace TensileLite
...
+                {
+                    std::cerr << "\nstreamk warning: insufficient workspace (given: "
```
jichangjichang: It would make a lot of terminal output if workspace size is not enough. not a good idea.

jichangjichang:
User should be able to know required workspace size from getHerursitic api.
If the user given workspace size is not enough for real stream-K function, should we just return no solution? This is how hipblaslt API designs.

tensilelite/Tensile/Source/lib/source/ContractionSolution.cpp
KKyang: can we use log level to print the message?

tensilelite/Tensile/Source/lib/source/ContractionSolution.cpp
```diff
@@ -2953,7 +2953,8 @@ namespace TensileLite
             auto   tiles  = problem.getNumTiles(sizeMapping);
             size_t skGrid = getSKGrid(problem, hardware, tiles);
             // Get space required for partial tiles
-            size += partialTileSize(skGrid);
+            if(tiles != skGrid || tiles % skGrid != 0)
```
daineAMD: Can just be `if(tiles % skGrid != 0)` pretty sure

AlexBrownAMD:
> User should be able to know required workspace size from getHerursitic api. If the user given workspace size is not enough for real stream-K function, should we just return no solution? This is how hipblaslt API designs.

Thanks for the feedback Jimmy. The team discussed the change and have an idea on how handle the change:

Running stream-k kernels without workspace (or with reduced workspace) could be a valid use-case, but it shouldn't be the "default" since it will have reduced performance. For example, I'm thinking of an HPC type application that requires large matrices, leaving no room for workspace. In this case we could run stream-k kernel in data-parallel mode without workspace. But the default user should pass in workspace for better performance.

We propose making a change so that if not enough workspace is provided, we return no solution found like you suggested. This would keep the default behavior to optimize performance. Then we can add an environment variable that will allow users to force it to run stream-k kernels with reduced workspace or no workspace. This way, no log output is needed since the user has to specifically opt-in to running it this way, and should be aware that performance may be affected.

If this sounds good to you, we'll go ahead and update the PR.

tensilelite/Tensile/Source/lib/include/Tensile/GranularitySelectionLibrary.hpp
```diff
@@ -188,8 +178,8 @@ namespace TensileLite
                     std::cout << row.second->description() << ": ";
                 }
 
-                if(softwarePredicate(searchType, *(row.second), problem)
-                   && (*row.second->hardwarePredicate)(hardware))
+                if(softwarePredicate(searchType, *(row.second), problem)
```
mahmoodw: softwarePredicate with searchType is new from Tensile perspective, to avoid breaking I've added my change in addition to this. @jichangjichang I would appreciate if you could advise on this section about whether it is sufficient.

mahmoodw:
@AlexBrownAMD  @bethune-bryant  @jichangjichang As advised I have made change to maintain API expectations. Now we will return error: NO solution found! When there is  insufficient workspace for streamk kernels. I have also added the env variable TENSILE_STREAMK_DATA_PARALLEL, when set will override the memory requirements and force data parallel mode for streamk runs (no workspace requirement). This involved moving the workspaceCheck problem predicate to a separate task predicate as done in classic Tensile. It also consolidates the logic for workspace to a single existing ContractionSolution::requiredWorkspaceSize.

All reviews will be greatly appreciated. The functionality has been tested locally but would like to CI results and hear back from others to make sure there are no breaking changes.

tensilelite/Tensile/Source/lib/include/Tensile/ContractionTaskPredicates.hpp
```diff
@@ -0,0 +1,131 @@
...
+                                            "prob",
+                                            0,
+                                            "<=",
+                                            "max",
+                                            task.problem.workspaceSize());
+
+                    return debugEvalCmp(task,
+                                        stream,
+                                        "prob",
+                                        task.solution.requiredWorkspaceSize(task.problem, task.hardware),
```
AlexBrownAMD: Does requiredWorkspaceSize function detect streamKDP setting? It should return 0 workspace required for that case.

tensilelite/Tensile/Source/lib/include/Tensile/SingleSolutionLibrary.hpp
```diff
@@ -92,18 +92,7 @@ namespace TensileLite
...
             if(solution)
             {
-                if(debug)
-                {
-                    std::cout << "hardwarePredicate:" << std::endl;
-                    solution->hardwarePredicate->debugEval(hardware, std::cout);
-                    std::cout << std::endl;
-                    std::cout << "problemPredicate:" << std::endl;
-                    solution->problemPredicate->debugEval(problem, std::cout);
-                    std::cout << std::endl;
```
AlexBrownAMD: The debug output in canSolve looks a bit different than the old code. Did you check if the debug output is still readable?

tensilelite/Tensile/Source/lib/source/ContractionSolution.cpp
```diff
@@ -1274,15 +1249,20 @@ namespace TensileLite
...
             AMDGPU const* pAMDGPU = dynamic_cast<AMDGPU const*>(&hardware);
             assert(pAMDGPU != nullptr && pAMDGPU->computeUnitCount != 0);
             if(sizeMapping.streamK != 0)
             {
-                skGrid             = getSKGrid(problem, hardware, tiles);
+                const bool streamKDP = Debug::Instance().useStreamKDataParrallel();
+                if(streamKDP)
+                    skGrid = tiles;
+                    //TODO Use heuristic to decide fallback to reduced grid instead of DP
+                else
```
AlexBrownAMD: Might be safer to include this check inside getSKGrid, since that is called from a couple places

AlexBrownAMD (commented):
I don't think there's a CI test case for this right now, is there? Do you have a test case that you ran locally? What error message does the user see when workspace is missing?

tensilelite/Tensile/Source/lib/include/Tensile/ContractionTaskPredicates.hpp
mahmoodw: good point - I've moved the logic from the predicate to inside requiredWorkspaceSize for a cleaner implementation

tensilelite/Tensile/Source/lib/source/ContractionSolution.cpp
mahmoodw: thanks for the suggestion, implemented

tensilelite/Tensile/Source/lib/include/Tensile/SingleSolutionLibrary.hpp
mahmoodw: good catch, I've added it back

mahmoodw:
yes I tested a use case locally, with and without the DATA PARALLEL env variable. When there is insufficient workspace size and the env variable is not set we see the output below

return error: NO solution found!

AlexBrownAMD (dismissed):
Thanks for the updates, looks good to me!

jichangjichang:
@mahmoodw  May I know why did you add taskPredicate? Any reason that we can't handle this fallback design by using original problemPredicate?

mahmoodw:
> @mahmoodw May I know why did you add taskPredicate? Any reason that we can't handle this fallback design by using original problemPredicate?

The already implemented requiredWorkspaceSize is not accessible from problemPredicate. These changes match classic Tensile style and avoid size logic being stored in multiple places for maintainability and redundancy

AlexBrownAMD (dismissed):

jichangjichang (dismissed):
please rebase

AlexBrownAMD (approved):

jichangjichang (approved):

PR #1686, opened 2025-02-20
Cache swizzled tensor for tuning

Serge45:
- Cache swizzled tensor according to its datatype and size, to avoid repeating host side swizzles.
 - An LRUCache was introduced for balancing memory usage and performance.
 - ~~Re-layout will be skipped if validation disabled.~~

### DIFF
--- a/tensilelite/Tensile/Source/client/source/DataInitialization.cpp
+++ b/tensilelite/Tensile/Source/client/source/DataInitialization.cpp
@@ -94,6 +94,10 @@ namespace TensileLite
                 return track.first;
             }
 
+            const K &back() const {
+                return entries.back();
+            }
+
         private:
             EntryMap entryMap;
             Entries  entries;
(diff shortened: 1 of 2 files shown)


jichangjichang (dismissed):

tensilelite/Tensile/Source/client/source/DataInitialization.cpp
```diff
@@ -1885,38 +1921,55 @@ namespace TensileLite
                       || (problem.swizzleTensorB() && i == ContractionProblemGemm::TENSOR::B);
 
                 void* ptr{};
-                //FIXME: Not good, need to use format to specify the way for swizzling.
-                //TODO: Support more swizzling type, such as 32x32x8, currently we have 16x16x8 only.
-                if(needSwizzle)
+
+                //if no validation, skip the swizzle
```
boringmorning: I understand we don't need to permute the tensor when no validation. But is it correct to skip the padding?

Serge45: After doing some experiments on swizzled and non-swizzled on STA problem, I observed that there's performance disparity. In the latest commit, it uses LRUCache to manage the cached tensor for balancing memory usage and runtime performance, and now it always performs swizzle for STA/STB problem.

geotseng-amd (approved):

jichangjichang (approved):

PR #1677, opened 2025-02-16
Remove global variables

davidd-amd:
The goal of this PR is to eliminate global variable usage in library code. Namely, the `globalParameters` and references to `validMFMA` or `validSMFMA` via `validParameters`. The motivation for doing so is to 

1. eliminate global state copied into processes via joblib parallel maps
2. enable usage of tools such as [ray](https://www.bing.com/ck/a?!&&p=3618c8ab4a5fc386336067535164056664e5d1ac0e0f80229d95d3747aa3fa63JmltdHM9MTc0MjE2OTYwMA&ptn=3&ver=2&hsh=4&fclid=05f93799-7824-6b0c-0d8a-23cd79876a65&psq=python+ray&u=a1aHR0cHM6Ly9kb2NzLnJheS5pby9lbi9sYXRlc3QvaW5kZXguaHRtbA&ntb=1) which cannot cope with global variable usage as it exists in the code today
3. Make the library code more testable and easier to understand

To do so, global variables were removed from library functions and state is passed through function calls. Some refactoring occurred to avoid global variable usage patterns altogether. As an example, kernel validation code that relied on large lists was factored out such that parts of the code that did not directly depend on the lists do not incur a ram usage penalty. The validation functionality is moved into the TensileLogic package such that it can be applied separate from the build system.

TensileLogic:

To validate the current set of logic files with all checks, issue
```bash
Tensile/bin/TensileLogic --check-all ./path/to/logic/directory/
```
As of Mar. 17, 2025, this takes roughly 4 minutes. For additional options, review the usage string for TensileLogic with `Tensile/bin/TensileLogic --help`.

Custom Kernels:

Many custom kernel files may have a section called `custom.config` that defines YAML to override the solution that references it. Previously, within this section, the matrix instruction for the kernel could be marked `MatrixInstruction: [w, x, ..., z]`, where the length of this array could be 0 (no matrix instruction), 4, or 9. Since converting between the length-9 and length-4 format requires the objects `validMFMA`, `validWMMA`, among others, this pattern imposes large memory costs (on the order of 5-7 GB+). These memory costs can be avoided by simply relying on the matrix instructions defined on the solution referencing the custom kernel and enforcing correctness of the solution through CI. The only requirement for this is that all custom kernels used in TensileCreateLibrary must use the length-4 matrix instruction. Please note that for prototyping, length-9 matrix instructions are still supported when being called from `./Tensile`.

Contributors:

[...]

PR #1687, opened 2025-02-20
Improve valid parameter checks

bstefanuk:
**Summary**

Prior to enabling CI with valid parameter checks, there remains some set up work to fully adapt the checks to the existing files.

This PR:
- Includes CustomKernel logic configurations in the check.
- Breaks the following two operations into separate functions: (i) matrix instruction conversion to MI parameters, and (ii) validation that the MI parameters are correct.
- Removes validMatrixInstruction and validWorkGroups from GlobalParameters.py


bstefanuk:
Integrated into larger feature branch #1677

PR #1696, opened 2025-02-21
We need hipblas-common-static for static-devel builds

davidd-amd:
While hipblas-common is a header only library, they produce static-devel packages so we have to express a dependency on the static-devel package for our static configs.

The rhel8 testing environments do not provide all of the tooling required to run Tensile. To avoid failures or hangs, we attempt to check if we are in a rhel8 environment. If so, we disable the frequency check. This does require the distro package which is imported in a try/except block to guard against environments where it isn't installed.


samjwu (dismissed):
fixes the compile stage failure seen here
http://math-ci.amd.com/job/static-libraries/job/static-build/job/hipBLAS/job/develop/7/pipeline-console/?selected-node=230

jichangjichang (dismissed):

tensilelite/Tensile/Tensile.py
```diff
@@ -396,6 +399,16 @@ def Tensile(userArgs):
     device_id = config["GlobalParameters"].get("Device", globalParameters["Device"])
     UseEffLike = config["GlobalParameters"].get("UseEffLike", globalParameters["UseEffLike"])
 
+    def isRhel8():
+        try:
+            import distro
+            dist = distro.linux_distribution()
+            return True if 'Red Hat Enterprise Linux' in dist[0] and '8.' in dist[1] else False
+        except:
```
davidd-amd, line 405:
```suggestion
        except:
            printWarning("Unable to detect linux distribution. Install run pip install requirements.txt to resolve warning.")
```

davidd-amd:
@jichangjichang I added some warnings when running on rhel8. Tests retriggered.

jichangjichang (dismissed):

bstefanuk (dismissed):

bstefanuk (approved):

jichangjichang (approved):

PR #1705, opened 2025-02-26
Fix pickle copying singleton

KKyang:
Prevents any modules or functions which use pickle as backend copy a new instance.


StaceyLai (approved):

vin-huang (approved):

davidd-amd:
@KKyang I think I understand the goal of this change but can you please provide a description of what this solves and any measurements of build times and ram usage before and after? Otherwise, it looks good to me. Thanks for sharing.

PR #1713, opened 2025-02-27
Deprecate default as cov option

davidd-amd:
If users are configuring with cmake directly and explicitly set cov to "default" this will change default to 4 and warn that the default option is deprecated.


bstefanuk:
I'm concerned this won't cover all of the use-cases (though it may). Some individuals are setting code object version in input configuration files, so they may be looking for a way to translate those entries internal to Tensile/TensileCreateLibrary. We may have to hunt down these workflows for more clarity.

davidd-amd:
@KKyang what we would like to do is stop adding eagerly adding `CodeObjectVersion` to the logic files generated by tensile. By doing this the application will use the default. If someone wants to add this field to the yaml file and override the deafult, they can use something like:

```
Tensile/bin/Tensile --code-object-version=abc123 config.yaml tensile-out
```
If this option is present then we can add it to the logic files generated by tensile. This will preserve backwards compatibility by relying on a default when the option isn't present and also provide the flexibility to use an alternative value if one would like to use a different cov.

jichangjichang (approved):

PR #1723, opened 2025-03-04
Added missing support for bpl==64

amd-mgeorgie:
**Symptom:** Error message of this kind:
AssertionError: Cijk_Ailk_Bljk.....
chooseGlobalRead: bad bpl 64
**Reason:** There is no condition handler for the case of 
bpl==64 (bytes) in the fille **KernelWriterAssembly.py**
**Occurrence:** The use of such is not very common and has been encountered for grid generating routine for I8II data type in several "cat" scripts
**How to reproduce:**
sudo bash run_I8II_NN_cat34.sh
for attached files (extensions changed to "txt")
I8II_NN_cat34.yaml
run_I8II_NN_cat34.sh
Validation:
The successful validation for the proposed solution has been logged in "I8II_NN_cat34-tensileliteVALIDM1.log", attached to this pull request. Can be reproduced for the very same scenario with 
**NumElementsToValidate: -1**

[I8II_NN_cat34-tensileliteVALIDM1.log](https://github.com/user-attachments/files/19070428/I8II_NN_cat34-tensileliteVALIDM1.log)
[I8II_NN_cat34.txt](https://github.com/user-attachments/files/19070435/I8II_NN_cat34.txt)

[run_I8II_NN_cat34.txt](https://github.com/user-attachments/files/19070444/run_I8II_NN_cat34.txt)

### DIFF
--- a/tensilelite/Tensile/KernelWriterAssembly.py
+++ b/tensilelite/Tensile/KernelWriterAssembly.py
@@ -11052,42 +11052,31 @@ class KernelWriterAssembly(KernelWriter):
           rv.add(BufferLoadB128(dst=dst, vaddr=addr0, saddr=addr1, \
                                 soffset=soffset, mubuf=mubuf2, comment=comment))
           return rv
-        elif bpl==64:
+        elif bpl==64 and not lds:
           rv = Module("emulated _buffer_load_b512")
           # +0.25
-          dst = None if lds else vgpr(destVgpr, rpv//4)
+          dst = vgpr(destVgpr, rpv//4)
           rv.add(BufferLoadB128(dst=dst, vaddr=addr0, saddr=addr1, \
                                 soffset=soffset, mubuf=mubuf, comment=comment))
+          
           mubuf2 = MUBUFModifiers(offen=True, offset12=offset+bpl/4, glc=glc, slc=slc, nt=nt, lds=lds)
-          
-          if isinstance(destVgpr, str):
-            dst2 = destVgpr + "+" + str(int(rpv//4))
-          else:       
-            dst2 = int(destVgpr + int(rpv//4))
-          
-          dst = None if lds else vgpr(dst2, rpv//4)          
+          dst2 = destVgpr + "+" + str(int(rpv//4)) if isinstance(destVgpr, str) else int(destVgpr + int(rpv//4))
+
+          dst = vgpr(dst2, rpv//4)          
           rv.add(BufferLoadB128(dst=dst, vaddr=addr0, saddr=addr1, \
                                 soffset=soffset, mubuf=mubuf2, comment=comment))
           #+0.5          
-          mubuf3 = MUBUFModifiers(offen=True, offset12=offset+bpl / 2, glc=glc, slc=slc, nt=nt, lds=lds)          
+          mubuf3 = MUBUFModifiers(offen=True, offset12=offset+bpl / 2, glc=glc, slc=slc, nt=nt, lds=lds)
+          dst3 = destVgpr + "+" + str(int(rpv//2)) if isinstance(destVgpr, str) else int(destVgpr + int(rpv//2))
           
-          if isinstance(destVgpr, str):
-            dst3 = destVgpr + "+" + str(int(rpv//2))
-          else:
-            dst3 = int(destVgpr + int(rpv//2))            
-          
-          dst = None if lds else vgpr(dst3, rpv//4)          
+          dst = vgpr(dst3, rpv//4)          
           rv.add(BufferLoadB128(dst=dst, vaddr=addr0, saddr=addr1, \
                                 soffset=soffset, mubuf=mubuf3, comment=comment))
           #+0.75
-          mubuf4 = MUBUFModifiers(offen=True, offset12=offset + 3 * (bpl / 4), glc=glc, slc=slc, nt=nt, lds=lds)
+          mubuf4 = MUBUFModifiers(offen=True, offset12=offset + 3 * (bpl / 4), glc=glc, slc=slc, nt=nt, lds=lds)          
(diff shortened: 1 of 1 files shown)


smalekta:
@amd-mgeorgie - Can you please check if with a bad code and NumElementsToValidate: -1 the tensile is still breaking and with your change and NumElementsToValidate: -1 it passes?

smalekta:
Can you please add a test tensilelite/Tensile/Tests?

amd-mgeorgie:
Yes, I have succeed to validate the test. Here you are the logged output
[Uploading I8II_NN_cat34-tensileliteVALIDM1.log…]()

And this is the test before (with bad code)

[I8II_NN_cat34-tensilelite_before.log](https://github.com/user-attachments/files/19122204/I8II_NN_cat34-tensilelite_before.log)

tensilelite/Tensile/KernelWriterAssembly.py
```diff
@@ -11052,6 +11052,44 @@ def bufferLoadImpl(soffset):
...
+          #+0.75
+          mubuf4 = MUBUFModifiers(offen=True, offset12=offset + 3 * (bpl / 4), glc=glc, slc=slc, nt=nt, lds=lds)
+          
+          if isinstance(destVgpr, str):
+            dst4 = destVgpr + "+" + str(3*int(rpv//4))
+          else:
+            dst4 = int(destVgpr + 3*int(rpv//4))
+          
+          dst = None if lds else vgpr(dst4, rpv//4)
+          rv.add(BufferLoadB128(dst=dst, vaddr=addr0, saddr=addr1, \
```
KKyang, line 11080: In which case do we need 4 b128?

amd-mgeorgie: Good question,
It has been deduced from 

elif bpl==32: (above)
in which we need 2b 128
and 1b 128 for bpl < 32

amd-mgeorgie (commented):
I hope the code is fine

tensilelite/Tensile/KernelWriterAssembly.py
```diff
@@ -11052,6 +11052,44 @@ def bufferLoadImpl(soffset):
           rv.add(BufferLoadB128(dst=dst, vaddr=addr0, saddr=addr1, \
                                 soffset=soffset, mubuf=mubuf2, comment=comment))
           return rv
+        elif bpl==64:
+          rv = Module("emulated _buffer_load_b256")
+          # +0.25
+          dst = None if lds else vgpr(destVgpr, rpv//4)
```
b-shi: Could we throw an error for the `lds` case instead of setting `dst = None` - so we only handle `blp == 64` and `not lds`. Something like
```
elif bpl==64 and not lds:
  rv = Module("emulated _buffer_load...)
  dst = vgpr(...
```
My concern here is that for the lds case we also need to update `m0` when we have multiple loads.

tensilelite/Tensile/KernelWriterAssembly.py
```diff
@@ -11052,6 +11052,44 @@ def bufferLoadImpl(soffset):
           rv.add(BufferLoadB128(dst=dst, vaddr=addr0, saddr=addr1, \
                                 soffset=soffset, mubuf=mubuf2, comment=comment))
           return rv
+        elif bpl==64:
+          rv = Module("emulated _buffer_load_b256")
```
b-shi:
```suggestion
          rv = Module("emulated _buffer_load_b512")
```

tensilelite/Tensile/KernelWriterAssembly.py
amd-mgeorgie: I will work on it right now

amd-mgeorgie: Ah, I see your point...the code will be cleaner, yes!

amd-mgeorgie:
The code has been successfully modified and tested for the very same file commands as above...the compilation process is passing by, and the library is providing the expected boost.

tensilelite/Tensile/KernelWriterAssembly.py
amd-mgeorgie: The code has been corrected according to your suggestions, the compilation process works now, the test conditions as given above are also passing and library is providing expected boost!

b-shi (dismissed):
LGTM, thanks!

tensilelite/Tensile/KernelWriterAssembly.py
```diff
@@ -11052,6 +11052,33 @@ def bufferLoadImpl(soffset):
...
                                 soffset=soffset, mubuf=mubuf2, comment=comment))
           return rv
+        elif bpl==64 and not lds:
+          rv = Module("emulated _buffer_load_b512")
+          # +0.25
+          dst = vgpr(destVgpr, rpv//4)
+          rv.add(BufferLoadB128(dst=dst, vaddr=addr0, saddr=addr1, \
+                                soffset=soffset, mubuf=mubuf, comment=comment))
+          
+          mubuf2 = MUBUFModifiers(offen=True, offset12=offset+bpl/4, glc=glc, slc=slc, nt=nt, lds=lds)
```
KKyang: We are planning to add a mechanism to make sure like offset12 should be an int instead of float. Can you change it to int(offset+bpl/4)? And can you also change the other related settings as well?

KKyang:
I don't know why your PR is not triggering CI properly.

tensilelite/Tensile/KernelWriterAssembly.py
amd-mgeorgie: Sure!

amd-mgeorgie:
The code has been updated according to the recommendations:
now integer data format is enforced
Along with a small polish to redundant brackets

amd-mgeorgie (commented):
all have been addressed

b-shi (approved):

KKyang (approved):

amd-mgeorgie (commented):
ready to merge

PR #1737, opened 2025-03-06
Reduce the number of duplicate kernel helpers generated

davidd-amd:
The current procedure for creating kernel helper objects is inefficient. In part, because they are created every time a solution is created. However, one kernel helper may map to many solutions so mapping kernel helper creation to the creation of a solution results in an operation where many helpers are created and then deduplicated. The problem with this approach is that it requires significant memory overhead and to a lesser degree runtime overhead.

Here we introduce an improved approach where solutions are created first and then used to create the kernel helper objects. To do so, a set of unique kernel helper object names is used to determine whether or not a kernel helper object should be initialized. Previously, a kernel helper name could not be computed without creating a kernel helper and modifying state. This PR adds new naming functions that do not require a kernel helper is created and do not modify state. 

| target | branch | build time (s)| RAM (GB) |
|--------|---------|---------------|------------|
| gfx90a | feature | 242 | 12 |
| gfx90a | develop | 342 | 17 |
| gfx942 | feature |  3151 |  46 |
| gfx942 | develop | 4873| 157  |


tensilelite/Tensile/KernelHelperNaming.py
```diff
@@ -0,0 +1,262 @@
...
+    ActivationFunction = 3
+    ActivationOnly = 4
+    Reduction = 5
+
+
+def conversionKernelObjectsNames(solution):
+  # need to check that the fields mutated in the the init function aren't used in the naming function
+  # e.g. usebias got set to zero if gradient which changes behavior of name function.
+  conversionKernelObjectsNames = []
+  load_vector_width = [1, 2] if solution["ProblemType"]["DataType"].isDouble() else [1, 2, 4]
```
bstefanuk: Use camelCase

tensilelite/Tensile/KernelHelperNaming.py
```diff
@@ -0,0 +1,262 @@
...
+  gsuList = [internalParameters["GlobalSplitUPGR"]]
+  if solution["GlobalSplitUAlgorithm"] == "SingleBuffer":
+    gsuList = [1]
+  elif solution["GlobalSplitUAlgorithm"] == "MultipleBufferSingleKernel":
+    return
```
bstefanuk, line 56: Empty early return.
Is this a bug? If not, how is it used?

tensilelite/Tensile/KernelHelperNaming.py
```diff
@@ -0,0 +1,262 @@
...
+    else:
+      conversionKernelObjectsNames.append(KernelWriterConversion._getKernelName(solution, vw))
+  return conversionKernelObjectsNames if conversionKernelObjectsNames else []
+
+
+def activationEnumHeaderObjectsNames(solution):
+  activationEnumHeaderObjectsNames = []
+  if solution["ProblemType"]["ActivationType"] in ['all', 'hipblaslt_all']:
+    activationEnumHeaderObjectsNames.append(KernelWriterActivationEnumHeader._getKernelName(solution))
+  return activationEnumHeaderObjectsNames
```
bstefanuk: This could be a one-liner
```suggestion
def activationEnumHeaderObjectsNames(solution):
  return [] if solution["ProblemType"]["ActivationType"] not in ['all', 'hipblaslt_all'] else [KernelWriterActivationEnumHeader._getKernelName(solution)]
```
edit: I see similar, but slightly more complex functions below that may not be oneline-able. If they can't all be made oneliners, then I support consistency, as it is coded now.

tensilelite/Tensile/KernelHelperNaming.py
```diff
@@ -0,0 +1,262 @@
...
+
+def reductionKernelObjectsNames(solution):
+  reductionKernelObjectsNames = []
+  if solution["ProblemType"]["Gradient"] and solution["ProblemType"]["UseBias"]:
+    for btype in solution["ProblemType"]["BiasDataTypeList"]:
+      reductionKernelObjectsNames.append(KernelWriterReduction._getKernelName(solution, btype))
+  return reductionKernelObjectsNames
+
+
+def betaOnlyKernelObjectsNames(solution):
```
bstefanuk: Could we drop "Objects" from these function names? It's verbose and repetitive when, afaict, these functions are generating "Kernel Names" in the conventional sense.

tensilelite/Tensile/KernelWriterConversion.py
```diff
@@ -785,6 +782,64 @@ def kernelBody(self):
...
+    indexChars = INDEX_CHARS
+    # C dimensions
+    name = "C"
+    for i in range(0, solution._state["ProblemType"]["NumIndicesC"]):
+      name += indexChars[i].lower()
+    name += "_"
+
+    # add input datatype into kernel name (the datatype of workspace)
+    inputTypeStr = DataType("I").toChar() if solution._state["ProblemType"]["DataType"].isInt8() or solution._state["ProblemType"]["DataType"].isInt32() else \
+                                  (DataType("D").toChar() if solution._state["ProblemType"]["DataType"].isDouble() else DataType("S").toChar())
```
bstefanuk: Couldn't we simply call `.toChar()` here? Constructing a `DataType` to call it's `.toChar()` method when we already have the DataType variant, seems odd.

tensilelite/Tensile/SolutionStructs/Solution.py
```diff
@@ -235,7 +235,7 @@ def __init__(
     )
     self._name = config["CustomKernelName"] if "CustomKernelName" in config and config["CustomKernelName"] else None
 
-    self.initHelperKernelObjects(targetIsas)
+    #self.initHelperKernelObjects(targetIsas)
```
bstefanuk: Delete

tensilelite/Tensile/KernelWriterActivationFunction.py
```diff
@@ -21,6 +21,7 @@
 ################################################################################
 
 from rocisa import rocIsa
+#from rocisa import fastdeepcopy as deepcopy
```
bstefanuk: Delete

tensilelite/Tensile/KernelWriterActivationEnumHeader.py
```diff
@@ -44,6 +44,12 @@ def __init__(self, state):
   def keys(self):
     return self.getKernelName()
 
+  @staticmethod
+  def _getKernelName(solution):
+    s = "Gradient" if solution._state["ProblemType"]["Gradient"] else ""
+    return "Tensile%sActivationEnum_%s"%(s,
+                                         solution._state["ProblemType"]["ActivationComputeDataType"].toChar())
```
bstefanuk: formatting
```suggestion
    return "Tensile%sActivationEnum_%s"%(s, solution._state["ProblemType"]["ActivationComputeDataType"].toChar())
```

tensilelite/Tensile/KernelHelperNaming.py
```diff
@@ -0,0 +1,262 @@
...
+    if kernelHelperType == KernelHelperEnum.ActivationEnumHeader:
+        return initActivationEnumHeaderObjects(solution)
+    if kernelHelperType == KernelHelperEnum.ActivationFunction:
+        return initActivationFunctionObjects(solution, cxxCompiler, isaInfoMap)
+    if kernelHelperType == KernelHelperEnum.ActivationOnly:
+        return initActivationOnlyKernelObjects(solution)
+    if kernelHelperType == KernelHelperEnum.Reduction:
+        return initReductionKernelObjects(solution)
+    else:
+        raise Exception("Failed to find kenerl helper type.")
```
bstefanuk:
```suggestion
        raise Exception("Failed to find kernel helper type.")
```

tensilelite/Tensile/KernelHelperNaming.py
```diff
@@ -0,0 +1,262 @@
...
+
+
+def initBetaOnlyKernelObjects(solution):
```
bstefanuk, line 140: What is the motivation to define these here rather than in their respective KernelWriter files?

tensilelite/Tensile/KernelWriterActivationEnumHeader.py
```diff
@@ -44,6 +44,12 @@ def __init__(self, state):
...
+
   def getKernelName(self):
     return "Tensile%sActivationEnum_%s"%(self.actGradientPrefix, \
                                          self.state["ProblemType"]["ActivationComputeDataType"].toChar())
```
bstefanuk: So there's a static private method `_getKernelName()` and a also a public method `getKernelName()`. Are they guaranteed to return the same results?

tensilelite/Tensile/KernelHelperNaming.py
```diff
@@ -0,0 +1,262 @@
...
+  if (solution["ActivationFused"] == False) and (solution["ProblemType"]["ActivationType"] != 'none'):
+    activationOnlyKernelObjectsNames.append(KernelWriterActivationOnly._getKernelName(solution))
+  return activationOnlyKernelObjectsNames
+
+
+def reductionKernelObjectsNames(solution):
+  reductionKernelObjectsNames = []
+  if solution["ProblemType"]["Gradient"] and solution["ProblemType"]["UseBias"]:
+    for btype in solution["ProblemType"]["BiasDataTypeList"]:
+      reductionKernelObjectsNames.append(KernelWriterReduction._getKernelName(solution, btype))
```
bstefanuk: I would say it's non-idiomatic to call privately marked methods like this from imported types. If this function were in the file where KernelWriterReduction is defined, this would be more appropriate. Or if all of the files were collected into a module so the underscore usage is "hidden".

Is the old `.getKernelName` function still needed on these types or can it be repurposed?

bstefanuk (commented):
Please add a description with build perf results.

Looking forward to this update!

tensilelite/Tensile/KernelHelperNaming.py
davidd-amd: Its a naming convention that was there before. I'm not particular.

tensilelite/Tensile/KernelHelperNaming.py
davidd-amd: These are free functions. They operate on solutions. I am not strongly opposed to it but I don't know that it adds considerable value.

tensilelite/Tensile/KernelWriterActivationEnumHeader.py
davidd-amd: If used consistently they are. Unfortunately I couldn't have a static method that has the same name. Using an `_` probably isn't the best convention. I will have to think about naming.

tensilelite/Tensile/KernelWriterActivationFunction.py
davidd-amd: I need to test using this copy vs the built in

tensilelite/Tensile/KernelWriterConversion.py
davidd-amd: Already have the DataType variant?

davidd-amd: These functions are nearly identical to the other name functions but they don't mutate and they do not refer to class state.

tensilelite/Tensile/KernelHelperNaming.py
davidd-amd: missing the list

KKyang (approved):

bstefanuk (approved):

PR #1741, opened 2025-03-07
solution proposed with small polish in client.cpp

amd-mgeorgie:
**Details:** There is a nested confusion for the default bias type (bias_type)  when I8II data format scheme is used

**Symptom:**

hipblaslt-bench client confuses the default bias type (bias_type) for I8II data format scheme
for 32-bit float (f32_r) instead of (i32_r) (c.f screenshot below, highlighted text)
image

**HOW TO REPRODUCE**

Run hipblaslt-bench for I8II kernel scheme
e.g. 
hipblaslt-bench --yaml ./bias_tests.yaml 
where 
bias_tests.yaml is a file with an exemplary test line as follows

- {function: matmul, transA: N, transB: N, a_type: i8_r, b_type: i8_r, c_type: i32_r, d_type: i32_r, scale_type: i32_r, compute_type: c_i32_r, bias_type: i32_r, M: 7, N: 16, K: 2048, ldc: 7, ldd: 7, lda: 7, ldb: 2048, initialization: rand_int, alpha: 1, beta: 1, iters: 406009, cold_iters: 406009}

**SOLUTION**
a solution is given for mixed default type in testing_matmul.hpp
a mild polishing is added in the helper, where additional optional data type added for i32_r
**VALIDATION**

image

### DIFF
--- a/clients/include/testing_matmul.hpp
+++ b/clients/include/testing_matmul.hpp
@@ -1056,8 +1056,10 @@ hipDataType derive_unset_bias_type(const Arguments& arg)
         {
             real_bias_type = arg.d_type;
         }
-    }
-
+    }     
+    else if(arg.compute_type == HIPBLAS_COMPUTE_32I && arg.bias_type == HIP_R_32F) //if bias_type equals to default (HIP_R_32_F)    
+         real_bias_type = HIP_R_32I; // default should be arg.d_type
+    
     if(supported_bias_types.count(real_bias_type) == 0)
         throw std::invalid_argument("Invalid bias type "
                                     + std::string(hip_datatype_to_string(real_bias_type)));
--- a/clients/benchmarks/client.cpp
+++ b/clients/benchmarks/client.cpp
@@ -419,15 +419,15 @@ try
 
         ("c_type",
          value<std::string>(&c_type), "Precision of matrix C. "
-         "Options: f32_r,f16_r,bf16_r,i8_r")
+         "Options: f32_r, f16_r, bf16_r, i8_r, i32_r")
 
         ("d_type",
          value<std::string>(&d_type), "Precision of matrix D. "
-        "Options: f32_r,f16_r,bf16_r,i8_r")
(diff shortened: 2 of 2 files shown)


KKyang:
The default bias for 32I should be 32F

clients/include/testing_matmul.hpp
```diff
@@ -1056,8 +1056,10 @@ hipDataType derive_unset_bias_type(const Arguments& arg)
         {
             real_bias_type = arg.d_type;
         }
-    }
-
+    }     
+    else if(arg.compute_type == HIPBLAS_COMPUTE_32I && arg.bias_type == HIP_R_32F) //if bias_type equals to default (HIP_R_32_F)    
+         real_bias_type = HIP_R_32I; // default should be arg.d_type
```
KKyang, line 1061: This is not correct for i32.

amd-mgeorgie: Ah, this answers so many questions :)
is this always the case for the compute_type == i32 ? (even if A, B are i8 ?)
If so, please reject the pull request :)

amd-mgeorgie: Just a short question? Is this in client.cpp misleading then for the bias_type?

"default(same with D type)"

Because d_type in the case of I8II is i32

amd-mgeorgie:
But in the KernelAssemblyWriter.py (c.f. screenshot below) there is an explicit condition that throws error if the bias_type for computing type of i32 is different from i32...which strongly implies that there is code confusion and
BTW, the bench works for both bias_types of  i32_r and f32_r....
Complicated :)

image

clients/include/testing_matmul.hpp
jichangjichang: yes, the default type is D but not for all datatype. ex: the default bias type of F8 gemm and I gemm are not D type.

jayhawk-commits:

PR #1753, opened 2025-03-11
Allow hipblaslt client build without Tensile

bstefanuk:
This PR adds the ability to build `hipblaslt-bench` and other clients, without having to rebuild tensilelite by providing a cmake customization point `-DTensile_SKIP_BUILD`. This customization point is toggled through the installation script via any of the flags `-n`, `--clients-only`, or `--no-tensile` (undocumented), which, for compatibility, has been repurposed from the previously broken implementation of `BUILD_WITH_TENSILE`.

**Results**
Given a prebuilt library, nominally under **./build/release/Tensile/library/**, build times for gfx90a and gfx942 are reduced to ~3 minutes. The following table demonstrates these results.

| Command                                             | Time        |
| --------------------------------------------------- | ----------- |
| `time ./install.sh -c -a gfx90a`                    | 10m22.137s   |
| `time ./install.sh -c -a gfx90a --prebuilt ./build` | 2m57.202s   |
| `time ./install.sh -c -a gfx942`                    | 105m18.830s |
| `time ./install.sh -c -a gfx942 --prebuilt ./build` | 2m56.004s   |

Documentation has been added to clients.rst to describe the new build option.


bstefanuk:
Draft will be removed after relevant documentation is added to the docs/clients.rst file.

docs/clients.rst
```diff
@@ -132,3 +132,20 @@ For more information, see command:
    --flush                            Flush icache, only works for gemm.
    --help |-h                         produces this help message
    --version <value>                  Prints the version number
+
+
+Building clients with prebuilt libraries
+========================================
+
+Sometimes it is desirable to build or re-build the clients without having to conduct a full build of the library. This can be done by adding the `--prebuilt <path>` option to the install script. For example, `./install.sh -c -a gfx942 --prebuilt /path/to/previous_build/release/Tensile/library` will build the clients using the prebuilt library at the specified path. Internally, this conducts two actions. First it sets the environment variable `HIPBLASLT_TENSILE_LIBPATH` to the provided path. Second, it passes the `-DTensile_PREBUILT_LIBRARY=ON` option to CMake. As such, if building hipBLASLt with CMake directly instead of through the install script is preferred, the same effect can be achieved with the following steps:
```
amd-jnovotny:
```suggestion
Sometimes it is desirable to build or rebuild the clients without having to conduct a full build of the library. This can be done by adding the ``--prebuilt <path>`` option to the install script. For example, ``./install.sh -c -a gfx942 --prebuilt /path/to/previous_build/release/Tensile/library`` will build the clients using the prebuilt library at the specified path. Internally, this conducts two actions. First, it sets the environment variable ``HIPBLASLT_TENSILE_LIBPATH`` to the provided path. Second, it passes the ``-DTensile_PREBUILT_LIBRARY=ON`` option to CMake. If you prefer to build hipBLASLt with CMake directly instead of through the install script, the same effect can be achieved with the following steps:
```

amd-jnovotny (dismissed):
Looks good overall, but code needs to be enclosed in double ticks.

docs/clients.rst
bstefanuk: Thank you @amd-jnovotny, the documentation has been updated according to your recommendations. Please mark as resolved if you are happy with the changes.

amd-jnovotny (dismissed):

install.sh
```diff
@@ -718,6 +723,16 @@ pushd .
     cmake_common_options="${cmake_common_options} -DBUILD_SHARED_LIBS=OFF"
   fi
 
+  # build with prebuilt Tensile libraries
+  if [[ -n "${build_prebuilt}" ]]; then
+    cmake_common_options="${cmake_common_options} -DTensile_PREBUILT_LIBRARY=ON"
+    # Check if HIPBLASLT_TENSILE_LIBPATH is set, if not, set it
+    if [[ -z "${HIPBLASLT_TENSILE_LIBPATH}" ]]; then
+      export HIPBLASLT_TENSILE_LIBPATH=${build_prebuilt}
+    fi
```
davidd-amd: Is there precedence for modifying users environment in the install script? I would rather not set a users environments like this. `HIPBLASLT_TENSILE_LIBPATH` is a runtime parameter so it can be set after building. I would like to warn that we are setting environment because it could cause confusion if they have subsequent builds and expect to use co files from the build dir.

bstefanuk: As a runtime parameter, you're right, it doesn't make much sense to be setting HIPBLASLT_TENSILE_LIBPATH here. I'll remove it. I notice as well that the user will be required to call hipblaslt-bench with the environment variable set. Otherwise, it doesn't find the required libraries, or worse, it finds the wrong libraries installed at /opt/rocm/lib.

Unless we update the tensile_host internals, the current invocation pattern is, for example,
```bash
HIPBLASLT_TENSILE_LIBPATH=/path/to/_saved_build/release/Tensile/library/ ./hipblaslt-bench --precision f32_r -v
```
If it's called without HIPBLASLT_TENSILE_LIBPATH, then the following output is provided.
```
rocblaslt info: HIPBLASLT_TENSILE_LIBPATH not set. Using /opt/rocm/lib

rocblaslt error: Cannot read /mnt/host/hipBLASLt/wk/bench-faster/build/release/library/../Tensile/library/TensileLibrary_lazy_gfx90a.dat: No such file or directory
```

davidd-amd: I think that is reasonable. If we do find them in /opt/rocm then it will run but we have provided a diagnostic. Hopefully it is not buried in other output.

bstefanuk:
Updated to develop. @stellaraccident I noticed you requested to review code changes related to CMake so I've added you as a reviewer. Your input is appreciated as well @bnemanich.

stellaraccident (approved):
Seems ok to me. I don't have an opinion on the install.sh as we never use that.

KKyang (approved):

KKyang:
The install.sh still removes the build folder, so we'll use cmake directly in build/relase, right?

jichangjichang (approved):

bstefanuk:
@KKyang Good point, let me clarify. If you need to rebuild the clients only, you'll have a couple options.

1. With install.sh, point to a new build directory with `--build_dir`.
2. As you correctly pointed out, yes, you could use cmake directly within build/release.
3. Move the build directory to a different name and re-run the install.sh command.

PR #1754, opened 2025-03-12
Perf: added new problem yaml and api overhead tracking

solaslin:
Resolve SWDEV-521447

- [x] Added api_overhead tracking
- [x] Added new matmul problem set yaml
- [x] Disable flush

#### Run method 1:
(After ./install.sh with client:)
`[repoFolder]/build/release/clients/staging/hipblaslt-api-overhead --cold_iters 10 --iters 1000`
to show on the screen like:

```
function,api_name,us/iter,best_us
------------------------------------
api_overhead,hipblasLtMatmulAlgoGetHeuristic,114.748415,104.525000
api_overhead,hipblasLtMatmul,9.120849,7.741000
api_overhead,hipblaslt_ext::algoGetHeuristic,109.551578,99.982000
api_overhead,hipblaslt_ext::getAllAlgos,114.286553,99.986000
api_overhead,hipblaslt_ext::getAlgosFromIndex,1.071942,0.680000
api_overhead,hipblaslt_ext::run,8.592682,4.713000

```

#### Run method 2:
(After ./install.sh with client:)
`[repoFolder]/clients/scripts/performance/hipblaslt-perf --suite api_overhead `
to print on screen and save to a csv file looks like:
image


clients/benchmarks/client_api_overhead.cpp
```diff
@@ -0,0 +1,872 @@
+/*******************************************************************************
+ *
+ * MIT License
+ *
+ * Copyright (C) 2022-2025 Advanced Micro Devices, Inc.
```
KKyang: 2025 only

clients/benchmarks/client_api_overhead.cpp
```diff
@@ -0,0 +1,872 @@
...
+                                                              matC,
+                                                              matD,
+                                                              pref,
+                                                              request_solutions,
+                                                              heuristicResult,
+                                                              &returnedAlgoCount));
+
+        if(returnedAlgoCount == 0)
+        {
+            std::cerr << "No valid solution found!" << std::endl;
```
jichangjichang, line 336: free allocated memory before return

clients/benchmarks/client_api_overhead.cpp
```diff
@@ -0,0 +1,872 @@
...
+
+#include <chrono>
+#include <hip/hip_runtime.h>
+#include <hipblaslt/hipblaslt-ext.hpp>
+#include <hipblaslt/hipblaslt.h>
+#include <iostream>
+
+#include "helper.h"
+
+using clock_point  = std::chrono::time_point<std::chrono::steady_clock>;
```
KKyang, line 35: Should we have both GPU time and CPU time?

solaslin: But this "api_overhead" is focusing on the cpu time spent on purely calling this API [from call->return]
GPU time could be related a lot such as size (mempcy), so we are just measuring the cpu time for each duration of the "API-Start-End"
@jichangjichang  how do you think ?

clients/benchmarks/client_api_overhead.cpp
solaslin: done

clients/benchmarks/client_api_overhead.cpp
solaslin: done

hcman2 (approved):

KKyang (approved):

PR #1770, opened 2025-03-13
Deprecate AMDGPU_TARGETS variable

lawruble13:
Summary of proposed changes:

- Change the CMake to set the GPU_TARGETS variable from AMDGPU_TARGETS if not set by the user
- Change usages of AMDGPU_TARGETS to GPU_TARGETS

Note that #1563 was opened against 6.4, which I did not merge in time for 6.4, hence why this PR is being opened.


jayhawk-commits:
Not migrating this PR to the monorepo due to the target branch.

PR #1773, opened 2025-03-14
Adding hipblaslt streamk testing

mahmoodw:
- This change enables streamk within the hipblaslt test framework by adding argument parameter solution_selection. This enables the associated env variable to enable streamk kernel inclusion, while overriding and restoring existing values over the lifespan of the client handle. 
- Kernels configs added to support activation functions 
- Streamk tests with and without activation function use are added

Motivation for change is to ensure testing of Aux buffer for streamk workloads


clients/common/utility.cpp
```diff
@@ -233,13 +233,25 @@ hipblaslt_local_handle::hipblaslt_local_handle()
 hipblaslt_local_handle::hipblaslt_local_handle(const Arguments& arg)
     : hipblaslt_local_handle()
 {
-
+    if(arg.solution_selection >= 0)
+    {
+        hipblaslt_cout<<"arg.solution_selection "<<arg.solution_selection<<std::endl;
```
daineAMD: Can this be moved to the hipblaslt-bench output, and be removed from hipblaslt-test output?

daineAMD (commented):
One comment but looks good to me otherwise.

clients/common/utility.cpp
mahmoodw: oops I think it was some personal debug comments that slipped

jichangjichang:
> * This change enables streamk within the hipblaslt test framework by adding argument parameter solution_selection. This enables the associated env variable to enable streamk kernel inclusion, while overriding and restoring existing values over the lifespan of the client handle.
> * Kernels configs added to support activation functions
> * Streamk tests with and without activation function use are added
> 
> Motivation for change is to ensure testing of Aux buffer for streamk workloads

Could you change the naming of new parameter as tensile_solution_selection_method, which can directly map to naming of its env?

KKyang (approved):

AlexBrownAMD (approved):

PR #1776, opened 2025-03-14
Enable MX Datatypes in hipBLASLt

bnemanich:
This PR enables MX datatypes through rocRoller.

Initial support allows for A and B to be any combination of MXFP8, MXBF8, MXFP6, MXBF6, MXFP4. C and D can only be FP32, FP16, BF16.
M and N must be multiples of 16 and K must be a multiple of 128.
No extra epilogues are supported.

Kernels are generated at runtime. They are saved in memory for the duration of the program, but never saved to disk.


jichangjichang:
Did you run .githook/pre-commit for all of change files?

bnemanich:
> Did you run .githook/pre-commit for all of change files?

Ran the script, which fixed a lot of formatting issues. There were some areas that the script touched that weren't added by this PR, so probably some other commits snuck in without running the script as well.

bnemanich:
Ran tests for gfx950. The only tests that failed are the existing failures. Added many new MX tests as can be seen in output.
[gfx950_tests.txt](https://github.com/user-attachments/files/19458212/gfx950_tests.txt)

jichangjichang (dismissed):

TonyYHsieh (dismissed):
LGTM

CMakeLists.txt
```diff
@@ -117,6 +117,8 @@ option(Tensile_NO_LAZY_LIBRARY_LOADING "Diasble loading kernels on demand?" OFF)
 # For roctx
 include(CMakeDependentOption)
 cmake_dependent_option(HIPBLASLT_ENABLE_MARKER "Enable roctx marker in hipBLASLt" ON "BUILD_SHARED_LIBS" OFF)
+# For rocRoller
+option(USE_ROCROLLER "Experimental: Build with the rocRoller library" OFF)
```
jichangjichang: CQE doesn't use install.sh, which means rocroller won't be built in 6.5 staging build even this PR get merged. Is this same as your expectation?

bnemanich: Thanks @jichangjichang , I updated the PR because of this. Will need another approval.

jichangjichang (dismissed):

install.sh
```diff
@@ -612,14 +630,19 @@ if [[ "${build_hip_clang}" == true ]]; then
   export PATH=${rocm_path}/bin:${rocm_path}/hip/bin:${rocm_path}/llvm/bin:${PATH}
 fi
 
+# Use RocRoller
+if [[ ! ( "${ID}" == "rhel" && "${VERSION_ID}" == "9.1" ) && ( "${skip_rocroller}" == false ) ]]; then
```
jichangjichang, line 634: You try to skip rocroller for RHEL 9 distro. However, CQE side would always build on all distro since they didn't use install.sh way. Is that ok? Or should you check the distro version in CMakeLists.txt directly?

bnemanich: CQE builds should have been moved to use RHEL 9.5, which should work. This is kind of a temporary fix until Math CI also moves to RHEL 9.5.

jichangjichang: Ok.

jichangjichang:
@bnemanich Please check the build failures of Azure CIs.

awhittle3 (dismissed):
All build issues identified should now hopefully be accounted for

jichangjichang:
Please check build error:
var/jenkins_home/workspace/ain_precheckin_hipBLASLt_PR-1776/iVw1dXHtX/hipblaslt/build/release/_deps/rocroller-src/lib/include/rocRoller/Graph/Hypergraph.hpp:30:10: fatal error: 'boost/multi_index/key.hpp' file not found
   30 | #include <boost/multi_index/key.hpp>
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~
1 error generated.

awhittle3 (approved):

bethune-bryant (approved):
I've looked through alot of this and have kind of been following it as it's been worked on.
I don't see any stand out issues.

PR #1797, opened 2025-03-20
[Fix] hipblaslt-bench update return to exit failure to capturee signal on shell

RuibinCheung:
Current hipblaslt-bench will exit 0 if solution is not found. We hope to capture right signal of hipblaslt-bench process.


jichangjichang (commented):
we may have many test items in single execution. No solution doesn't mean actual error. It shoudn't call exit() to skip the rest test items.

RuibinCheung:
> we may have many test items in single execution. No solution doesn't mean actual error. It shoudn't call exit() to skip the rest test items.

Acknowledge. I will close this PR. Thanks a lot !

PR #1808, opened 2025-03-25
Add support to use TensileCreateLibrary independently

KKyang:
**Edit**
It won't rebuild unless you change the C++ code.
image

First time build

image

### DIFF
--- a/tensilelite/MANIFEST.in
+++ b/tensilelite/MANIFEST.in
@@ -1,5 +1,5 @@
 recursive-include Tensile *.py *.h *.hpp *.cpp *.txt *.cmake *.yaml
 recursive-include Tensile/CustomKernels *.s
-include Tensile/bin/Tensile Tensile/bin/TensileCreateLibrary
+include Tensile/bin/TensilePy Tensile/bin/TensileCreateLibraryPy
 include Tensile/Ops/gen_assembly.sh
 include requirements.txt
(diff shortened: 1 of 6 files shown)


hcman2 (approved):

tensilelite/Tensile/bin/TensileCreateLibrary
```diff
@@ -24,21 +23,30 @@
...
+    ROCM_PATH="/opt/rocm"
+fi
+WORKING_DIR=/tmp
+ROCISA_LIB_DIR=$(relative2absolute $WORKING_DIR)/rocisa/lib
+SCRIPT_DIR=$(dirname "$(readlink -f "$0")")
+ROCISA_DIR=$SCRIPT_DIR/../../rocisa
+### Build rocisa
+# Currently the path to rocisa is fixed
+mkdir -p $WORKING_DIR/rocisa
+cmake -DCMAKE_CXX_COMPILER=$ROCM_PATH/bin/amdclang++ -DCMAKE_PREFIX_PATH=$(python3 -m nanobind --cmake_dir) -DCMAKE_BUILD_TYPE=Debug -DPYTHON_VERSION=$(python3 --version 2>&1 | cut -d ' ' -f 2) -S $ROCISA_DIR -B $WORKING_DIR/rocisa
```
bnemanich, line 47: Do we really want to have cmake be a dependency here and attempt to do a build every time TensileCreateLibrary is called? Shouldn't this be pre-built already and just installed as a python module?

KKyang: No, it won't build every time unless you change the code, and this is a development environment, this is more like a dummy guard in case someone forgets or doesn't mention that this is a separate module.

aazz44ss (approved):

PR #1827, opened 2025-03-26
Support LDS>64K when PGR=0 but LdsNumElementsAlignedA/B exceed 64K

adityalj:
For MI [16, 16, 32, 1, 1, 1, 10, 1, 4] we can see that even if PGR=0 as numLDSbyetsB exceeds 64K we need two base address registers for B in LDS. We have made similar correction for A as well.

### DIFF
--- a/tensilelite/Tensile/KernelWriter.py
+++ b/tensilelite/Tensile/KernelWriter.py
@@ -3946,19 +3946,19 @@ class KernelWriter(metaclass=abc.ABCMeta):
     self.states.a.numVgprLocalWriteAddr = 0 if kernel["LocalWriteUseSgprA"] else 1 * self.states.rpla
     self.states.b.numVgprLocalWriteAddr = 0 if kernel["LocalWriteUseSgprB"] else 1 * self.states.rpla
 
-    if self.states.archCaps["HasLDSGT64K"] and not kernel["1LDSBuffer"] and not kernel["LocalWriteUseSgprA"] :
-      if kernel["LdsOffsetA_Blk"]>=131072:
+    if self.states.archCaps["HasLDSGT64K"] and not kernel["LocalWriteUseSgprA"] :
+      if (kernel["LdsOffsetA_Blk"]>=131072 and kernel["ExpandPointerSwap"]) or kernel["LdsNumElementsAlignedA"]>=131072:      
         self.states.a.numVgprLocalReadAddr =3* self.states.rpla
         self.states.a.numVgprLocalWriteAddr = 3* self.states.rpla
-      elif kernel["LdsOffsetA_Blk"]>=65536:
+      elif (kernel["LdsOffsetA_Blk"]>=65536 and kernel["ExpandPointerSwap"])or kernel["LdsNumElementsAlignedA"]>=65536: 
         self.states.a.numVgprLocalReadAddr =2* self.states.rpla
         self.states.a.numVgprLocalWriteAddr = 2* self.states.rpla
 
-    if self.states.archCaps["HasLDSGT64K"] and not kernel["1LDSBuffer"] and not kernel["LocalWriteUseSgprB"] :
-      if kernel["LdsOffsetA_Blk"]>=131072:
+    if self.states.archCaps["HasLDSGT64K"] and not kernel["LocalWriteUseSgprB"] :
+      if (kernel["LdsOffsetA_Blk"]>=131072 and kernel["ExpandPointerSwap"]) or kernel["LdsNumElementsAlignedB"]>=131072:
         self.states.b.numVgprLocalReadAddr =3* self.states.rpla
         self.states.b.numVgprLocalWriteAddr = 3* self.states.rpla
-      elif kernel["LdsOffsetA_Blk"]>=65536:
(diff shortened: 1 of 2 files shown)


alex391a (dismissed):
LGTM

tensilelite/Tensile/KernelWriter.py
```diff
@@ -3943,18 +3943,18 @@ def readWriteVectors(mat, vw, kernel):
     self.states.b.numVgprLocalWriteAddr = 0 if kernel["LocalWriteUseSgprB"] else 1 * self.states.rpla
 
     if self.states.archCaps["HasLDSGT64K"] and not kernel["1LDSBuffer"] and not kernel["LocalWriteUseSgprA"] :
-      if kernel["LdsOffsetA_Blk"]>=131072:
+      if kernel["LdsOffsetA_Blk"]>=131072 or kernel["LdsNumElementsAlignedA"]>=131072:      
         self.states.a.numVgprLocalReadAddr =3* self.states.rpla
```
nakajee, line 3951: Not directly related to this change, but numVgprLocalReadAddr should not be affected by LocalWriteUseSgprA.
I would suggest to use 
        self.states.a.numVgprLocalReadAddr *=3
here and remove not kernel["LocalWriteUseSgprA"] at line 3945.

nakajee: LocalWriteUseSgprA is already 0 in case of kernel["LocalWriteUseSgprA"].
So, we can simply use *=3 for both numVgprLocalWriteAddr and numVgprLocalReadAddr without checking LocalWriteUseSgprA again.

tensilelite/Tensile/KernelWriter.py
```diff
@@ -3943,18 +3943,18 @@ def readWriteVectors(mat, vw, kernel):
...
         self.states.a.numVgprLocalWriteAddr = 3* self.states.rpla
-      elif kernel["LdsOffsetA_Blk"]>=65536:
+      elif kernel["LdsOffsetA_Blk"]>=65536 or kernel["LdsNumElementsAlignedA"]>=65536: 
         self.states.a.numVgprLocalReadAddr =2* self.states.rpla
         self.states.a.numVgprLocalWriteAddr = 2* self.states.rpla
 
     if self.states.archCaps["HasLDSGT64K"] and not kernel["1LDSBuffer"] and not kernel["LocalWriteUseSgprB"] :
-      if kernel["LdsOffsetA_Blk"]>=131072:
+      if kernel["LdsOffsetA_Blk"]>=131072 or kernel["LdsNumElementsAlignedB"]>=131072:
```
jichangjichang: Do we really need to consider LdsOffsetA_Blk for b.numVgprLocalReadAddr ?

tensilelite/Tensile/KernelWriter.py
b-shi: > remove not kernel["LocalWriteUseSgprA"] at line 3945.

@adityalj, that `not kernel["LocalWriteUseSgprA"]` check was causing some failures in anther PR. Could you bring in the num local read/write vgpr checks from this [commit](https://github.com/b-shi/hipBLASLt/commit/4acd6e879303fe7468a88e58706c9614f56c2f92) on top of your changes here?

tensilelite/Tensile/KernelWriter.py
alex391a: I think we need to consider LdsOffsetA_Blk when EPS=1 ?

tensilelite/Tensile/KernelWriter.py
```diff
@@ -3943,18 +3943,18 @@ def readWriteVectors(mat, vw, kernel):
     self.states.b.numVgprLocalWriteAddr = 0 if kernel["LocalWriteUseSgprB"] else 1 * self.states.rpla
 
     if self.states.archCaps["HasLDSGT64K"] and not kernel["1LDSBuffer"] and not kernel["LocalWriteUseSgprA"] :
```
alex391a:
```suggestion
    if self.states.archCaps["HasLDSGT64K"] and not kernel["LocalWriteUseSgprA"] :
```
We still need more VGPRs when LdsNumElementsAligned is large, even if 1LDSBuffer is enabled.

tensilelite/Tensile/KernelWriter.py
```diff
@@ -3943,18 +3943,18 @@ def readWriteVectors(mat, vw, kernel):
...
-      if kernel["LdsOffsetA_Blk"]>=131072:
+      if kernel["LdsOffsetA_Blk"]>=131072 or kernel["LdsNumElementsAlignedA"]>=131072:
```
alex391a:
```suggestion
      if (kernel["LdsOffsetA_Blk"]>=131072 and kernel["ExpandPointerSwap"]) or kernel["LdsNumElementsAlignedA"]>=131072:      
```

alex391a: Similar change is needed for other cases.

alex391a: When ExpandPointerSwap is enabled, we need to add LdsOffsetA_Blk in the offset.

alex391a (approved):

msujon-AMD (approved):

PR #1829, opened 2025-03-26
Fix bad sparse key in mi9 to mi4 conversion

bstefanuk:
In the newest code to update MI-9 to MI-4 format, an extra sparsity key was added. This removes that key and ensures that tuning files remain consistent.

### DIFF
--- a/tensilelite/Tensile/SolutionStructs/Validators/MatrixInstruction.py
+++ b/tensilelite/Tensile/SolutionStructs/Validators/MatrixInstruction.py
@@ -83,7 +83,7 @@ def matrixInstructionToMIParameters(
 
     result["WavefrontSize"] = wavefrontSize
     if workGroup:
-      # NOTE: Typically, the WorkGroup is set to the default [16, 16, 1] for a 
+      # NOTE: Typically, the WorkGroup is set to the default [16, 16, 1] for a
       # length-9 matrix instruction. However, some custom kernel solutions used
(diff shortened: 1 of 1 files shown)


AlexBrownAMD (approved):

KKyang (approved):

vin-huang:
Hi @bstefanuk 
I'm not sure who or where is going to use this result table, but if the sparse key been removed will this impact the sparse kernel?

jichangjichang (approved):

AlexBrownAMD:
> Hi @bstefanuk I'm not sure who or where is going to use this result table, but if the sparse key been removed will this impact the sparse kernel?

Sparse key still exists in the type. A recent change added it to the regular parameters as well and that was causing naming clashes between newly generated liblogics and the old ones. This change reverts the Sparse key to it's previous location so it shouldn't change the behaviour of any existing sparse kernels.

vin-huang:
> > Hi @bstefanuk I'm not sure who or where is going to use this result table, but if the sparse key been removed will this impact the sparse kernel?
> 
> Sparse key still exists in the type. A recent change added it to the regular parameters as well and that was causing naming clashes between newly generated liblogics and the old ones. This change reverts the Sparse key to it's previous location so it shouldn't change the behaviour of any existing sparse kernels.

@AlexBrownAMD Thank you for your explanation. I have no further questions.

vin-huang (approved):

PR #1833, opened 2025-03-27
Add some DTL related optimizations

b-shi:
This PR adds some DTL-related optimizations:
- Enable `PGR=2` with DTL
   - Removed rejection condition and fixed bug causing accuracy issues.
- Add support for LDS padding with DTL. 
  - Add option (`StoreSwapAddr`) to store second lds buffer offset in {s,v}gprs when `(single lds buffer size) > (maxLDS size) /2`


tensilelite/Tensile/Tests/common/gemm/gfx950/dtl.yaml
```diff
@@ -73,8 +72,8 @@ BenchmarkProblems:
         - StorePriorityOpt: [0]
         - VectorStore: [-1]
         - StoreSyncOpt: [0]
-        - LdsPadA: [-1]
-        - LdsPadB: [-1]
+        - LdsPadA: [0, 8, 16]
```
msujon-AMD, line 75: With -1, we will get our best guess for each matrix instructions, right? 
We no longer tune this parameter AFAIK.

msujon-AMD (dismissed):

tensilelite/Tensile/Tests/common/gemm/gfx950/dtl.yaml
b-shi: Ah, good question, it looks like it will choose padding based on calculations from non-dtl cases. I'm not sure if we can select the best padding based on matrix instruction since we can't add padding in between elements written by each thread.

Thoughts on just making the padding equal to the local read instruction width (i.e. 4,8,16 bytes)?

msujon-AMD: > Ah, good question, it looks like it will choose padding based on calculations from non-dtl cases. I'm not sure if we can select the best padding based on matrix instruction since we can't add padding in between elements written by each thread.
> 
> Thoughts on just making the padding equal to the local read instruction width (i.e. 4,8,16 bytes)?

Yea, sounds good. We need good default value since tuning script may have -1 by default.

msujon-AMD (dismissed):

tensilelite/Tensile/Components/LocalRead.py
```diff
@@ -215,7 +215,8 @@ def __call__(self, writer, kernel, bufferIdx, iui, epsi, tP):
             highBits = 0
             for tIdx in range(0, numberMTilesPerWave):
                 offset_val = (tP["localReadOffset"]+MIWaveGroupShape[tile01]*tIdx) * tP["bpeDS"]
-                if (kernel["LdsBlockSizePerPad%s"%tc] != 0) and (kernel["LdsPad%s"%tc] != 0):
+                # Disable this calculation for DTL, since we are only doing padding with M0
+                if (kernel["LdsBlockSizePerPad%s"%tc] != 0) and (kernel["LdsPad%s"%tc] != 0) and not kernel["DirectToLds"]:
```
nakajee: I suggest to use kernel["DirectToLds%s"%tc] anywhere out of Soluitons.py.
Precise meaning of kernel["DirectToLds"] is kernel["DirectToLdsA"] or kernel["DirectToLdsB"].
It does not consider tc ="MetaData" case.
We need to make sure this part of code works in case tc = "MetaData".
On top of that, kernel["DirectToLds"]=1 does not always mean kernel["DirectToLdsA"]=kernel["DirectToLdsB"]=1.
If DirectToVgprA is enabled with kernel["DirectToLds"], kernel["DirectToLdsA"] will be 0 and kernel["DirectToLdsB"] will be 1.

tensilelite/Tensile/KernelWriterAssembly.py
```diff
@@ -4106,6 +4136,20 @@ def lraDeclareAddresses(self, kernel, tP):
           comment="Final vgprLocalReadAddr%s+2 Offset Plus 128K"%tP["tensorChar"] ))
     return module
 
+  def lraSwapAddressesForDTLPad(self, kernel, tP):
+    module = Module("lraSwapAddressesForDTLPad")
+
+    tc = tP["tensorChar"]
+    if kernel["StoreSwapAddr"]:
+      ldsSecondBufferOffset = kernel["LdsNumElementsAlignedA"] + kernel["LdsNumElementsAlignedB"]
```
nakajee: Don't we need to consider LdsNumElementsAlignedMetadata?

tensilelite/Tensile/SolutionStructs/Solution.py
```diff
@@ -1660,10 +1661,9 @@ def calcLdsPad(lrvw: int, isaInfoMap: Dict[str, IsaInfo]) -> int:
           assert(ldsPadM >= 0)
 
         # set ldsPadA,B=0 for DirectToLds or DirectToVgpr
```
nakajee: Please remove "DirectToLds or" from this line

tensilelite/Tensile/SolutionStructs/Solution.py
```diff
@@ -1701,11 +1701,18 @@ def calcLdsBlockSizePerPad(lrvw: int) -> int:
             LdsBlockSizePerPadB = 0
 
         # set LdsBlockSizePerPadA,B=0 for DirectToLds or DirectToVgpr
```
nakajee: Please remove "DirectToLds or" from this line

tensilelite/Tensile/SolutionStructs/Solution.py
```diff
@@ -1701,11 +1701,18 @@ def calcLdsBlockSizePerPad(lrvw: int) -> int:
...
-        if state["DirectToLds"] or state["DirectToVgprA"]:
+        if state["DirectToVgprA"]:
           LdsBlockSizePerPadA = 0
-        if state["DirectToLds"] or state["DirectToVgprB"]:
+        if state["DirectToVgprB"]:
           LdsBlockSizePerPadB = 0
 
+        if state["DirectToLds"]:
```
nakajee: As I said, DirectToLds can be combined with DirectToVgpr.
Either check DirectToLdsA,B separately or do this before checking DirectToVgprA,B above.

tensilelite/Tensile/SolutionStructs/Solution.py
```diff
@@ -1602,6 +1604,10 @@ def calcLdsPad(lrvw: int, isaInfoMap: Dict[str, IsaInfo]) -> int:
               optPadA *= 2
             if readRegsB == 4 or readRegsB == 1:
               optPadB *= 2
+        if state["DirectToLdsA"] and ldsPadA == -1:
```
nakajee: Don't we need to consider UnrollMajorLDSA/B here?

tensilelite/Tensile/SolutionStructs/Solution.py
```diff
@@ -1701,11 +1701,18 @@ def calcLdsBlockSizePerPad(lrvw: int) -> int:
...
+        if state["DirectToVgprA"]:
           LdsBlockSizePerPadA = 0
-        if state["DirectToLds"] or state["DirectToVgprB"]:
+        if state["DirectToVgprB"]:
           LdsBlockSizePerPadB = 0
 
+        if state["DirectToLds"]:
+          bpeA = state["ProblemType"]["DataTypeA"].numBytes()
+          bpeB = state["ProblemType"]["DataTypeB"].numBytes()
+          LdsBlockSizePerPadA = (state[f"GlobalReadVectorWidthA"] * bpeA) * state["WavefrontSize"]
```
nakajee: Don't we need to consider UnrollMajorLDSA/B here?

tensilelite/Tensile/KernelWriter.py
```diff
@@ -3967,11 +3981,19 @@ def readWriteVectors(mat, vw, kernel):
...
     # do not allocate local write address register if DirectToVgpr is enabled
-    if kernel["DirectToVgprA"]:
+    if kernel["DirectToVgprA"] or kernel["DirectToLdsA"]:
       self.states.a.numVgprLocalWriteAddr = 0
-    if kernel["DirectToVgprB"]:
+    if kernel["DirectToVgprB"] or kernel["DirectToLdsB"]:
       self.states.b.numVgprLocalWriteAddr = 0
 
+    if kernel["StoreSwapAddr"]:
+      self.states.a.numVgprLocalReadSwapAddr = 1
```
nakajee: Why this is always 1?
Should this be same as numVgprLocalReadAddr?

nakajee:
Not directly related to this change, but I think we need to ensure ConvertAfterDS is true with DirectToLds.
So far, we do not have a reject condition to disable type conversion (like A=fp8, B=fp16 case).
Either disable TypeConversion or always use ConvertAfterDS with DirectToLds.
I forgot to add such a condition in isDirectToLdsDoable. 
I added a similar condition to isDirectToVgprDoable.

tensilelite/Tensile/Components/LocalRead.py
b-shi: > If DirectToVgprA is enabled with kernel["DirectToLds"],  ...

Does DTV take priority over DTL?

tensilelite/Tensile/KernelWriterAssembly.py
b-shi: Oops, yes, I'm assuming the total size for a single buffer would be `sizeA + sizeB + sizeMetaData`?

tensilelite/Tensile/KernelWriter.py
b-shi: My understanding is that we only have multiple lra/lwa to handle cases where the instruction offsets (LDS) would be over 64KB - I'm not sure if there are other use cases. For `StoreSwapAddr` the instruction offset will be less than 64KB since we move the offsets to the vgprs when trying to swap between lds buffers.

b-shi:
> Not directly related to this change, but I think we need to ensure ConvertAfterDS is true with DirectToLds. So far, we do not have a reject condition to disable type conversion (like A=fp8, B=fp16 case). Either disable TypeConversion or always use ConvertAfterDS with DirectToLds. I forgot to add such a condition in isDirectToLdsDoable. I added a similar condition to isDirectToVgprDoable.

Thanks for taking the time to look through the changes! :)

Regarding the `UnrollMajorLDS` comments, I am not sure if we need to consider it. This is mainly due to DTL supporting very limited forms of padding. Does `UnrollMajorLDS{X}` mean that the layout for `X` is contiguous in the accumulation dimension?

nakajee:
> > Not directly related to this change, but I think we need to ensure ConvertAfterDS is true with DirectToLds. So far, we do not have a reject condition to disable type conversion (like A=fp8, B=fp16 case). Either disable TypeConversion or always use ConvertAfterDS with DirectToLds. I forgot to add such a condition in isDirectToLdsDoable. I added a similar condition to isDirectToVgprDoable.
> 
> Thanks for taking the time to look through the changes! :)
> 
> Regarding the `UnrollMajorLDS` comments, I am not sure if we need to consider it. This is mainly due to DTL supporting very limited forms of padding. Does `UnrollMajorLDS{X}` mean that the layout for `X` is contiguous in the accumulation dimension?

Basically,
  UnrollMajorLDSA means TransA=T
  UnrollMajorLDSB means TransB=N
This assumes we do not do transpose when loading data from global memory and storing to LDS.
UnrollMajorLDSA/B represents data layout in LDS.
Tensile has an option to do transpose at LDS write, but I think it is not enabled for hipblaslt.
If we do transpose before LDS write, UnrollMajorLDSA will be False even if TransA is T.

tensilelite/Tensile/Components/LocalRead.py
nakajee: Yes.
Please check this part of the code.
https://github.com/ROCm/hipBLASLt/blob/develop/tensilelite/Tensile/SolutionStructs/Solution.py#L2598

If DTVA is true, isDirectToLds is not called and force DirectToLdsA=False.

tensilelite/Tensile/KernelWriterAssembly.py
nakajee: Please check the code calculating lds offset.
We probably need to consider alignment for xor address swapping.

tensilelite/Tensile/SolutionStructs/Solution.py
nakajee: Regarding your comment below
> Regarding the UnrollMajorLDS comments, I am not sure if we need to consider it. This is mainly due to DTL supporting very limited forms of padding. Does UnrollMajorLDS{X} mean that the layout for X is contiguous in the accumulation dimension?

Basically, LDS padding strategy is different for UnrollMajorLDS = False and True cases.
UnrollMajorLDS = False case, we basically do not need LDS padding because Each thread reads data horizontally and does not cause LDS bank conflicts.
(but, we have some exceptions for this logic. You can learn more details from LDS_pad auto calculation code)
So, it is better not to enable LDS pad and LdsBlockSizePerPadA in UnrollMajorLDS = False case.
(means set 0 to them)

tensilelite/Tensile/KernelWriter.py
nakajee: If we do not need multiple vgprs for local read/write address for StoreswapAddr, we need to disable allocating multiple vgprs for local read/write if StoreswapAddr is true.

tensilelite/Tensile/SolutionStructs/Solution.py
b-shi: Thanks for the explanations, I've updated the `LdsBlockSizePerPadX` calculations to use the existing logic and ~~round to the nearest instruction load size ( `bytes loaded per thread * wave size`) when DTL is enabled~~.

edit: Actually padding does not need to be a multiple of the DTL load size, just larger or equal to.

tensilelite/Tensile/KernelWriterAssembly.py
b-shi: Okay, it looks like this exact calculation is already done in `Solutions` - I've updated to use those values instead of recomputing here... :)

msujon-AMD (dismissed):

tensilelite/Tensile/Components/LocalRead.py
```diff
@@ -219,7 +219,8 @@ def __call__(self, writer, kernel, bufferIdx, iui, epsi, tP):
             highBits = 0
             for tIdx in range(0, numberMTilesPerWave):
                 offset_val = (tP["localReadOffset"]+MIWaveGroupShape[tile01]*tIdx) * tP["bpeDS"]
-                if (kernel["LdsBlockSizePerPad%s"%tc] != 0) and (kernel["LdsPad%s"%tc] != 0):
+                # Disable this calculation for DTL, since we are only doing padding with M0
+                if (kernel["LdsBlockSizePerPad%s"%tc] != 0) and (kernel["LdsPad%s"%tc] != 0) and not kernel["DirectToLds%s"%tc]:
```
nakajee: Do we really need this condition?
DTL is related to write only.
For local read, it should not matter if it is DTL or non DTL.

tensilelite/Tensile/SolutionStructs/Solution.py
```diff
@@ -1716,21 +1737,31 @@ def calcLdsNumBytes(ldsPadA: int, LdsBlockSizePerPadA: int, ldsPadB: int, LdsBlo
...
         ldsNumBytesAlignedA = roundUpToNearestMultiple(ldsNumBytesA, ldsAlign)
-
         # DirectToVgpr case, set 0 to lds related variables
         if state["DirectToVgprA"]:
           ldsNumBytesA = 0
           ldsNumBytesAlignedA = 0
 
         if state["UnrollMajorLDSB"]:
-          ldsNumBytesB = (state["_DepthUB"] + ldsPadB) * state["MacroTileB"] * bpeB
+          if state["DirectToLdsB"]:
```
nakajee: We should not need this extra calculation if LdsBlockSizePerPad is properly set for DTL case.

nakajee: We should only rely on LdsPad and LdsBlockSizePerPad to calculate LDS size.

tensilelite/Tensile/SolutionStructs/Solution.py
```diff
@@ -1618,10 +1627,10 @@ def calcLdsPad(lrvw: int, isaInfoMap: Dict[str, IsaInfo]) -> int:
               else:
                 ldsPadA = state["VectorWidthA"]
           else:
-            ldsPadA = max(state["GlobalReadVectorWidthA"],optPadA)
-            ## turn-off padding for directToLds
             if state["DirectToLdsA"]:
-              ldsPadA = 0
+              ldsPadA = lrvw
```
nakajee: It is not a good idea to ignore optPadA/B for DTL.
(optPad is for UnrollMajorLDS only)
As long as DepthU * bpe is larger than or equal to DLT load unit (per 64 threads), optPadA/B should work as non DTL.
If not, not sure but LDSPad might be useless if DepthU * bpe is less than DLT load unit

I think we can ignore GlobalReadVectorWidthA/B for DTL.
(LDS write is always continuous for 64 threads and padding should not matter)

tensilelite/Tensile/SolutionStructs/Solution.py
```diff
@@ -1604,6 +1612,7 @@ def calcLdsPad(lrvw: int, isaInfoMap: Dict[str, IsaInfo]) -> int:
               optPadA *= 2
             if readRegsB == 4 or readRegsB == 1:
               optPadB *= 2
+
         if ldsPadA == -1:
           if not state["UnrollMajorLDSA"]:
```
nakajee: This LdsPad calculation for UnrollMajorLDS=False is not always applicable to DTL.
As long as MT0(for A)/MT1(for B) * bpe is larger than or equal to DLT load unit (per 64 threads), this should work as non DTL.
If not, not sure but LDSPad might be useless if MT0(for A)/MT1(for B) * bpe is less than DLT load unit.

tensilelite/Tensile/Components/LocalRead.py
b-shi: Ah you're right, the DTL check is not actually needed, removed.

tensilelite/Tensile/SolutionStructs/Solution.py
b-shi: :facepalm: :facepalm: ... good point! Fixed.

tensilelite/Tensile/SolutionStructs/Solution.py
b-shi: > not sure but LDSPad might be useless if DepthU * bpe is less than DLT load unit

Our initial motivation to enable padding for DTL was to cover such a case. Neighboring threads were accessing in 128 byte strides (DU=64, bpe=2, DTL load unit = 1KB) and we had 2x 8-way bank conflicts for each group of 16 threads. Each read was accessing LDS data written by different global loads so padding by `lrvw` every DTL load unit helped mitigate the issue.

Okay, I see your point about `optPad`. I've set the default ldsPad value to be `max(lrvw, optPad)` :)

tensilelite/Tensile/SolutionStructs/Solution.py
b-shi: Yeah, I think you are right, I don't know if lds padding (for DTL) would be helpful for `UnrollMajorLDS=False`. I've set it to zero (as you suggested), and added a note to check if there are cases where it may help.

tensilelite/Tensile/SolutionStructs/Solution.py
nakajee: I do not think it is correct.
LdsBlockSizePerPadX should be multiple of DTL load unit (for each wave).

tensilelite/Tensile/KernelWriter.py
```diff
@@ -1182,8 +1185,9 @@ def hasAnyDependency(lr: DSLoadInstruction, insts: List[Instruction]):
                                         })
         # if start to schedule localwrite, but still have localreads not scheduled yet,
         # reject to use 1LDSB, since it will write and read same lds buffer at same time.
-        if mfmaIndex > self.states.sync1LdsMfmaIndex and localReadItemsThisLoop and oneBufferScheduling:
-          self.states.overflowedResources = 5
+        #if mfmaIndex > self.states.sync1LdsMfmaIndex and localReadItemsThisLoop and oneBufferScheduling:
+          # TODO: can we remove this restriction?
+          #self.states.overflowedResources = 5
```
nakajee: Does this condition hit if you do not comment out?
I do not expect this actually happens.
We have a logic to generate all remaining local read before local write for one buffer scheduling case.

nakajee: We probably do not need to remove this condition.
We can keep it for sanity check.

tensilelite/Tensile/KernelWriter.py
```diff
@@ -3974,11 +3990,19 @@ def readWriteVectors(mat, vw, kernel):
...
     # do not allocate local write address register if DirectToVgpr is enabled
-    if kernel["DirectToVgprA"]:
+    if kernel["DirectToVgprA"] or kernel["DirectToLdsA"]:
       self.states.a.numVgprLocalWriteAddr = 0
-    if kernel["DirectToVgprB"]:
+    if kernel["DirectToVgprB"] or kernel["DirectToLdsB"]:
       self.states.b.numVgprLocalWriteAddr = 0
 
+    if kernel["StoreSwapAddr"]:
+      self.states.a.numVgprLocalReadSwapAddr = 1
```
nakajee: We do not need to allocate numVgprLocalReadSwapAddr in DirectToVgpr case.

nakajee: Same for WriteSwap

tensilelite/Tensile/KernelWriterAssembly.py
```diff
@@ -8405,7 +8465,14 @@ def localWriteSwapOffsets(self, kernel, internalPointerSwap, tP):
                 src1=sgpr("LocalWriteAddr%s"%tPM["tensorChar"]), \
                 comment="swap Red Blk SGPR"))
           else:
-            numLwa = self.states.m.numVgprLocalWriteAddr
+            numLwa = 0;
```
nakajee: I think this part of the code is for metadata only. No need to add A/B code here.

tensilelite/Tensile/KernelWriterAssembly.py
```diff
@@ -8368,16 +8406,38 @@ def localWriteSwapOffsets(self, kernel, internalPointerSwap, tP):
...
-      if internalPointerSwap:
-        tP["localWriteSwapByteOffset"] = 0 if tP["localWriteSwapByteOffset"] else kernel["LdsOffsetA_Blk"]
-        module.addComment1("(EPS=1) local write swap internal offset -> %u" % tP["localWriteSwapByteOffset"])
-      else:
-        if kernel["LocalWriteUseSgpr%s"%tc]:
-          module.add(SXorB32(
+      if internalPointerSwap or kernel["StoreSwapAddr"]:
+
+        if not kernel["StoreSwapAddr"]:
+          if kernel["LocalWriteUseSgpr%s"%tc]:
```
nakajee, line 8432: I suggest to unify LocalWriteUseSgpr case code.
No need to duplicate almost same code at 3 locations.
Only difference is src0 for StoreSwapAddr case.
We just need to switch src0 for StoreSwapAddr.

By the way, LocalWriteUseSgpr code did not exist in internalPointerSwap before your change.
Was this missing before your change?

tensilelite/Tensile/SolutionStructs/Solution.py
b-shi: Oops, yes, I mistook lds padding and ldsblocksizeperpad. Reverted.

tensilelite/Tensile/KernelWriter.py
b-shi: Yeah, previously it was hit and causing an abort which was why I commented it out - I don't recall the exact issue back then. I just tried uncommenting this check and it seems fine now. Possibly the misc fixes added in this PR resolved the initial issue.

What exactly is this check? I do notice that more solutions are filtered out with it enabled.

tensilelite/Tensile/KernelWriterAssembly.py
b-shi: Good point, refactored.

> By the way, LocalWriteUseSgpr code did not exist in internalPointerSwap before your change.
Was this missing before your change?

No, I added this to fix a test failure - but I don't think its actually needed, removed. I think the initial test failure was resolved when adding additional checks for `numVgprLocal{Read,Write}Addr`

tensilelite/Tensile/KernelWriter.py
```diff
@@ -3974,11 +3990,21 @@ def readWriteVectors(mat, vw, kernel):
...
-    if kernel["DirectToVgprB"]:
+    if kernel["DirectToVgprB"] or kernel["DirectToLdsB"]:
       self.states.b.numVgprLocalWriteAddr = 0
 
+    if kernel["StoreSwapAddr"]:
+      if self.states.a.numVgprLocalReadAddr > 0:
+        self.states.a.numVgprLocalReadSwapAddr = 1
+      if self.states.b.numVgprLocalReadAddr > 0:
+        self.states.b.numVgprLocalReadSwapAddr = 1
+      if not kernel["LocalWriteUseSgprA"]:
```
nakajee: I think num check is necessary for write as well.

nakajee:
Could you add some test cases with DTL + DTV?

b-shi:
> Could you add some test cases with DTL + DTV?

Yes, but I will mark it as a TODO for now - there are still some work regarding DTV on gfx950 that needs to be done.

tensilelite/Tensile/KernelWriter.py
nakajee: This checks if all local read instructions are scheduled before 1LDS buffer sync.
If not, local write can overwrite LDS before read by local read.
This happened before adding the logic to force to generate all remaining local read at 1 LDS buffer sync.

nakajee:
> > Could you add some test cases with DTL + DTV?
> 
> Yes, but I will mark it as a TODO for now - there are still some work regarding DTV on gfx950 that needs to be done.

OK. I got it.
I do not have further comments for now.
Please get reviewed by others for approval.

tensilelite/Tensile/SolutionStructs/Solution.py
```diff
@@ -959,9 +959,9 @@ def isDirectToLdsDoable(state, tc, isaInfoMap, printRejectionReason: bool):
 
     # so far, DirectToLds does not work well with PGR=2
     # performance is not good and a lot of ds_read for DTL can cause scheduling issue(need fix)
-    if state["PrefetchGlobalRead"] == 2:
-      reject(state, printRejectionReason, "can't use DirectToLds for PrefetchGlobalRead == 2")
-      return False
+    #if state["PrefetchGlobalRead"] == 2:
```
jichangjichang: remove unused code

msujon-AMD (dismissed):

TonyYHsieh:
random failed appear in dtl.yaml 
docker: compute-artifactory.amd.com:5000/mathlibs/ci/gfx950:compute-rocm-npi-mi350-881-ubuntu22

Index:  Device | Reference
[0]  elem=57344 idx=57344: 15.7891!=19
[1]  elem=57345 idx=57345: 19.2969!=12
[2]  elem=57346 idx=57346: 38.6875!=49
[3]  elem=57347 idx=57347: -64.1875!=-61
Found 498 incorrect values in 65536 total values compared.
Check failed in output tensor: [d]3-tensor<Half>( sizes(256, 256, 1), strides(1, 256, 65536) )
0,1/3,9/12,Contraction_l_Alik_Bljk_Cijk_Dijk,"(256,256,1,128)",None,,None,Cijk_Alik_Bljk_HHS_BH_UserArgs_MT128x128x64_MI32x32x1_SN_CLR1_GSU1_GSUC0_GSUWGMRR0_K1_LPA8_LPB8_MIWT2_2_PGR2_PLR2_SU32_SUM0_SUS256_SVW2_VWA2_VWB2_WG64_4_1_WGM8_WGMXCC1_WGMXCCGn1,FAILED,12,1398.1,,0.015625,0.015625,256,1,1,0.015625,1,393216,131072,,,,,,,2025-04-11 10:06:09.987352

TonyYHsieh:
is it possible to seperate each feature to seperate PR?
it is hard to understand which change below to which feature.

tensilelite/Tensile/SolutionStructs/Solution.py
```diff
@@ -2657,8 +2673,14 @@ def subCheckLdsBlockSizePerPad(tc, idx):
       state["LdsOffsetMetadata"] = state["LdsOffsetA"] + state["LdsNumElementsAlignedA"]
       state["LdsOffsetB"] = state["LdsOffsetMetadata"] + state["LdsNumElementsAlignedMetadata"]
 
-      offsetBlk = state["LdsOffsetB"] +  ldsNumBytesAlignedB
-      if offsetBlk > 0:
+      offsetBlk = state["LdsOffsetB"] + ldsNumBytesAlignedB
```
TonyYHsieh, line 2677: is it possible to use offsetBlk to replace int(2**(math.ceil(math.log(offsetBlk, 2)))) for all case??
swapValue = 1_LDS + offset_A
swapValue = swapValue xor offset_A 
offset_A_blk = offset_A xor swapValue 
offset_A  = offset_A_blk xor swapValue 
is it correct?

b-shi: > is it possible to use offsetBlk to replace int(2**(math.ceil(math.log(offsetBlk, 2)))) for all case??

Sorry, I don't understand the question? Do you mean always store the swapValue in a register?

> swapValue = 1_LDS + offset_A
swapValue = swapValue xor offset_A
offset_A_blk = offset_A xor swapValue
offset_A = offset_A_blk xor swapValue

Yes, that is correct. Just checking, in your comment, `offset_A` is first buffer and `offset_A_blk` is second buffer right?

b-shi:
> is it possible to seperate each feature to seperate PR? it is hard to understand which change below to which feature.

This will be tricky because the changes are dependent on each other. There are only two main changes, if it helps I can point out the changes here:
- Enable PGR=2 for DTL (changes in `SIA.py` and removing the restriction in `Solutions.py`)
- All other changes are for implementing `StoreSwapAddr` and LDS Padding - these two are coupled together.

TonyYHsieh:
> > is it possible to seperate each feature to seperate PR? it is hard to understand which change below to which feature.
> 
> This will be tricky because the changes are dependent on each other. There are only two main changes, if it helps I can point out the changes here:
> 
> * Enable PGR=2 for DTL (changes in `SIA.py` and removing the restriction in `Solutions.py`)
> * All other changes are for implementing `StoreSwapAddr` and LDS Padding - these two are coupled together.

StoreSwapAddr looks like ExpandPointerSwap (0), but use two unroll loop body and different swap value.
And StoreSwapAddr is hard to be ture.
      state["StoreSwapAddr"] = (state["PrefetchGlobalRead"] == 2) and \
        (state["1LDSBuffer"] == 0) and \
        (offsetBlk + int(2**(math.ceil(math.log(offsetBlk, 2)))) > depthUConfig.maxLDS)
is it possible to use StoreSwapAddr swap method in ExpandPointerSwap (0)?

b-shi:
> random failed appear in dtl.yaml docker: compute-artifactory.amd.com:5000/mathlibs/ci/gfx950:compute-rocm-npi-mi350-881-ubuntu22
> 
> Index: Device | Reference [0] elem=57344 idx=57344: 15.7891!=19 [1] elem=57345 idx=57345: 19.2969!=12 [2] elem=57346 idx=57346: 38.6875!=49 [3] elem=57347 idx=57347: -64.1875!=-61 Found 498 incorrect values in 65536 total values compared. Check failed in output tensor: [d]3-tensor( sizes(256, 256, 1), strides(1, 256, 65536) ) 0,1/3,9/12,Contraction_l_Alik_Bljk_Cijk_Dijk,"(256,256,1,128)",None,,None,Cijk_Alik_Bljk_HHS_BH_UserArgs_MT128x128x64_MI32x32x1_SN_CLR1_GSU1_GSUC0_GSUWGMRR0_K1_LPA8_LPB8_MIWT2_2_PGR2_PLR2_SU32_SUM0_SUS256_SVW2_VWA2_VWB2_WG64_4_1_WGM8_WGMXCC1_WGMXCCGn1,FAILED,12,1398.1,,0.015625,0.015625,256,1,1,0.015625,1,393216,131072,,,,,,,2025-04-11 10:06:09.987352

Thanks for checking, we were missing some waits for the second set of prefetched global reads, fixed.

b-shi:
> StoreSwapAddr looks like ExpandPointerSwap (0), but use two unroll loop body and different swap value. And StoreSwapAddr is hard to be ture. state["StoreSwapAddr"] = (state["PrefetchGlobalRead"] == 2) and (state["1LDSBuffer"] == 0) and (offsetBlk + int(2**(math.ceil(math.log(offsetBlk, 2)))) > depthUConfig.maxLDS) is it possible to use StoreSwapAddr swap method in ExpandPointerSwap (0)?

> And StoreSwapAddr is hard to be ture. 

Yes, if possible we will try to use inline constant instead of storing the swap values in SGPRs. `StoreSwapAddr` was added to handle cases where we cannot inline the constant (i.e. large macro tile + double lds buffers).

> is it possible to use StoreSwapAddr swap method in ExpandPointerSwap (0)?

Do you mean have `ExpandPointerSwap` support both inline constant swap values and storing swap values in SGPRs? If so, that would not really be DTL related - its a general improvement that should probably have its own PR.

KKyang (approved):

msujon-AMD (approved):

PR #1837, opened 2025-03-27
Only override internal keys if not already written

daineAMD:
- InternalKeys includes ["UseSgprForGRO","VectorStore"]
- When creating the Logic file, if UseSgprForGRO == -1, _UseSgprForGRO gets set to some value, say 0. These are both written the the logic file. From here, TensileCreateLibrary then _also_ updates _UseSgprForGRO to -1 instead of using the value set in the Logic file. I believe here, we are getting a mixture of interpreting _UseSgprForGRO as -1 (passing boolean comparisons where the code thinks it should be in the set [0, 1]

### DIFF
--- a/tensilelite/Tensile/SolutionStructs/Solution.py
+++ b/tensilelite/Tensile/SolutionStructs/Solution.py
@@ -1049,6 +1049,7 @@ class Solution(collections.abc.Mapping):
     state["AssignedDerivedParameters"] = False
 
     for s in Solution.InternalKeys:
+      if '_'+s not in state:
         state['_'+s] = state[s]
         #del state[s]
 


AlexBrownAMD (approved):
Looks good! Thanks for this bug fix!

jichangjichang (approved):
Please also ask David to check this PR.

davidd-amd:
> Please also ask David to check this PR.

Thanks for adding me to the review @jichangjichang!

davidd-amd:
@daineAMD were we winding up with inconsistent values of `UseSgprForGRO` and `_UseSgprForGRO` between when the logic file was created and when we call TensileCreateLibrary? How do we know which one to prefer?

@jichangjichang @AlexBrownAMD @daineAMD do we know why we need both keys? That is one with and without `_`. Should keys with `_` go into the logic file? In python, functions with a leading `_` are considered an implementation detail and not intended to be exported. I wonder if the same thing was intended for keys with `_`.

The docs in the code only describe the meaning of -1:

```
    # Converting VGPR GRO into SGPR GRO is usually a win
    # However, the mode may exhaust all available SGPR, in particular for large unroll
    # -1 attempt to use a heuristic to determine when the tile size will use too many SGPR and fall back to VGPR
    "UseSgprForGRO": [-1, 0, 1],
```

What is the meaning of 0 and 1?

KKyang:
-1 is auto, 0 is disable, and 1 is enable in most cases in the fork parameters.

Not sure but normally any configurations starts with ``_`` is not used in SolutionStruct and will be override in the generator. Since those keys are saved in a dict, they were saved to the file with other parameters.

daineAMD:
> @daineAMD were we winding up with inconsistent values of `UseSgprForGRO` and `_UseSgprForGRO` between when the logic file was created and when we call TensileCreateLibrary? How do we know which one to prefer?
> 

UseSgprForGRO is not updated and remains -1, my reading of the original implementation is that it intended to keep `UseSgprForGRO` untouched, and for `_UseSgprForGRO` to be used for logic and sgpr calculations. TensileCreateLibrary, however, doesn't use the same logic and incorrectly copies `UseSgprForGRO` (even when it's -1) to the _ version without doing the "automated" part, which causes the issue.

> @jichangjichang @AlexBrownAMD @daineAMD do we know why we need both keys? That is one with and without `_`. Should keys with `_` go into the logic file? In python, functions with a leading `_` are considered an implementation detail and not intended to be exported. I wonder if the same thing was intended for keys with `_`.
> 

I'm fine with removing `_UseSgprForGRO` completely myself. Can probably do the same with `_VectorStore` but I'll have to go through the code path again. The change in this PR is just less intrusive in case there was reasoning to leave `UseSgprForGRO` untouched.

davidd-amd:
> > @daineAMD were we winding up with inconsistent values of `UseSgprForGRO` and `_UseSgprForGRO` between when the logic file was created and when we call TensileCreateLibrary? How do we know which one to prefer?
> 
> UseSgprForGRO is not updated and remains -1, my reading of the original implementation is that it intended to keep `UseSgprForGRO` untouched, and for `_UseSgprForGRO` to be used for logic and sgpr calculations. TensileCreateLibrary, however, doesn't use the same logic and incorrectly copies `UseSgprForGRO` (even when it's -1) to the _ version without doing the "automated" part, which causes the issue.
> 
> > @jichangjichang @AlexBrownAMD @daineAMD do we know why we need both keys? That is one with and without `_`. Should keys with `_` go into the logic file? In python, functions with a leading `_` are considered an implementation detail and not intended to be exported. I wonder if the same thing was intended for keys with `_`.
> 
> I'm fine with removing `_UseSgprForGRO` completely myself. Can probably do the same with `_VectorStore` but I'll have to go through the code path again. The change in this PR is just less intrusive in case there was reasoning to leave `UseSgprForGRO` untouched.

Souds good thanks @daineAMD! Maybe we should create an issue for `_VectorStore` and deal with that separately.

PR #1850, opened 2025-03-31
perf: Supported multi-samples for PTS to calc significance value (p-value)

solaslin:
- [x] Able to set multiple samples: run the whole benchmark for multi-times. `e.g. ./hipblaslt-perf --suite all --samples 10 (default 5)`
- [x] Modified the csv data format, added sample_num, mean, median, samples values

The new csv looks like this:
[amax] - new: sample_num, mean_us, median_us, us result *5 
image

[matmul] - new: sample_num, mean_GB/s, median_GB/s, mean_gflops, median_gflops, mean_us, median_us
and then 5 GB/s results, 5 gflops results, 5 us result. 
image

Note that sample_num =5 doesn't mean that each problem is bench with 5 iterations.
**Each problem still follows the "cold_iter" & "hot_iter" setting. The sample_num is another outer loop, which repeats 5 times.**

### DIFF
--- a/clients/scripts/performance/bench_sample.py
+++ b/clients/scripts/performance/bench_sample.py
@@ -1,0 +1,95 @@
+# Copyright (C) 2025 Advanced Micro Devices, Inc. All rights reserved.
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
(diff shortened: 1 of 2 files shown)


KKyang (approved):

Serge45 (approved):

solaslin:
precheckin is a known issue and fixed in another PR. Performance Job will be updated by PTS. Merging

PR #1852, opened 2025-03-31
Update README.md and set Python_ROOT to /usr if not set

KKyang:
The CMakeLists.txt is a tool inside TensileLite and is currently not used anywhere except local development.


hcman2 (approved):
love and peace

aazz44ss (approved):

PR #1864, opened 2025-04-01
Windows enablement

amd-mtrifuno:


tensilelite/Tensile/Common/GlobalParameters.py
```diff
@@ -539,16 +541,22 @@ def assignGlobalParameters(config, isaInfoMap: Dict[IsaVersion, IsaInfo]):
...
         globalParameters["ROCmPath"] = os.environ.get("TENSILE_ROCM_PATH")
-    if os.name == "nt" and "HIP_DIR" in os.environ:
-        globalParameters["ROCmPath"] = os.environ.get("HIP_DIR")  # windows has no ROCM
+    if os.name == "nt":
+        possibleHipPaths = ('HIP_DIR', 'HIP_PATH',)
+
+        for p in possibleHipPaths:
+            if p in os.environ:
+                globalParameters["ROCmPath"] = os.environ.get(p) # windows has no ROCM
+                break # use the first non-null one
```
bstefanuk: We're iteratively moving away from depending on `assignGlobalParameters` for the setup and management of paths. Much of this functionality is implemented without dependence on global state in `Tensile.Toolchain.Validators`. I would recommend taking a look through that module and using it for this purpose.

tensilelite/Tensile/Common/GlobalParameters.py
```diff
@@ -287,6 +288,7 @@
 )
 
 globalParameters["UseEffLike"] = True  # Set to False to use winnerGFlops as the performance metric
+globalParameters["MakeProgram"] = None
```
bstefanuk: Ideally, we don't need to introduce any new global state into the program. This is a requirement for the larger build improvement work we're focusing on.

tensilelite/Tensile/Common/GlobalParameters.py
```diff
@@ -579,10 +587,16 @@ def assignGlobalParameters(config, isaInfoMap: Dict[IsaVersion, IsaInfo]):
...
+        if os.name == "nt":
+            os.environ['HIP_USE_PERL_SCRIPTS'] = '1'
+            compiler = os.environ.get("HIP_PATH") + '/bin/hipcc'
+            compileArgs = ['perl'] + [compiler] + ['--version']
+            output = subprocess.run(compileArgs, check=True, stdout=subprocess.PIPE).stdout.decode()
+        else:
+            compiler = "hipcc"
+            output = subprocess.run(
+                [compiler, "--version"], check=True, stdout=subprocess.PIPE
+            ).stdout.decode()
```
bstefanuk: Similar as above. Perhaps this is a good chance to deprecate this section of code altogether 🙂

tensilelite/Tensile/Toolchain/Validators.py
```diff
@@ -115,7 +115,10 @@ class ToolchainDefaults(NamedTuple):
     CXX_COMPILER = osSelect(linux="amdclang++", windows="clang++.exe")
     C_COMPILER = osSelect(linux="amdclang", windows="clang.exe")
     OFFLOAD_BUNDLER = osSelect(linux="clang-offload-bundler", windows="clang-offload-bundler.exe")
-    DEVICE_ENUMERATOR = osSelect(linux="rocm_agent_enumerator" if isRhel8() else "amdgpu-arch", windows="hipinfo")
+    if os.name == 'nt':
+        DEVICE_ENUMERATOR = "hipInfo"
+    else:
+        DEVICE_ENUMERATOR = "rocm_agent_enumerator" if isRhel8() else "amdgpu-arch"
```
bstefanuk: How come we've removed the usage of `osSelect` to match the pattern of the other toolchain components.

tensilelite/Tensile/Toolchain/Component.py
```diff
@@ -262,13 +262,14 @@ def compress(self, srcPath: str, destPath: str, target: str):
...
         """
+        input = "/dev/null" if os_name != "nt" else "NUL"
         args = [
             self._component_path,
             "--compress",
             "--type=o",
             "--bundle-align=4096",
             f"--targets=host-x86_64-unknown-linux-gnu,hipv4-amdgcn-amd-amdhsa-unknown-{target}",
-            "--input=/dev/null",
+            f"--input={input}",
```
bstefanuk: Not a big deal, but since this is pointing to the null device on linux and windows, perhaps we rename input here to "devnull" or something more semantically meaningful.

tensilelite/Tensile/Toolchain/Validators.py
amd-mtrifuno: This part of the code before modification is:
DEVICE_ENUMERATOR = osSelect(windows="hipinfo.exe", linux="rocm_agent_enumerator" if isRhel8() else "amdgpu-arch") 

Although the osSelect function ultimately uses the "windows" parameter, Python evaluates all function arguments before calling function, and the expression in the linux parameter is evaluated regardless of the operating system. isRhel8() is executed even if its result is not later used, and this function fails on Windows.

tensilelite/Tensile/Toolchain/Component.py
amd-mtrifuno: input renamed to devnull

tensilelite/Tensile/Common/GlobalParameters.py
amd-mtrifuno: globalParameters["MakeProgram"] removed.

tensilelite/Tensile/Common/GlobalParameters.py
amd-mtrifuno: I replaced this part of code with call to `_windowsSearchPaths/_posixSearchPaths` function from `Tensile.Toolchain.Validators` file. Check if this change is enough to remove the dependency on global state. 
This also required change in the ClientExecutable.py file on line 84 and 85, where I obtained paths to the compiler with the `validateToolchain` function.

tensilelite/Tensile/Common/GlobalParameters.py
amd-mtrifuno: Getting compiler path has been replaced by calling a function from `Tensile.Toolchain.Validators` in case this code remains.

bstefanuk:
@amd-mtrifuno Is this a feature targeting 6.5? How come we're targeting release staging prior to targeting develop and then cherry picking?

tensilelite/Tensile/Toolchain/Validators.py
bstefanuk: I see. There has been a recent PR that updates how `isRhel8` operates (#1873). Can you try to see if the previous usage of `osSelect` now operates correctly? If not, I think it's preferable to update that function to handle Windows appropriately.

tensilelite/Tensile/Common/GlobalParameters.py
bstefanuk: This is a step in the right direction; however, I think as a part of this effort we should look to avoid adding logic the `assignGlobalParameters` that relies on private underscore function, e.g., `_windowsSearchPaths()`. The preference would be to instead extend the utility of the current logic in `Tensile.Toolchain.Validators` so that we can isolate the dependence on smi/ld utilities to where they're needed in Tensile/TensileCreateLibrary.

tensilelite/Tensile/KernelWriterActivationFunction.py
```diff
@@ -136,6 +136,8 @@ def getHeaderFileString(self):
     self._tf.setKernelInfo(isa, self.state["Kernel"]["WavefrontSize"])
 
     activationCDataType = self.state["ProblemType"]["ActivationComputeDataType"]
+    if not self._tf.isInit():
+      self._tf.init(tuple(self.state["Kernel"]["ISA"]), globalParameters["AssemblerPath"])
```
bstefanuk: Is "AssemblerPath" still being used? If it is still in the 6.5 release staging branch, I expect to see some pretty nasty merge conflicts when we try and integrate this into develop.

tensilelite/Tensile/Ops/AMaxGenerator.py
```diff
@@ -31,6 +31,15 @@
...
 from contextlib import contextmanager
+
+try:
+    from Tensile import Tensile
+except ImportError:
+    import os.path
+    import sys
+    parentdir = os.path.normpath(os.path.join(os.path.dirname(os.path.realpath(__file__)), "..", ".."))
+    sys.path.append(parentdir)
+
```
bstefanuk: These previously weren't required. I'm curious, what has changed?

tensilelite/Tensile/Toolchain/Validators.py
amd-mtrifuno: Updated version of isRhel8() function works correctly, in the case of Windows always returns False. I'm back to using osSelect.

tensilelite/Tensile/KernelWriterActivationFunction.py
amd-mtrifuno: No, "AssemblerPath" is no longer used. This code is result of automatic merging. I removed it.

tensilelite/Tensile/Ops/AMaxGenerator.py
amd-mtrifuno: I added support for the ExtOp libraries on Windows. PR on the develop branch reported error when importing the TensileInstruction in AMaxGenerator.py, LayerNormGenerator and SoftmaxGenerator.py files. However, I have tested without changes on the rocm-rel-6.5 branch and build is successful without this part of the code, I reverted changes.

stellaraccident:
Hi folks - David and I discussed a strategy for getting Windows support merged because we have two patch sets that are somewhat disjoint, and they are both in the "make it work" variety and can't be landed like this.

Here are the branches we have:

* https://github.com/ROCm/hipBLASLt/pull/1864 (by an AMD employee)
* https://github.com/ROCm/TheRock/pull/441 (by an outside contributor)

We also have another effort underway which should produce results in the next week, where we are landing a much simplified/cleaned up CMake build.

I'm in the loop on all of these, and I have a pretty good understanding of all of the changes and reasons for making them. I'm also developing on Windows and am going to help the hipBLASLt team get these patches staged and landed in a more organized fashion over the next week. As such, we are going to pause further review of 1864 and 441, using them more as reference material to assemble changes to land. Since I'm working on a Windows development environment, I'll be able to verify these more effectively than the regular hipBLASLt team. I will be prioritizing/ordering by: C++ changes, Python changes, and then CMake changes. My hope is that by holding off on the CMake changes, we can just switch to the new build files for Windows and have a better time of it all around.

I'll be staging changes on the branch https://github.com/ROCm/hipBLASLt/tree/slaurenz_therock_win_stage which I will keep rebased to develop as we land things.

Thanks for the time finding the "bottom" of what it will take to support Windows -- these exercises of cutting through in one swath are really useful to get the whole scope known.

jayhawk-commits:
Not migrating this PR to the monorepo due to the target branch.

PR #1876, opened 2025-04-03
Remove float6 float4 headers

bnemanich:
The hipblaslt_float4.h and hipblaslt_float6.h were causing some issues with XLA builds. These can be removed since there is currently not support for GEMMs that output fp4 or fp6 data.


jichangjichang:
Can we just use a custom complier flag to disable it for now?

bnemanich:
> Can we just use a custom complier flag to disable it for now?

I'm not positive, since I'm unable to replicate their issues locally. But this code that I'm removing is not being used anywhere anyways, meaning it is untested. I would prefer to remove the untested code and add it back in when we are ready to use it. Ideally these would be using datatypes from the hip headers, but those aren't available yet.

jichangjichang (approved):

amd-chunxlin (approved):

msujon-AMD (approved):

PR #1896, opened 2025-04-09
bad_grid I8II, NN, NT, TN, TT, gfx942 cu:80

amd-mgeorgie:
grid solution libraries added for 
I8II, NN, NT, TN, TT
gfx942 cu:80


amd-mgeorgie (commented):
just a comment symbol to restart the checks of the azure

PR #1909, opened 2025-04-10
Deterministic naming

davidd-amd:
This PR makes changes to the naming functionality in tensilelite such that naming function queries always return the same name regardless of the order that solutions/kernels are processed and/or the number of targets built. Previously, naming functions relied on a dictionary typically named `kernelNameMin` or `solutionNameMin` which was previously mutable. As a result, the number solutions processed and the order of processing would result in a different name. 

With these changes, for a given kernel/solution state, the computed names will always be the same. As a result, we can update the logic files in a subsequent PR to include the deterministic name. Computing a kernel name will not be required at runtime and we no longer need all solutions in memory at once to compute a consistent name. This will aide in providing updates to the build system to expose more parallelism and establish a means to extend logic file validation. It also enables reproduceable builds because the kernel names are the symbols in the co files and should be fixed from this point forward.

Some naming functions were replaced to provide more clarity in how they are intended to be used. The new naming functions and the previous definition include:

- `getKernelNameMin` -> `getNameMin(kernel, kernelMinNaming, splitGSU, True)`
- `getSolutionNameMin` ->  `getNameMin(kernel, kernelMinNaming, splitGSU, False)`
- `getSolutionNameFull` ->  `getNameMin(kernel, list(validParameters.keys()), splitGSU, False)`

Here is an example of how the names differ when building for just the 90a vs building for all

  
```
Same kernel but building for 90a vs all
90a
Cijk_Ailk_Bjlk_BBS_BH_Bias_HAS_SAV_UserArgs_MT112x128x32_MI16x16x1_SN_LDSB0_AFC1_AFEM1_CLR1_EPS0_GRVWA2_GRVWB2_GSUAMB_K1_LBSPPA896_LBSPPB1024_LPA16_LPB16_LPM0_LRVW4_MIAV0_MIWT7_2_MO40_NTn1_NTA0_NTB0_NTC0_NTD0_NTM0_NEPBS2_NLCA7_NLCB1_PGR2_PLR1_SIA3_SS1_SPO1_SRVW0_SSO0_SVW1_TLDS0_USFGROn1_VSn1_VWA1_VWB2_WSGRA0_WSGRB0_WG16_16_1

all
Cijk_Ailk_Bjlk_BBS_BH_Bias_HAS_SAV_UserArgs_MT112x128x32_MI16x16x1_SN_LDSB0_AFC1_AFEM1_AFEM1_ASEM1_CLR1_CADS0_DTVA0_DTVB0_EPS0_FDSI0_GRPM1_GRVWA2_GRVWB2_GSUAMB_GLS0_ISA90a_IU1_K1_LBSPPA896_LBSPPB1024_LBSPPM0_LPA16_LPB16_LPM0_LRVW4_LWPMn1_MIAV0_MIWT7_2_MO40_NTn1_NTA0_NTB0_NTC0_NTD0_NTM0_NEPBS2_NLCA7_NLCB1_ONLL1_PGR2_PLR1_PKA1_SIA3_SS1_SPO1_SRVW0_SSO0_SVW1_SK0_SKXCCM0_TLDS0_ULSGRO0_USL1_UIOFGRO0_USFGROn1_VSn1_VWA1_VWB2_WSGRA0_WSGRB0_WS64_WG16_16_1
```

With the new naming scheme, we compared all of the co files for the 90a to ensure that the same number of kernels appear in each co file and that the symbol names were the same when compared to an `all` build. There were no differences in the kernel count or symbol names.

### DIFF
--- a/tensilelite/Tensile/BenchmarkProblems.py
+++ b/tensilelite/Tensile/BenchmarkProblems.py
@@ -37,8 +37,10 @@ from Tensile.KernelWriter import DebugConfig
 from Tensile.Toolchain.Component import Assembler
 from Tensile.SolutionStructs.Problem import ProblemType, ProblemSizes
 from Tensile.SolutionStructs.Solution import Solution
-from Tensile.SolutionStructs.Validators.MatrixInstruction import matrixInstructionToMIParameters, validateMIParameters
-from Tensile.SolutionStructs.Naming import getMinNaming, getNameMin, getSerialNaming, getNameFull, getKeyNoInternalArgs
+from Tensile.SolutionStructs.Validators.MatrixInstruction import matrixInstructionToMIParameters, \
+                                                                 validateMIParameters
+from Tensile.SolutionStructs.Naming import getSerialNaming, getKeyNoInternalArgs, getSolutionNameMin, \
+                                           getKernelNameMin
 
 from .BenchmarkStructs import BenchmarkProcess, constructForkPermutations
 from .Contractions import ProblemType as ContractionsProblemType
@@ -53,8 +55,6 @@ from Tensile.Common import HR, print1, print2, IsaInfo, IsaVersion, \
         BENCHMARK_PROBLEMS_DIR, BENCHMARK_DATA_DIR, DepthUConfig
 from Tensile.Common.Architectures import isaToGfx, gfxToVariants
 from Tensile.Common.GlobalParameters import globalParameters, startTime
-from Tensile.Common.RequiredParameters import getRequiredParametersMin
-
 
 
 def _generateForkedSolutions(problemType, constantParams, forkPermutations, assembler: Assembler, \
@@ -237,9 +237,7 @@ def writeBenchmarkFiles(
                 kernelHelperNames.add(kname)
 
     kernelSerialNaming = getSerialNaming(kernels)
-    kernelMinNaming = getMinNaming(kernels)
     kernelWriterAssembly = KernelWriterAssembly(
-                               kernelMinNaming,
                                kernelSerialNaming,
                                asmToolchain.assembler,
                                debugConfig,
@@ -260,17 +258,15 @@ def writeBenchmarkFiles(
                             debugConfig.splitGSU,
                             cmdLineArchs,
                             kernelSerialNaming,
-                            kernelMinNaming,
                             errorTolerant=True,
                             generateSourcesAndExit=globalParameters["GenerateSourcesAndExit"], # put in debug config
                             compress=False,
                             useShortNames=useShortNames
                         )
     # ^ this is where solutions is mutated
-    solutionMinNaming = getRequiredParametersMin()
     for s in solutions:
-        s["SolutionNameMin"] = getNameMin(solution, solutionMinNaming, debugConfig.splitGSU)
-        s["KernelNameMin"]   = getNameMin(solution, solutionMinNaming, debugConfig.splitGSU, True)
+        s["SolutionNameMin"] = getSolutionNameMin(solution, debugConfig.splitGSU)
+        s["KernelNameMin"]   = getKernelNameMin(solution, debugConfig.splitGSU)
 
     newLibraryDir = ensurePath(sourcePath / 'library')
     newLibraryFile = os.path.join(newLibraryDir, "TensileLibrary")
@@ -283,7 +279,7 @@ def writeBenchmarkFiles(
                      depthUConfig,
                      isaInfoMap,
                  )
-    newLibrary.applyNaming(debugConfig.splitGSU, kernelMinNaming)
+    newLibrary.applyNaming(debugConfig.splitGSU)
     LibraryIO.write(newLibraryFile, state(newLibrary), globalParameters["LibraryFormat"])
 
     codeObjectFiles = [os.path.relpath(f, sourcePath) \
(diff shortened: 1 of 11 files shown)


tensilelite/Tensile/BenchmarkProblems.py
```diff
@@ -430,7 +431,7 @@ def _benchmarkProblemType(problemTypeConfig, problemSizeGroupConfig, problemSize
                 printExit(msg)
 
             for solution in solutions:
-                print2("#    ({}:{}) {}".format(0, 0, getNameFull(solution, debugConfig.splitGSU)))
+                print2("#    ({}:{}) {}".format(0, 0, getSolutionNameMin(solution, debugConfig.splitGSU)))
```
davidd-amd, line 434:
```suggestion
                print2("#    ({}:{}) {}".format(0, 0, getSolutionNameFull(solution, debugConfig.splitGSU)))
```

tensilelite/Tensile/BenchmarkProblems.py
```diff
@@ -457,12 +458,11 @@ def _benchmarkProblemType(problemTypeConfig, problemSizeGroupConfig, problemSize
...
 
             # add SolutionIndex and SolutionNameMin into benchmark yaml
-            solutionMinNaming = getMinNaming(solutions)
             for i in range(0, len(solutions)):
                 solution = solutions[i]
                 solution["SolutionIndex"] = i
-                solution["SolutionNameMin"] = getNameMin(solution, solutionMinNaming, debugConfig.splitGSU)
-                solution["KernelNameMin"]   = getNameMin(solution, solutionMinNaming, debugConfig.splitGSU, True)
+                solution["SolutionNameMin"] = getSolutionNameMin(solution, debugConfig.splitGSU)
+                solution["KernelNameMin"]   = getKernelNameMin(solution, debugConfig.splitGSU)
```
bstefanuk, line 465: We compute the names here, but we also compute the names on L268, L269 just after `writeSolutionsAndKernels`. Since we are already updating the names in place inside `writeBenchmarkFiles`, why are we computing the names again here?

tensilelite/Tensile/SolutionStructs/Naming.py
```diff
@@ -156,63 +94,55 @@ def getParameterValueAbbreviation(key, value):
     raise Exception(f"Parameter {key}={value} is new object type ({type(value)})")
 
 
-def getNameMin(state, requiredParameters, splitGSU: bool, ignoreInternalArgs = False):
+def _getName(state, requiredParameters: frozenset, splitGSU: bool, ignoreInternalArgs):
```
bstefanuk, line 97:
```suggestion
def _getName(state, requiredParameters: frozenset, splitGSU: bool, ignoreInternalArgs: bool) -> str:
```

tensilelite/Tensile/SolutionStructs/Naming.py
```diff
@@ -269,16 +199,24 @@ def shortenFileBase(kernelMinNaming, splitGSU, kernel):
...
   else:
-    fileBase = shortenFileBase(kernelMinNaming, splitGSU, kernel)
+    fileBase = shortenFileBase(splitGSU, kernel)
   return fileBase
 
 
-def getKernelName(kernelMinNaming, splitGSU, kernel):
-  kernelName = getNameMin(kernel, kernelMinNaming, splitGSU, True)
-  return kernelName
+def getKernelNameMin(kernel, splitGSU: bool):
```
bstefanuk, line 212: I understand this was updated to match the field in the solutions, however, there isn't a notion of KernelNameFull. Does it make sense to call this KernelNameMin, or is it just the kernel name?

tensilelite/Tensile/SolutionStructs/Naming.py
```diff
@@ -156,63 +94,55 @@ def getParameterValueAbbreviation(key, value):
...
+  else:
+    requiredParametersTemp = requiredParametersTemp.union(["WorkGroupMapping",
+                                                           "WorkGroupMappingXCC",
+                                                           "WorkGroupMappingXCCGroup",
+                                                           "StaggerU",
+                                                           "StaggerUStride",
+                                                           "StaggerUMapping",
+                                                           "GlobalSplitUCoalesced",
+                                                           "GlobalSplitUWorkGroupMappingRoundRobin"])
+  components = [f'{str(state["ProblemType"])}']
```
bstefanuk, line 125: Do you have details of how the duplicate counts differ (if at all) before and after this update?

tensilelite/Tensile/LibraryLogic.py
```diff
@@ -293,14 +291,11 @@ def __init__(self, problemType, problemSizesList, solutionsList, \
...
     self.numSolutions = len(self.solutions)
-    self.solutionMinNaming = getMinNaming(self.solutions)
     self.solutionNames = []
     self.solutionTiles = []
     for solution in self.solutions:
-      self.solutionNames.append(getNameMin(solution, \
-          self.solutionMinNaming, self.splitGSU))
-      self.solutionTiles.append("%ux%u"%(solution["MacroTile0"], \
-          solution["MacroTile1"]))
+      self.solutionNames.append(getSolutionNameMin(solution, self.self.splitGSU))
```
davidd-amd:
```suggestion
      self.solutionNames.append(getSolutionNameMin(solution, self.splitGSU))
```

KKyang (approved):

tensilelite/Tensile/SolutionStructs/Naming.py
davidd-amd: They do not differ.

tensilelite/Tensile/SolutionStructs/Naming.py
davidd-amd: I like having parity with solution naming functions. I could add getKernelNameFull but it would be used anywhere.

tensilelite/Tensile/BenchmarkProblems.py
davidd-amd: I need to look at this - thanks.

bstefanuk (approved):

tensilelite/Tensile/SolutionStructs/Naming.py
davidd-amd: Will resolve in follow on PR

PR #1917, opened 2025-04-11
Fix problem name ambiguity and improve performance of ProblemType __str__

davidd-amd:
When input type is the same as output type, the type portion of the kernel name gets condensed. For example, B8NBH would be equivalent to B8NB8NS_BH if it were not shortened. The abbreviated names are particularly confusing to read since the "BH" part of the name is no longer separated by an underscore. These letters indicate the kernel supports beta and is high-precision-accumulate. Without the underscore, they are easy to confuse as the output type being BF16 and compute type being F16 (which is not supported).

The names should be updated to at least include the underscore (B8N_BH would be clearer). Additionally, for 8-bit types where there are many possible type combinations, a full name for all cases might be the better option for ease of use.

***This will change the name of `ProblemTypes` and as a consequence the names that go into the .dat files and the symbols in the binaries.***

### DIFF
--- a/tensilelite/Tensile/SolutionStructs/Problem.py
+++ b/tensilelite/Tensile/SolutionStructs/Problem.py
@@ -966,105 +966,98 @@ class ProblemType(Mapping):
   def __str__(self):
     indexChars = INDEX_CHARS
     # C dimensions
-    name = "C"
-    for i in range(0, self["NumIndicesC"]):
-      name += indexChars[i].lower()
+    name = ["C" + "".join(indexChars[i].lower() for i in range(0, self["NumIndicesC"]))]
+      
     # A dimensions
-    name += "_A"
-    for i in self["IndexAssignmentsA"]:
-      name += indexChars[i] if i in self["MirrorDimsA"] else indexChars[i].lower()
+    name.append("A" + "".join(indexChars[i] if i in self["MirrorDimsA"] else indexChars[i].lower() for i in self["IndexAssignmentsA"]))
     if self["ComplexConjugateA"]:
-      name += "C"
+      name.append("C")
+    
     # B dimensions
-    name += "_B"
-    for i in self["IndexAssignmentsB"]:
-      name += indexChars[i] if i in self["MirrorDimsB"] else indexChars[i].lower()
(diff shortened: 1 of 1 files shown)


tensilelite/Tensile/SolutionStructs/Problem.py
```diff
@@ -966,105 +966,98 @@ def assignDerivedParameters(state, printIndexAssignmentInfo: bool=False):
...
 
     # Other
-    if self["UseBeta"]: name += "B"
-    if self["HighPrecisionAccumulate"] and not self["SilentHighPrecisionAccumulate"]: name += "H"
-    if self["UseInitialStridesAB"]: name += "I"
-    if self["UseInitialStridesCD"]: name += "Ic"
+    if self["UseBeta"]: name.append("B")
+    if self["HighPrecisionAccumulate"] and not self["SilentHighPrecisionAccumulate"]: name[-1]+= ("H")
+    if self["UseInitialStridesAB"]: name[-1] += "I"
+    if self["UseInitialStridesCD"]: name[-1] += "Ic"
```
AlexBrownAMD: This part looks like it'd have an inconsistency in name segments. Beta uses append, so there'd be "_B", but the HPA's "H" is attached to the last element, which could be the "_B", or if no beta then whatever comes before that. Should this be something like this?

Other = ""
if ... Other += "B"
if ... Other += "H"
if Other is not empty name.append(Other)

davidd-amd: Good catch

tensilelite/Tensile/SolutionStructs/Problem.py
```diff
@@ -966,105 +966,101 @@ def assignDerivedParameters(state, printIndexAssignmentInfo: bool=False):
...
+      name.append("GG")
     else:
-      name += "" if self["StridedBatched"] else "_GB" # legacy
+      if not self["StridedBatched"]:
+        name.append("GB") # legacy
 
     # Activation Naming
     if self["ActivationType"] != 'none':
       if self["ActivationType"] == 'all':
         name += "_A"
```
KKyang: Should be ``name.append("A")``?

davidd-amd: Thanks @KKyang !

AlexBrownAMD (approved):

KKyang (approved):

PR #1930, opened 2025-04-14
Disable GSU code when GSO=0

AlexBrownAMD:
GSU=0 should disable all GSU code. This change updates some sections of code that were still generating GSU-related code when GSU was disabled.


hliuca (dismissed):
Looks good. Thank you for the fix.

hliuca:
I tried BiasSrc: A and B. Everything is fine.

jichangjichang:
The purpose of GSU=0 still keep GSU codes is we want to reduce the number of kernels.  @KKyang 
Did this impact anything if without this change?

tensilelite/Tensile/SolutionStructs/Solution.py
```diff
@@ -1115,6 +1115,7 @@ def assignDerivedParameters(
         if state["LocalSplitU"] > 1:
           reject(state, printRejectionReason, "Atomic Stream-K not working with LocalSplitU")
       if not state["Valid"]:
+        print2("in assignDerivedParameters, state['Valid'] = False")
```
KKyang, line 979: Please add the following assert such as
if StreamK == False, GSU must > 0
if StreamK == True, GSU must == 0

jichangjichang (commented):
Or If GSU=0 is only used for StreamK case, please reject solution for GSU=0 + StreamK=False.

AlexBrownAMD:
> The purpose of GSU=0 still keep GSU codes is we want to reduce the number of kernels. @KKyang Did this impact anything if without this change?

Yes, we need the ability to disable GSU code since it is not compatible with StreamK code. Without this change, stream-k kernels with epilogues fail to generate due to some extra GSU code being added. If GSU=0, there should not be any GSU code present.

tensilelite/Tensile/SolutionStructs/Solution.py
AlexBrownAMD: Fixed

AlexBrownAMD:
> Or If GSU=0 is only used for StreamK case, please reject solution for GSU=0 + StreamK=False.

Added a reject condition to require either GSU or StreamK to be enabled

AlexBrownAMD:
Merged latest changes from develop

KKyang (approved):

hliuca (approved):

PR #1931, opened 2025-04-14
Do not use GTEST_BOTH_LIBRARIES variable.

### LINKED ISSUE #414: [hipBLASLt] Do not use GTEST_BOTH_LIBRARIES variable.
Upstream: https://github.com/ROCm/hipBLASLt/pull/1931

stellaraccident:
* This is a legacy variable that is not recommended for modern use and is only available via the GTest find module, not the actual config provided by the package.
* Recommended usage is just to depend on the libraries.
* Also removes GTEST_INCLUDE_DIRS, since that is implicit in depending on the library.

Fixes https://github.com/ROCm/TheRock/issues/414


stellaraccident:
No longer needed with next-cmake.

PR #1934, opened 2025-04-14
Hide static library symbols

bnemanich:
Hide the symbols coming from LLVM. This has been shown to allow PyTorch builds to succeed.


lawruble13 (approved):

awhittle3 (approved):

KKyang (approved):

jichangjichang (approved):

jichangjichang:
most of platform/arch are verified and only gfx942 is still running, which should be enough since this change is not for specific platform. I would like to merge it to unblock mainline build.

PR #1951, opened 2025-04-18
Port rocisa to Windows.

stellaraccident:
* Adds some missing headers that the Windows STL is stricter about.
* Separates `run` and `demangle` from `helper.hpp` into a new translation unit and conditionally provides Windows or POSIX-ey versions. Adds some RAII safety and error handling above and beyond the contributed original Windows implementation.
* Works around an interaction between nanobind and the Windows ABI not being able to differentiate between generated member function trampolines that differ only by const-ness.

### DIFF
--- a/tensilelite/rocisa/rocisa/include/helper.hpp
+++ b/tensilelite/rocisa/rocisa/include/helper.hpp
@@ -21,16 +21,17 @@
  *
  * ************************************************************************ */
 #pragma once
+#include <algorithm>
+#include <array>
 #include <stdexcept>
 #include <string>
-#include <sys/wait.h>
-#include <unistd.h>
 #include <vector>
-#ifdef __GNUG__
-#include <cxxabi.h>
-#endif
 
 using IsaVersion = std::array<int, 3>;
+
+std::pair<int, std::string>
+    run(const std::vector<char*>& cmd, const std::string& input, bool debug = false);
+std::string demangle(const char* name);
 
 inline std::string getGfxNameTuple(const IsaVersion& isaVersion)
 {
@@ -47,52 +48,6 @@ inline std::string getGfxNameTuple(const IsaVersion& isaVersion)
            + int_to_hex[isaVersion[2]];
 }
 
-inline std::pair<int, std::string>
-    run(const std::vector<char*>& cmd, const std::string& input, bool debug = false)
-{
-    int   p[2];
-    pid_t pid;
-    if(pipe(p) == -1)
-    {
-        throw std::runtime_error("cmd failed!");
-    }
-
-    pid = fork();
-    if(pid == 0)
-    {
-        close(p[1]);
-        dup2(p[0], STDIN_FILENO);
-        if(!debug)
-        {
-            dup2(p[0], STDERR_FILENO);
-            dup2(p[0], STDOUT_FILENO);
-        }
-        close(p[0]);
-        execvp(cmd[0], cmd.data());
-        perror("execvp");
-        exit(1);
-    }
-    else
-    {
-        close(p[0]);
-        write(p[1], input.c_str(), input.size());
-        close(p[1]);
-        char        buf[128] = {0};
-        std::string result;
-        read(p[0], buf, 128);
-        result += buf;
-        int rcode;
-        waitpid(pid, &rcode, 0);
-        if(WIFEXITED(rcode))
-        {
-            rcode = WEXITSTATUS(rcode);
-        }
-        return {rcode, result};
-    }
-    // Should not go here.
-    return {0, "0"};
-}
-
 template <typename T>
 bool checkInList(const T& a, const std::vector<T> b)
 {
@@ -103,18 +58,4 @@ template <typename T>
 bool checkNotInList(const T& a, const std::vector<T> b)
 {
     return std::find(b.begin(), b.end(), a) == b.end();
-}
-
-inline std::string demangle(const char* name)
-{
-    std::string result = name;
-#ifdef __GNUG__
-    int   status    = -1;
-    char* demangled = abi::__cxa_demangle(name, nullptr, nullptr, &status);
-    result          = (status == 0) ? demangled : name;
-    free(demangled);
-#else
-#error "Windows not supported"
-#endif
-    return result;
 }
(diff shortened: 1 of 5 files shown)


tensilelite/rocisa/rocisa/src/helper.cpp
davidd-amd, line 1: If I recall correctly, these run methods are used to invoke the compiler. I am pretty sure we could use the comgr c API directly to do this. I am a little rusty on linux/unix system programming but the new windows code looks less likely to create zombie processes. Not sure about the posix impl (I understand that was pre-existing).

tensilelite/rocisa/rocisa/include/helper.hpp
davidd-amd, line 1: Were transitively including the algo header because the linux build failed:

```
hipblaslt/tensilelite/rocisa/rocisa/include/helper.hpp:53:17: error: no member named 'find' in namespace 'std'
```

davidd-amd, line 1: I think amdclang++ wants us to omit `= false` in the function signature in `helper.cpp`:

```
hipblaslt/tensilelite/rocisa/rocisa/src/helper.cpp:199:71: error: redefinition of default argument
  199 |     run(const std::vector<char*>& cmd, const std::string& input, bool debug = false)
      |                                                                       ^       ~~~~~
hipblaslt/tensilelite/rocisa/rocisa/include/helper.hpp:32:71: note: previous definition is here
   32 |     run(const std::vector<char*>& cmd, const std::string& input, bool debug = false);
```

tensilelite/rocisa/rocisa/src/code.cpp
```diff
@@ -166,7 +171,12 @@ void init_code(nb::module_ m)
         .def("count", &rocisa::Module::count)
         .def("getItem", &rocisa::Module::getItem)
         .def("setItem", &rocisa::Module::setItem)
-        .def("setItems", &rocisa::Module::setItems)
+        // nanobind cannot reliably bind the same method signatures that differ
+        // only by const-ness, so indirectly bind via a lambda.
+        .def("setItems", [](rocisa::Module& self,
```
KKyang: Isn't this method signature unique?

tensilelite/rocisa/rocisa/src/code.cpp
```diff
@@ -152,7 +152,12 @@ void init_code(nb::module_ m)
         .def("setInlineAsmPrintMode", &rocisa::Module::setInlineAsmPrintMode)
         .def("addSpaceLine", &rocisa::Module::addSpaceLine)
         .def("add", &rocisa::Module::add, nb::arg("item"), nb::arg("pos") = -1)
-        .def("addItems", &rocisa::Module::addItems)
+        // nanobind cannot reliably bind the same method signatures that differ
```
KKyang, line 155: Isn't this method signature unique?

KKyang (dismissed):

tensilelite/rocisa/rocisa/src/code.cpp
stellaraccident: Not in the Microsoft abi. It's a rather esoteric corner case interaction between the template metaprogramming nanobind is doing and some constraints on the Microsoft side. My suggestion here would be to just accept it and move on. The full explanation requires language lawyering c++ and some of the higher order method pointer strength reduction that nanobind does...

tensilelite/rocisa/rocisa/src/helper.cpp
stellaraccident: Yeah, I would add the posix implementation to the technical debt pile and rewrite it. I didn't look at it super closely but it has some anti patterns in it that are obvious and likely to leak.

tensilelite/rocisa/rocisa/src/code.cpp
KKyang: Can you add a comment that this is a Windows only issue?

tensilelite/rocisa/rocisa/src/helper.cpp
KKyang: I can fix it using comgr C API after this commit is merged.

tensilelite/rocisa/rocisa/include/helper.hpp
stellaraccident: Oh right - copypasta and I left testing the Linux side to the ci

tensilelite/rocisa/rocisa/src/helper.cpp
stellaraccident: That would be better most likely. These are not the most robust ways to be invoking the compiler.

stellaraccident (commented):
PTAL

tensilelite/rocisa/rocisa/include/label.hpp
```diff
@@ -21,8 +21,10 @@
  *
  * ************************************************************************ */
 #pragma once
+#include <algorithm>
```
davidd-amd, line 24: This didn't resolve the error - should this be included in helper.hpp?

stellaraccident: Ok, added to helper.hpp and giving the CI a spin (seems to just be a Linux thing so I'm not catching locally on my windows machine).

tensilelite/rocisa/rocisa/src/code.cpp
```diff
@@ -166,7 +172,13 @@ void init_code(nb::module_ m)
         .def("count", &rocisa::Module::count)
         .def("getItem", &rocisa::Module::getItem)
         .def("setItem", &rocisa::Module::setItem)
-        .def("setItems", &rocisa::Module::setItems)
+        // nanobind cannot reliably bind the same method signatures that differ
```
KKyang, line 175: I ran clang-format in my PR and found this file was modified. Can you run clang-format?

KKyang: We also have a pre-commit file in the hooks folder in hipblaslt.

stellaraccident: done.

davidd-amd (approved):
Many thanks!

KKyang (approved):

PR #1952, opened 2025-04-18
[rocisa] Use comgr instead of calling amdclang++

KKyang:
Use comgr C API instead of calling amdclang++ directly.


hcman2 (dismissed):
ok if ci pass

tensilelite/rocisa/CMakeLists.txt
```diff
@@ -26,6 +26,17 @@ set(CMAKE_CXX_STANDARD 20)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_CXX_EXTENSIONS OFF)
 
+# Make sure the folder is the same as the one CXX_COMPILER is in if ROCM_PATH is not set
+if(NOT DEFINED ROCM_PATH)
```
stellaraccident: This should not be needed and we can't be doing this kind of path manipulation assuming a fixed rocm_path.

Still reading the rest to figure out why this is needed.

tensilelite/rocisa/rocisa/include/helper.hpp
```diff
@@ -21,15 +21,17 @@
...
 #include <string>
-#include <sys/wait.h>
 #include <unistd.h>
 #include <vector>
 #ifdef __GNUG__
 #include <cxxabi.h>
 #endif
 
+#include <amd_comgr/amd_comgr.h>
+
```
stellaraccident, line 29: Please land this after my change #1951 and then don't land any more non trivial definitions in the header: put them in helper.cpp

tensilelite/rocisa/rocisa/include/helper.hpp
```diff
@@ -47,50 +49,87 @@ inline std::string getGfxNameTuple(const IsaVersion& isaVersion)
            + int_to_hex[isaVersion[2]];
 }
 
-inline std::pair<int, std::string>
-    run(const std::vector<char*>& cmd, const std::string& input, bool debug = false)
+#define CHECK_COMGR(call)                                                                  \
```
stellaraccident: Tip; you should almost never need a macro like this defined in a header for a local use case. This is a sure sign that you should be using a real translation unit.

stellaraccident: Further nit: since this is just raising an exception on failure, it is perfect for an inline static function, not a macro. Only use big multi line macros like this if doing something structural (ie. Return on error, etc). If just wanting to capture a preprocessor value like the current location, have a one line check macro that calls the check function, with most of the logic in a real function.

tensilelite/rocisa/rocisa/include/helper.hpp
```diff
@@ -47,50 +49,87 @@ inline std::string getGfxNameTuple(const IsaVersion& isaVersion)
...
+        {                                                                                  \
+            const char* status_string = nullptr;                                           \
+            amd_comgr_status_string(status, &status_string);                               \
+            throw std::runtime_error(std::string("COMGR call failed: ") + #call            \
+                                     + " (COMGR status: " + std::to_string(status) + ", "  \
+                                     + (status_string ? status_string : "Unknown") + ")"); \
+        }                                                                                  \
+    } while(0)
+
+inline std::pair<int, std::string> run(const std::string& input,
```
stellaraccident: Tip: if ever marking big functions like this inline just so you can define them in a header vs impl file, that is a sign they are misplaced.

stellaraccident: As part of this patch, please rename this function to be indicative of what it does: it runs the compiler not just "run".

tensilelite/rocisa/rocisa/include/helper.hpp
```diff
@@ -47,50 +49,87 @@ inline std::string getGfxNameTuple(const IsaVersion& isaVersion)
...
-        exit(1);
+        CHECK_COMGR(amd_comgr_set_data(data, input.size(), input.c_str()));
     }
-    else
+    CHECK_COMGR(amd_comgr_set_data_name(data, "input.s"));
+
+    // Create data set and add the input data
+    CHECK_COMGR(amd_comgr_create_data_set(&dataSet));
+    CHECK_COMGR(amd_comgr_create_data_set(&outputDataSet));
+    CHECK_COMGR(amd_comgr_data_set_add(dataSet, data));
```
stellaraccident: What happens if any of these get an error and throw? As I read it, every one of them leaks memory. 

You need to rewrite this so that your allocated entities are contained by a raii instance that will deallocate any allocated things on destruction. See #1951 for one (of many) ways you can do this to bridge to a c based API safely.

stellaraccident (changes requested):
Let's land #1951 first since that introduces a structural improvement.

This needs more work to not leak and be exception safe and should be implemented in the helper.cpp introduced in the above. 

Also, please don't land those cmake changes to rocm_path. I couldn't follow why they were needed so we can discuss more what the intent is.

tensilelite/rocisa/CMakeLists.txt
KKyang: I was told that we have to use env ROCM_PATH to get the root folder. Is there a new way to get the root folder?

tensilelite/rocisa/rocisa/include/helper.hpp
KKyang: I'll put this into cpp

tensilelite/rocisa/rocisa/include/helper.hpp
KKyang: I'll change to unique_ptr instead

tensilelite/rocisa/CMakeLists.txt
stellaraccident: There is no root folder. I'm curious why this patch needs it.

tensilelite/rocisa/rocisa/include/helper.hpp
stellaraccident: I'm not sure that is a good way: these have explicit apis for create/release and do not have the same semantics as delete with respect to nullptr afaict. But I'll have a look at what you come up with.

tensilelite/rocisa/rocisa/include/helper.hpp
stellaraccident: Since this patch needs more review/iteration and is a functional change, please stack it onto #1951 which is NFC and makes the structural change. 

If this one was ready to go, I would have approved it and adapted mine, but we shouldn't have two independent patches changing the same structure. 

Generally, NFC/structure patches, which have a low risk of regression should precede functional changes that can introduce regressions and further bugs.

KKyang:
Merge after #1951 is merged

tensilelite/rocisa/CMakeLists.txt
KKyang: the root of the rocm folder.

KKyang: I guess we have to go the CMAKE_PREFIX_PATH, but I'm not sure if the CI team is aware of this.

tensilelite/rocisa/rocisa/include/helper.hpp
KKyang: Now the PR is on top of #1951

tensilelite/rocisa/CMakeLists.txt
stellaraccident: Let me look more closely at it tomorrow. We are already locating hip and should find this the same way (they are distributed together).

The other part that I wasn't sure if you were trying to do is to make sure that the c++ compiler, which is still used for much of the project, is from the same distribution as comgr. It would be gross if those two got out of alignment, but there isn't a good was to ensure that.

davidd-amd: > Let me look more closely at it tomorrow. We are already locating hip and should find this the same way (they are distributed together).
> 
> The other part that I wasn't sure if you were trying to do is to make sure that the c++ compiler, which is still used for much of the project, is from the same distribution as comgr. It would be gross if those two got out of alignment, but there isn't a good was to ensure that.

That is a really good point. Let me ask around about this.

stellaraccident:
I tested this on Windows, and indeed, we are going to need more work there. The issue is that you can't just have a cross-filesystem dep on a DLL like this (which is what a Python extension is) without making arrangements for how to find it at runtime (i.e. there is no RPATH on Windows).

```
ImportError: DLL load failed while importing rocisa: The specified module could not be found.
```

The Python frameworks solve this in one of two ways:

1. Bundle ROCm as a sibling to the Python binary distribution in a way that it can be found. PyTorch does this in the current version and it is as invasive as it sounds.
2. Use a Python level pre-loading mechanism to load the necessary DLLs prior to importing the Python extension (which links to DLLs by name). This is what CUDA based dependencies do and what ROCm based PyTorch will do in the next version.

Neither option is particularly satisfying for this kind of situation. In this case, since it is a build-only dep, we would most likely want to emulate the Linux RPATH mechanic to ensure the DLL is loaded properly. However, doing that properly will require a bit more pre-work.

Could we put this change on ice for a little while? I know roughly how to enable it but have other priorities right this minute and would like to come up with a mechanism that will work for any ROCm project vs just a one off. If this becomes urgent, I could do something project specific in a few hours if needed.

stellaraccident (changes requested):
Marking request changes per the above discussion. We can apply this once a bit more work is done on the windows side.

stellaraccident:
Thanks for rebasing. I think I know how to fix this on the windows build, but I won't be able to get to it for some days. Should be an improvement, though -- just need to sequence it with some other work.

jayhawk-commits:

PR #1954, opened 2025-04-18
Physical design change to avoid copying header files

davidd-amd:
There is no reason to copy header files into the build directory when they aren't being modified during the configuration step. This PR moves files previously copied into a directory inconsistent with the source tree. Instead, we nest them in the appropriate directory and install them into that same directory in the install tree. We also update the impacted header files such that consumption of header files via build or install interface is consistent.


stellaraccident (dismissed):
Good / thank you. The improper nesting of headers in some of the ROCm projects has been a perennial source of problems, and then it has a tendency to lead to the kind of workarounds like you are removing here. Always best to have the files organized properly in the source tree and then have target based includes.

stellaraccident (approved):

antiagainst (approved):

jichangjichang (approved):

PR #1957, opened 2025-04-19
Roll-up of minimal changes to TensileLite needed to build the library on Windows.

stellaraccident:
* Sets the compiler binary name (clang++.exe vs amdclang++) in top level CMake file. In the future, I feel that we should just be passing the absolute path the the C++ compiler we know we are using vs doing the guesswork, but this makes things work for now (on Linux amdclang++ is a symlink, but on Windows, we don't materialize convenience symlinks because they are not well supported on all systems).
* Find the linker as `ld.lld.exe` on Windows. In the future, I feel that we should pass the absolute path to this as known to the CMake invocation vs relying on guesswork.
* Blackhole stderr for hipconfig and hipcc probes. On Windows they sometimes spew some additional "information" to stderr which is in no way helpful.
* Let locateExe take a None defaultPath. This ended up not being used in this patch, but it was the start of excising all of the hard-coded paths in Tensile. This function is a very bad thing as it stands now (it ignores the PATH invoked with and looks in shoddy system specific absolute paths, silently preferring them if found). We need to eliminate this mechanism entirely as it is very fragile and silent-error prone.
* Make /dev/null input redirection work for Windows and add some misc escaping.
* Marks Tensile library generation as USES_TERMINAL. Without this, any hangs or deadlocks (or just long execution) gets buffered and is very difficult to diagnose. Since this is the longest running, full machine utilizing, part of the build, it makes sense for it to run in the foreground.
* Removes the `HIP` language from rocisa. This was blowing up badly on Windows based on the presence or lack of some specific symlinks, and it is known to require some special project setup to work on that platform. Doesn't seem to be used so just dropped it.
* Limits multiprocessing concurrency to 61 on Windows. This is a known limitation in how Windows multiprocessing is implemented. If exceeding this, it will actually print a warning saying the limit is 61, then proceed anyway and hang/deadlock if you try to use it... This is what prompted the USES_TERMINAL tag above. Otherwise, it is next to impossible to see the error.

Some of the upstream patches introduced more dependence on system specific install locations, but we will not be landing those changes or adding any more absolute path inference to the hipBLASLt build (and we need to remove what is there as it is error prone).

Includes a couple of fixes from #1864

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -200,7 +200,11 @@ else()
 
     set( Tensile_LOGIC "asm_full" CACHE STRING "Tensile to use which logic?")
     set( Tensile_CODE_OBJECT_VERSION "4" CACHE STRING "Tensile code_object_version")
-    set( Tensile_COMPILER "amdclang++" CACHE STRING "Tensile compiler")
+    if(WIN32)
+      set( Tensile_COMPILER "clang++.exe" CACHE STRING "Tensile compiler")
+    else()
+      set( Tensile_COMPILER "amdclang++" CACHE STRING "Tensile compiler")
+    endif()
     set( Tensile_LIBRARY_FORMAT "msgpack" CACHE STRING "Tensile library format")
     set( Tensile_CPU_THREADS "" CACHE STRING "Number of threads for Tensile parallel build")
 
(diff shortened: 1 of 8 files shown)


KKyang (approved):

powderluv (approved):

davidd-amd (commented):
Looks good.

PR #1958, opened 2025-04-19
enable UseScaleAB for fp8 gemm with gelu aux

chengchingwen:
ticket: https://ontrack-internal.amd.com/browse/SWDEV-528425


jichangjichang (commented):
file name should be revised as well.

jichangjichang (approved):

vin-huang (approved):

PR #1961, opened 2025-04-20
Replace assemble/compilation shell scripts with proper CMake targets.

davidd-amd:
The current hipBLASLt has a few targets that use an ad hoc scripted approach to generating binaries. These scripts force Windows users to apply patches to build hipblaslt as they are not cross platform. This PR provides native CMake equivalents in efforts to move hipblaslt towards supporting Windows builds out of the box.


davidd-amd:
@stellaraccident I was concerned that CMake would be able to cope with unix paths in custom commands when on Windows. However, after some testing it seems to implicitly convert paths. Can you confirm my understanding is correct?

I just need to create a full build of the impacted binaries (extop and matrix transform) and compare them with what is generated on develop to be sure I haven't inadvertently dropped symbols. Once that is done I will mark as ready for review.

library/src/amd_detail/rocblaslt/src/extops/CMakeLists.txt
```diff
@@ -0,0 +1,144 @@
...
+    )
+    add_custom_target(
+        ExtOp_${arch} ALL
+        DEPENDS ${output_path}/extop_${arch}.co
+    )
+endforeach()
+
+
+unset(ops_path)
+unset(output_path)
```
davidd-amd: Fix

library/src/amd_detail/rocblaslt/src/kernels/CMakeLists.txt
```diff
@@ -0,0 +1,45 @@
...
+    COMMAND ${CMAKE_CXX_COMPILER} -x hip ${CMAKE_CURRENT_SOURCE_DIR}/matrix_transform.cpp --offload-arch="${archs}" -c --offload-device-only -Xoffload-linker --build-id=${Tensile_BUILD_ID} -O3 -o ${output_library}
+    OUTPUT ${PROJECT_BINARY_DIR}/Tensile/library/hipblasltTransform.hsaco
+    COMMENT "Compiling matrix transform kernels"
+)
+
+add_custom_target(MatrixTransformKernels ALL
+    DEPENDS ${PROJECT_BINARY_DIR}/Tensile/library/hipblasltTransform.hsaco
+)
+
+add_dependencies(MatrixTransformKernels TENSILE_LIBRARY_TARGET)
```
davidd-amd, line 45:
```suggestion
add_dependencies(MatrixTransformKernels TENSILE_LIBRARY_TARGET)

unset(output_library)
 
```

tensilelite/Tensile/Ops/LayerNormGenerator.py
```diff
@@ -23,6 +23,8 @@
 from rocisa.code import Label, Module, RegSet, TextBlock, ValueSet
 from rocisa.container import EXEC, VCC, MUBUFModifiers
 
+
+
```
davidd-amd, line 27: Remove white space diff

tensilelite/Tensile/Ops/SoftmaxGenerator.py
```diff
@@ -31,6 +31,9 @@
 import json
 import subprocess
 from contextlib import contextmanager
+import os.path
+import sys
+
```
davidd-amd, line 36: Where did this come from?

stellaraccident:
> @stellaraccident I was concerned that CMake would be able to cope with unix paths in custom commands when on Windows. However, after some testing it seems to implicitly convert paths. Can you confirm my understanding is correct?
> 
> I just need to create a full build of the impacted binaries (extop and matrix transform) and compare them with what is generated on develop to be sure I haven't inadvertently dropped symbols. Once that is done I will mark as ready for review.

It is usually ok to use paths with forward slashes these days without conversion in these tools. There are still some dark holes where this is not true, but I typically just deal with those when they arise (and I haven't had to do so in a long time).

The main thing to watch out for us that path separators (colon vs semicolon) must be handled manually as they are actually ambiguous and prone to various escaping issues. Typically only impacts PATH like environment variables and usually just be a bit of special casing.

davidd-amd:
@stellaraccident because we have rocisa installed in the binary dir (needed at build time) and Tensile installed into a pyenv (during config time) and required at build time, I had to do some things that I didn't like. Namely setting the variable `python_command` and setting PYTHONPATH when invoking the python scripts. Let me know how these details will impact the rock and if you think there is a way around it?

KKyang:
Please also make sure this works in hipsparselt as well

library/CMakeLists.txt
```diff
@@ -68,13 +68,13 @@ set(BIN_INSTALL_DIR ${CMAKE_INSTALL_BINDIR})
 set(LIB_INSTALL_DIR ${CMAKE_INSTALL_LIBDIR})
 set(INCLUDE_INSTALL_DIR ${CMAKE_INSTALL_INCLUDEDIR})
 
-# Include sources
-include(src/CMakeLists.txt)
-
 # Create hipBLASLt library
-add_library(hipblaslt ${hipblaslt_source} ${RocRoller_SRC} ${hipblaslt_headers_public})
+add_library(hipblaslt ${RocRoller_SRC} ${hipblaslt_headers_public})
```
bstefanuk: Are these sources still properly added to the target?

library/src/CMakeLists.txt
```diff
@@ -21,15 +21,18 @@
...
+  add_subdirectory(amd_detail/rocblaslt)
+  target_sources(hipblaslt
+      PRIVATE
+          amd_detail/hipblaslt.cpp
+          amd_detail/hipblaslt-ext.cpp
+          amd_detail/hipblaslt-ext-op.cpp
+          ${hipblaslt_source_common}
+          ${rocblaslt_source}
+  )
+  #set_source_files_properties(${hipblaslt_source} PROPERTIES LANGUAGE CXX COMPILE_OPTIONS "-std=c++17")
```
bstefanuk, line 37: Should we still be setting the cxx 17 compile feature on this target?

library/src/amd_detail/rocblaslt/src/extops/CMakeLists.txt
```diff
@@ -0,0 +1,138 @@
...
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+#
+# ########################################################################
+
+set(ops_path ${PROJECT_SOURCE_DIR}/tensilelite/Tensile/Ops)
+
+string(REGEX MATCHALL "gfx[a-z0-9]+" archs "${Tensile_ARCHITECTURE}")
```
stellaraccident, line 31: Does this account for the generic target names? I thought I saw some of those contain hyphens. Might be better to match word characters (`\w`). Another way to do this (I think you are just trying to split by spaces is:

```
string(REPLACE " " ";" archs "${Tensile_ARCHITECTURE})
```
Then in your loop:

```
if(NOT arch)
  continue()
endif()
```

library/src/amd_detail/rocblaslt/src/extops/CMakeLists.txt
```diff
@@ -0,0 +1,138 @@
...
+list(REMOVE_DUPLICATES archs)
+
+set(python_command "${VIRTUALENV_BIN_DIR}/${VIRTUALENV_PYTHON_EXENAME}")
+
+foreach(arch IN LISTS archs)
+    add_library(ExtOpObj_${arch} OBJECT)
+    add_dependencies(ExtOpObj_${arch} rocisa)
+    target_compile_options(ExtOpObj_${arch}
+        PRIVATE
+        -Wno-unused-command-line-argument -x assembler -target amdgcn-amd-amdhsa -mcode-object-version=4 -mcpu=${arch} -mwavefrontsize64 -c
```
stellaraccident, line 42: I assume that `-mwavefrontsize64` is going to need to be changed soon, right? But I guess no worse here than hardcoded in a script.

library/src/amd_detail/rocblaslt/src/extops/CMakeLists.txt
```diff
@@ -0,0 +1,138 @@
...
+    )
+    target_sources(ExtOpObj_${arch}
+        PRIVATE
+            ${CMAKE_CURRENT_BINARY_DIR}/L_256_4_1_${arch}.s
```
stellaraccident, line 46: I wouldn't mind having a legend comment for what these different prefixes mean and what they need to mate up with elsewhere in the project (presumably in the generator).

(there is a lot of duplication of these generated file lists, but I'm willing to believe that it is better to be explicit than to try to meta-program them)

library/src/amd_detail/rocblaslt/src/extops/CMakeLists.txt
```diff
@@ -0,0 +1,138 @@
...
+            ${CMAKE_CURRENT_BINARY_DIR}/S_8_32_${arch}.s
+            ${CMAKE_CURRENT_BINARY_DIR}/S_16_16_${arch}.s
+            ${CMAKE_CURRENT_BINARY_DIR}/S_4_64_${arch}.s
+            ${CMAKE_CURRENT_BINARY_DIR}/S_2_128_${arch}.s
+            ${CMAKE_CURRENT_BINARY_DIR}/S_1_256_${arch}.s
+            ${CMAKE_CURRENT_BINARY_DIR}/A_S_S_256_4_${arch}.s
+            ${CMAKE_CURRENT_BINARY_DIR}/A_H_H_256_4_${arch}.s
+            ${CMAKE_CURRENT_BINARY_DIR}/A_H_S_256_4_${arch}.s
+            ${CMAKE_CURRENT_BINARY_DIR}/A_S_H_256_4_${arch}.s
+        COMMAND PYTHONPATH=${PROJECT_BINARY_DIR}/lib ${python_command} LayerNormGenerator.py -o ${CMAKE_CURRENT_BINARY_DIR}/L_256_4_1_${arch}.s -w 256 -c 4 --sweep-once 1 --arch ${arch}
```
stellaraccident: I would hoist `PYTHONPATH=${PROJECT_BINARY_DIR}/lib ${python_command}` to the top of the file as, say `set(python_launch_prefix "PYTHONPATH=${PROJECT_BINARY_DIR}/lib" "${python_venv_executable}")` and then use it without quotes in all of these places.

Leave a TODO that `python_venv_executable` should be replaced with `Python3_EXECUTABLE when the venv is no longer needed.

Future us will thank us the next time you need to manipulate the PYTHONPATH, set additional options, etc.

library/src/amd_detail/rocblaslt/src/kernels/CMakeLists.txt
```diff
@@ -0,0 +1,47 @@
...
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+#
+# ########################################################################
+
+# Need to support Release/Debug
+
+string(REGEX MATCHALL "gfx[a-z0-9]+" archs "${Tensile_ARCHITECTURE}")
```
stellaraccident, line 26: Same note on splitting the Tensile_ARCHITECTURE string as above.

library/CMakeLists.txt
```diff
@@ -68,13 +68,22 @@ set(BIN_INSTALL_DIR ${CMAKE_INSTALL_BINDIR})
...
 
+# Include sources
+add_subdirectory(src)
+
+if(USE_ROCROLLER)
+    set_source_files_properties(
+      "${CMAKE_CURRENT_SOURCE_DIR}/src/amd_detail/rocblaslt/src/rocroller_host.cpp"
+        PROPERTIES 
+            LANGUAGE CXX 
+            COMPILE_OPTIONS "-std=c++20;-x;c++"
```
stellaraccident, line 83: I'm not 100% sure where the need for this is coming from. Can we take an issue to uniformly upgrade the project defaults to c++20 (at the project level)? Seems like the `-x c++` and `LANGUAGE CXX` should not be needed since CMake should infer that.

Should add a TODO to remove this. I imagine it was copy-pasta'd from somewhere, but typically this kind of thing is done at the target level with something like `target_compile_features(my_library PUBLIC cxx_std_14)` when it needs to be changed different from the project wide setting.

stellaraccident (dismissed):
Generally lgtm. Left some small nits/suggestions. Feel free to ping for an additional stamp if it won't let you land after making requested changes.

library/CMakeLists.txt
davidd-amd: yes we use target_sources now

library/src/CMakeLists.txt
davidd-amd: It is done elsewhere.

library/CMakeLists.txt
davidd-amd: Really rocroller should be setting these requirements in that project. I don't think that source file actually uses any C++20 but rocroller uses concepts in header files included transitively by hipblaslt.

library/src/amd_detail/rocblaslt/src/kernels/CMakeLists.txt
davidd-amd: I will test this for some generic targets however we are generating assembly for a specific isa and I don't know if we should be doing that for generics.

library/CMakeLists.txt
KKyang: I think we already have CXX standard in the root folder of hipblaslt. We can just change it from 17 to 20.

library/src/amd_detail/rocblaslt/src/extops/CMakeLists.txt
davidd-amd: I was considering doing that too. How do we accommodate a user that has already set their PYTHONPATH to point to rocisa or pip installed it?

library/CMakeLists.txt
stellaraccident: Yes, what KKyang says would be best. This kind of thing ends up living forever as copy-pasta if not controlled for. Minimally, if you'd like to come back and fix it, at least leave a TODO.

library/src/amd_detail/rocblaslt/src/extops/CMakeLists.txt
davidd-amd: There is indeed duplication which is why I moved it into a stand alone file. I tend to lean towards being explicit in cases like this but I am not that attached to being explicit.

library/src/amd_detail/rocblaslt/src/extops/CMakeLists.txt
stellaraccident: Sounds like the realm of an option `Tensile_USE_BUNDLED_PYTHON_LIBS` or something. Another reason to put the env var magic at the top of the file, as we'll probably want to hoist it either further up in the future or control with options.

library/src/amd_detail/rocblaslt/src/extops/CMakeLists.txt
stellaraccident: I ran through the same thought process and concluded your choice was probably the right one for this case. Only suggestion I would make would be to quote each file so that stray spaces don't change the syntax of the list.

library/src/amd_detail/rocblaslt/src/extops/CMakeLists.txt
davidd-amd: I should ask around to figure out if we should be building these files for all archs. It could be that we only ever intended to do so for archs with wavefront size of 64 but never gated out the other archs.

davidd-amd:
> Please also make sure this works in hipsparselt as well

I don't see any evidence that `TensileCreateExtOpLibraries` is used or the various py files in the Ops dir. What else needs to be done to confirm this works?

stellaraccident:
> > Please also make sure this works in hipsparselt as well
> 
> I don't see any evidence that `TensileCreateExtOpLibraries` is used or the various py files in the Ops dir. What else needs to be done to confirm this works?

As an aside, we need to rework that hipSPARSELt dep mechanism as it is an extreme anti-pattern. However, even though an ill-advised way to take a dep, I don't see an interaction potential between this PR and that project.

KKyang (approved):

stellaraccident (approved):

PR #1966, opened 2025-04-21
[Windows] Undo clang-format that put windows headers in the wrong order.

stellaraccident:


davidd-amd (approved):
How do we prevent the formatter from doing this again? Can we tell the formatter to skip those lines?

tensilelite/rocisa/rocisa/src/helper.cpp
```diff
@@ -29,8 +29,9 @@
 
 #ifdef WIN32
 
-#include <dbghelp.h>
 #include <windows.h>
+// windows.h must be loaded before other windows headers.
+#include <dbghelp.h>
```
KKyang, line 34: Please wrap 
// clang-format off/on around these headers

KKyang: Or just simply add an empty line

stellaraccident: I usually just use header groups like this as clang-format will not sort across different groups of headers (blank lines or comments delineate groups).

KKyang (approved):

stellaraccident:
Landing as-is since this is in a CI untested ifdef block that compile fails at head.

PR #1967, opened 2025-04-22
Windows compatibility changes for C++ code.

stellaraccident:
Includes findings/fixes from #1864.

Fixes all portability issues that resulted in compilation errors in C++ code. There are still warnings and some possible functional issues which need more work, but this is a first step we should commit. This patch was written/tested in a combined tree that is capable of building on Windows, and we need to verify that any rewritten paths do not regress Posix. With this and new CMake patches, the project compiles on Windows. It has some linkage issues on the tests which need further investigation (we have another branch with some of these findings/workarounds).

Several additional changes from the above:

* Rewrote shared library finding code to be based on resolving a library by address vs scanning for a named library. While the code for both is not identical on Windows/Posix, the same semantics for such a helper are directly available.
* Rewrote a number of things for clarity and to use std::filesystem vs use of low level path and string munging.
* Removed several conditional `#ifdef _WIN32` when the branches could be expressed in a portable manner.

Explanations for some changes:

* `s/ABC/ABC_dims/`: `ABC` is used as a bareword in Windows headers.
* `s/WIN32/_WIN32/`: `_WIN32` is the modern/correct way to detect if compiling on Windows. It is defined in all cases whereas `WIN32` is only defined in some situations.
* `ssize_t` is non-standard. In the place it is used, did the usual typedef from the `SSIZE_T` type that windows provides in its headers.
* `portable_uniform_int_distribution` template specializations instead of direct use of `std::uniform_int_distribution` enables compilation on standards compliant C++ standard libs, which do not include char specializations (includes MSVC but also some others).
* Remove non-standard `DL_LIB` ways to include `-ldl` on Posix. That is what `CMAKE_DL_LIBS` is for.
* Remove HIPBLASLT_EXPORT from all deleted members. This is illegal and was not being caught in Posix builds.
* Replace `libgen.h` with `std::filesystem` equivalents. We should not allow `libgen.h` in new code.
* `u_int16_t` is not a real/standard type. Use `uint16_t`.
* Rewrite some unreadably nested ifdefs in initialization code to use normal if statements and compile both paths vs leaving some compilation paths only for certain build configurations. We should enforce this as a coding convention.

### DIFF
--- a/clients/common/hipblaslt_init_device.cpp
+++ b/clients/common/hipblaslt_init_device.cpp
@@ -116,7 +116,7 @@ __device__ int8_t random_hpl(size_t idx)
 }
 
 template <typename T>
-void hipblaslt_init_device(ABC                      abc,
+void hipblaslt_init_device(ABC_dims                 abc,
                            hipblaslt_initialization init,
                            bool                     is_nan,
                            T*                       A,
@@ -140,11 +140,11 @@ void hipblaslt_init_device(ABC                      abc,
         switch(init)
         {
         case hipblaslt_initialization::rand_int:
-            if(abc == ABC::A || abc == ABC::C)
+            if(abc == ABC_dims::A || abc == ABC_dims::C)
                 fill_batch(A, M, N, lda, stride, batch_count, [](size_t idx) -> T {
                     return random_int<T>(idx);
                 });
-            else if(abc == ABC::B)
+            else if(abc == ABC_dims::B)
             {
                 stride = std::max(lda * N, stride);
                 fill_batch(A, M, N, lda, stride, batch_count, [stride, lda](size_t idx) -> T {
@@ -158,14 +158,14 @@ void hipblaslt_init_device(ABC                      abc,
             break;
         case hipblaslt_initialization::trig_float:
             stride = std::max(lda * N, stride);
-            if(abc == ABC::A || abc == ABC::C)
+            if(abc == ABC_dims::A || abc == ABC_dims::C)
                 fill_batch(A, M, N, lda, stride, batch_count, [M, N, stride, lda](size_t idx) -> T {
                     auto b = idx / stride;
                     auto j = (idx - b * stride) / lda;
                     auto i = (idx - b * stride) - j * lda;
                     return T(sin(double(i + j * M + b * M * N)));
                 });
-            else if(abc == ABC::B)
+            else if(abc == ABC_dims::B)
                 fill_batch(A, M, N, lda, stride, batch_count, [M, N, stride, lda](size_t idx) -> T {
                     auto b = idx / stride;
                     auto j = (idx - b * stride) / lda;
@@ -179,15 +179,15 @@ void hipblaslt_init_device(ABC                      abc,
             });
             break;
         case hipblaslt_initialization::special:
-            if(abc == ABC::A)
+            if(abc == ABC_dims::A)
                 fill_batch(A, M, N, lda, stride, batch_count, [](size_t idx) -> T {
                     return T(hipblasLtHalf(65280.0));
                 });
-            else if(abc == ABC::B)
+            else if(abc == ABC_dims::B)
                 fill_batch(A, M, N, lda, stride, batch_count, [](size_t idx) -> T {
                     return T(hipblasLtHalf(0.0000607967376708984375));
                 });
-            else if(abc == ABC::C)
+            else if(abc == ABC_dims::C)
                 fill_batch(A, M, N, lda, stride, batch_count, [](size_t idx) -> T {
                     return T(pseudo_random_device(idx) % 10 + 1.f);
                 });
@@ -213,7 +213,7 @@ void hipblaslt_init_device(ABC                      abc,
     }
 }
 
-void hipblaslt_init_device(ABC                      abc,
+void hipblaslt_init_device(ABC_dims                 abc,
                            hipblaslt_initialization init,
                            bool                     is_nan,
                            void*                    A,
--- a/clients/common/utility.cpp
+++ b/clients/common/utility.cpp
@@ -24,15 +24,22 @@
  *
  *******************************************************************************/
 
-#include "utility.hpp"
+#ifdef _WIN32
+#include <windows.h>
+// Must include windows.h before dependent headers.
+#include <libloaderapi.h>
+#endif
+
 #include "d_vector.hpp"
+#include "utility.hpp"
+#include <atomic>
 #include <chrono>
 #include <cstdlib>
+#include <fcntl.h>
+#include <limits.h>
 #include <new>
 #include <stdexcept>
 #include <stdlib.h>
-
-#include <fcntl.h>
 
 #include "Tensile/Source/client/include/Utility.hpp"
 
@@ -48,27 +55,65 @@ namespace fs = std::experimental::filesystem;
 
 /* ============================================================================================ */
 // Return path of this executable
+static std::string get_self_path()
+{
+#ifdef _WIN32
+    std::string result(MAX_PATH + 1, '\0');
+    DWORD       length = 0;
+    for(;;)
+    {
+        length = GetModuleFileNameA(nullptr, result.data(), result.size());
+        if(length < result.size() - 1)
+        {
+            result.resize(length);
+            return result;
+        }
+        result.resize(result.size() * 2);
+    }
+#else
+    return std::string(realpath("/proc/self/exe", 0));
+#endif
+}
+
 std::string hipblaslt_exepath()
 {
-    std::string pathstr;
-    char*       path = realpath("/proc/self/exe", 0);
-    if(path)
-    {
-        char* p = strrchr(path, '/');
-        if(p)
-        {
-            p[1]    = 0;
-            pathstr = path;
-        }
-        free(path);
-    }
-    return pathstr;
+    fs::path exepath(get_self_path());
+    exepath.remove_filename();
+    std::string result = exepath.string();
+    if(result.empty())
+        result.append("/");
+    return result;
 }
 
 /* ============================================================================================ */
 // Temp directory rooted random path
+// TODO: This function is inherently unsafe because it returns a string vs an open
+// file handle which will block further colliding creates. On Posix, this will leak
+// a file handle for the life of the process. On Windows, there is no way to ensure that
+// the created file name is unique without racing. To counter this on Windows, we
+// also include the process id and a process specific counter in the generated name,
+// as that will at least race consistently vs based on a random number generator
+// collision. This and its consumers should be rewritten and under no circumstances
+// copied to new code.
 std::string hipblaslt_tempname()
 {
+#ifdef _WIN32
+    static std::atomic<int> counter;
+    // Generate "/tmp/rocblas-XXXXXX" like file name
+    const std::string alphanum     = "0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuv";
+    int               stringlength = alphanum.length() - 1;
+    std::string       uniquestr    = "hipblaslt-";
+    uniquestr.append(std::to_string(GetCurrentProcessId()));
+    uniquestr.append("-");
+    uniquestr.append(std::to_string(counter.fetch_add(1)));
+    uniquestr.append("-");
+
+    for(auto n : {0, 1, 2, 3, 4, 5})
+        uniquestr += alphanum.at(rand() % stringlength);
+
+    fs::path tmpname = fs::temp_directory_path() / uniquestr;
+    return tmpname.string();
+#else
     char tmp[] = "/tmp/hipblaslt-XXXXXX";
     int  fd    = mkostemp(tmp, O_CLOEXEC);
     if(fd == -1)
@@ -78,6 +123,7 @@ std::string hipblaslt_tempname()
     }
 
     return std::string(tmp);
+#endif
 }
 
 /* ============================================================================================ */
@@ -145,7 +191,7 @@ double get_time_us_no_sync(void)
 
 /* ============================================================================================ */
 /*  device query and print out their ID and name; return number of compute-capable devices. */
-int64_t query_device_property(int device_id, hipDeviceProp_t &props)
+int64_t query_device_property(int device_id, hipDeviceProp_t& props)
 {
     int             device_count;
     hipblasStatus_t status = (hipblasStatus_t)hipGetDeviceCount(&device_count);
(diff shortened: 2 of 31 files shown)


stellaraccident:
Reviewers: please provide a detailed review, especially focusing on code paths that were not just forked between Posix/Windows but were rewritten to be portable. The intent was that these should be NFC, but the original code had numerous bugs which I could spot (and expect were corner cases that were not being exercised), and I made judgment calls to fix them vs preserve broken behavior.

clients/common/hipblaslt_init_device.cpp
```diff
@@ -116,7 +116,7 @@ __device__ int8_t random_hpl(size_t idx)
 }
 
 template <typename T>
-void hipblaslt_init_device(ABC                      abc,
+void hipblaslt_init_device(ABC_dims                 ABC_dims,
```
davidd-amd: Maybe it's just late but are we using a type for a variable name? I haven't ever considered doing that so I am honestly not sure if it is valid code.

stellaraccident: Oh whoops, find/replace mistake. Let me switch that back. Thanks for the eyes. It is actually legal but indeed, not a good idea.

clients/include/hipblaslt_random.hpp
```diff
@@ -69,6 +69,27 @@ inline void hipblaslt_seedrand()
 /*! \brief  Random number generator which generates NaN values */
 class hipblaslt_nan_rng
 {
+    // The C++ standard does not support std::uniform_int_distribution on all
+    // data types (notably not on char), although some libc++ implementation do.
+    // Notably, MSVC does not. Therefore, explicitly specialize non-standard
+    // types to narrow from a supported type.
```
davidd-amd, line 75: I did not know that and I've sat through a few of Walter Brown's talks. I'm a big fan of the random header except that it doesn't provide a skip mechanism that I'm aware of.

Does this just fail to compile on windows without the specializations?

clients/include/hipblaslt_random.hpp
```diff
@@ -91,7 +112,13 @@ class hipblaslt_nan_rng
     template <typename T, std::enable_if_t<std::is_integral<T>{}, int> = 0>
     explicit operator T()
     {
-        return std::uniform_int_distribution<T>{}(t_hipblaslt_rng);
+        return portable_uniform_int_distribution<T>();
```
davidd-amd, line 115: why don't we need to pass `t_hipblaslt_rng` here?

stellaraccident: Look above - inlined into `portable_uniform_int_distribution`. It's all statics so should be equivalent.

clients/include/hipblaslt_random.hpp
stellaraccident: TBH - I didn't either, but MSVCRT has a relatively explicit static_assert that sent me down the rabbit hole.

stellaraccident: (for years we cried -- "why isn't MSVC standards compliant". These days it is more "why is MSVC so aggressively standards compliant???" :) )

library/CMakeLists.txt
```diff
@@ -155,7 +155,7 @@ endif()
 # Target link libraries
 if(NOT BUILD_CUDA)
 # Target link libraries
-  target_link_libraries(hipblaslt PRIVATE hip::device ${DL_LIB})
+  target_link_libraries(hipblaslt PRIVATE hip::device ${CMAKE_DL_LIBS})
```
davidd-amd, line 158: Do we want to gaurd DL with a `$<BUILD_INTERFACE:>? See post:

https://discourse.cmake.org/t/cmake-dl-libs/1159/4

library/src/amd_detail/hipblaslt-ext-op-internal.hpp
```diff
@@ -660,7 +660,7 @@ namespace hipblaslt_ext
         explicit ExtOpMasterLibrary(const std::string& libPath)
             : libPath(libPath)
         {
-            libDir = std::string(dirname(&this->libPath[0]));
+            libDir = std::filesystem::path(this->libPath).parent_path().string();
```
davidd-amd, line 663: big improvement

library/src/amd_detail/rocblaslt/src/rocblaslt_auxiliary.cpp
```diff
@@ -2119,35 +2123,55 @@ std::string rocblaslt_internal_get_arch_name()
...
+#ifdef _WIN32
+std::string rocblaslt_internal_get_so_path()
 {
-    // uncomment to see all dependent .so files
-    // fprintf(stderr, "hipblaslt so file: %s\n", hdr_info->dlpi_name);
-    std::pair<std::string, std::string>* typedData
-        = reinterpret_cast<std::pair<std::string, std::string>*>(data);
-    if(hdr_info->dlpi_name && strstr(hdr_info->dlpi_name, typedData->second.c_str()))
+    HMODULE hModule = NULL;
+    if(!GetModuleHandleEx(GET_MODULE_HANDLE_EX_FLAG_FROM_ADDRESS,
```
KKyang, line 2139: Is this API from hipruntime? Some APIs in hipruntime have large latencies so may impact the overall performance, but I'm not sure if this one will impact or not.

library/src/amd_detail/hipblaslt-ext-op.cpp
```diff
@@ -162,22 +163,22 @@ namespace
             return libPath;
         }
 
-        auto        soPath = rocblaslt_internal_get_so_path("hipblaslt");
-        std::string libPath(dirname(&soPath[0]));
+        auto                  soPath = rocblaslt_internal_get_so_path();
```
davidd-amd, line 166: why remove `"hipblaslt"`

library/src/amd_detail/hipblaslt-ext-op.cpp
```diff
@@ -162,22 +163,22 @@ namespace
...
-        if(rocblaslt_internal_test_path(libPath + "/../Tensile/library"))
-            libPath += "/../Tensile/library";
-        else if(rocblaslt_internal_test_path(libPath + "library"))
-            libPath += "/library";
-        else
-            libPath += "/hipblaslt/library";
+        auto pathIfExists = [](std::filesystem::path p) -> std::optional<std::filesystem::path> {
+            if(std::filesystem::exists(p))
+                return p;
+            return {};
```
davidd-amd, line 172: is `{}` equivalent to `std::nullopt`?

library/CMakeLists.txt
stellaraccident: Will review the post, but I don't *think* it should be necessary, right? Private link libs do not escape to consumers, including imported targets, I think. Also, BUILD_INTERFACE is a real pita for things that can be multiple (although I don't know any that are in this case) because it treats the pointed to entity as a single item in the list.

stellaraccident: (read that post, the critical thing there is that it is a PUBLIC scoped linked library. And even then, it was deemed of marginal benefit to isolate. This should be a completely non-issue with a PRIVATE scoped library)

library/src/amd_detail/hipblaslt-ext-op.cpp
```diff
@@ -162,22 +163,22 @@ namespace
...
-        {
-            return libPath;
-        }
+        if(auto p
+           = pathIfExists(libPath / ".." / "Tensile" / "library" / "hipblasltExtOpLibrary.dat"))
+            return p->string();
+        if(auto p = pathIfExists(libPath / "library" / "hipblasltExtOpLibrary.dat"))
+            return p->string();
+        if(auto p = pathIfExists(libPath / "hipblaslt" / "library" / "hipblasltExtOpLibrary.dat"))
+            return p->string();
```
davidd-amd, line 181: I've learned today when reworking the build system that this pattern shows up in three different cpp files. Not worth holding up a merge but if you have to push another commit consider lifting this code into a reusable function.

library/src/amd_detail/hipblaslt-ext-op.cpp
stellaraccident: Look at the implementation: the prior implementation on Posix was scanning the dl table looking for the hipblaslt library by name. But this is not possible/reliable on Windows. What I see done more often is to use dladdr to resolve a loaded shared library by a symbol that it is known to contain. No string munging, guaranteed precise, and portable across Posix/Windows/MacOS. In this case, no callers were asking this helper for anything but hipblaslt, so I just hardcoded the symbol to use. If that ever changes (which I doubt it will), we should add a void* for the symbol to look up vs going back to string munging.

library/src/amd_detail/hipblaslt-ext-op.cpp
stellaraccident: Yes. Probably an older form that my fingers memorialized (i.e. default constructor of optional)

library/src/amd_detail/rocblaslt/src/rocblaslt_auxiliary.cpp
stellaraccident: GetModuleHandleEx? That's a Windows API. It's a bit old/gross and when used in this way, it is equivalent to `dladdr`.

library/src/amd_detail/hipblaslt-ext-op.cpp
stellaraccident: Yeah, I learned that too but was tired by the time I found the third one :/

library/src/amd_detail/rocblaslt/src/rocblaslt_auxiliary.cpp
```diff
@@ -2119,35 +2123,55 @@ std::string rocblaslt_internal_get_arch_name()
...
+                          // Should be the address of code in this library.
+                          (LPCTSTR)rocblaslt_internal_get_so_path,
+                          &hModule))
     {
-        typedData->first.assign(hdr_info->dlpi_name);
-        return 1;
+        throw std::runtime_error("Cannot get module for function");
     }
-    return 0;
-}
```
davidd-amd, line 2142: Having a hard time understanding how we can delete this but not at all sad to see it go.

library/src/amd_detail/hipblaslt-ext-op.cpp
stellaraccident: I need to pull this patch over to my Linux box and I'm sure there will be fixes needed on that side. I'll refactor it then.

library/src/amd_detail/rocblaslt/src/rocblaslt_auxiliary.cpp
stellaraccident: See comment above where you asked about the callsite. I think I explained it as best I could there but you are probably reading in reverse order.

library/src/amd_detail/rocblaslt/src/rocblaslt_auxiliary.cpp
```diff
@@ -2119,35 +2123,55 @@ std::string rocblaslt_internal_get_arch_name()
...
 
-std::string rocblaslt_internal_get_so_path(const std::string& keyword)
+    std::string path;
+    path.resize(256);
+    for(;;)
+    {
+        auto stored_size = GetModuleFileNameA(hModule, path.data(), path.size());
+        if(stored_size < path.size())
+        {
+            // Success: size to what was stored (which does not include NUL).
```
davidd-amd, line 2154: NUL terminator?

library/src/amd_detail/rocblaslt/src/utility.cpp
```diff
@@ -25,7 +25,18 @@
...
+
+#ifdef _WIN32
+#include <process.h>
+#else
 #include <unistd.h>
+#endif
+
+#include <iostream>
+#include <memory>
+#include <string>
```
davidd-amd, line 38: Missing includes on Windows?

tensilelite/Tensile/Source/lib/include/Tensile/Serialization/PlaceholderLibrary.hpp
```diff
@@ -32,8 +32,9 @@
 #include <Tensile/MasterSolutionLibrary.hpp>
 #include <Tensile/PlaceholderLibrary.hpp>
 //Replace std::regex, as it crashes when matching long lines(GCC Bug #86164).
-#ifdef WIN32
+#ifdef _WIN32
 #include "shlwapi.h"
+#pragma comment(lib, "shlwapi.lib")
```
davidd-amd, line 37: I've never used `pragma comment` what is this doing for us?

stellaraccident: It's the usual way of including a windows specific standard library that isn't in msvc's implicit link list. It can be abused but for this narrow case of an optional standard library that must be linked for the header to work, it's typically the least bad. You see it a lot for advapi and some of the other things that are not implicitly linked for reasons that are lost to time (kind of like how no one actually considers libm separate from the crt anymore and most build systems just always include it).

library/src/amd_detail/rocblaslt/src/utility.cpp
stellaraccident: Probably. I didn't root cause every missing include issue. Just added them when the compiler said so. Msvc tends to be more strict about not pulling in transitive STL headers.

stellaraccident:
PTAL - Rebased and fixed a couple of minor issues on the Posix side (dladdr call needed a cast, and some missed `abc` rename fixes). gfx1100 tests pass locally for me. Will see what the CI says.

clients/common/utility.cpp
```diff
@@ -48,27 +55,65 @@ namespace fs = std::experimental::filesystem;
...
-            pathstr = path;
+            result.resize(length);
+            return result;
         }
-        free(path);
+        result.resize(result.size() * 2);
     }
-    return pathstr;
+#else
+    return std::string(realpath("/proc/self/exe", 0));
```
bstefanuk, line 74: If realpath cannot resolve /proc/self/exe then a nullptr is returned, which would cause undefined behaviour here when constructing the std::string. However, I understand this is an edge case and I'm not too familiar with /proc/self/exe. Is this guaranteed to be defined on all Posix-compliant systems?

library/src/amd_detail/hipblaslt-ext-op.cpp
davidd-amd: Great thanks!

library/src/amd_detail/rocblaslt/src/rocblaslt_auxiliary.cpp
davidd-amd: yeah this was a hard diff to follow. Maybe we could do a quick walkthrough in the future.

davidd-amd:
@stellaraccident I am rebooting tests because there was a spurious docker failure. I am going to approve now because tests are otherwise passing.

davidd-amd (approved):

stellaraccident:
Determined offline with the CI team that the failures are spurious and to be ignored.

stellaraccident:
Apologies for squash/merging with a bad default description. I'm not used to the project being configured to override your PR title/description when landing. I have updated the project settings for Squash settings to have the commit message be "Pull request title and description" so that we get better commit messages overall.

PR #1969, opened 2025-04-22
Add rocisa path into the setup.py

vin-huang:
* this is used to let other library who also uses tensilelite as a backend, can download the rocisa
 
 For hipsparselt to build correctly with rocisa enabled.
 https://github.com/ROCm/hipSPARSELt/pull/229

### DIFF
--- a/tensilelite/MANIFEST.in
+++ b/tensilelite/MANIFEST.in
@@ -1,5 +1,5 @@
 recursive-include Tensile *.py *.h *.hpp *.cpp *.txt *.cmake *.yaml
-recursive-include rocisa/rocisa/include *.hpp
+recursive-include rocisa *.cpp *.hpp *.txt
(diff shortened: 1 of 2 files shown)


jichangjichang (approved):

stellaraccident:
FYI - we cannot support this mode of taking an interproject dependency and it will keep breaking until we rework how hipsparselt is doing this. We can discuss at office hours. 

Fine to proceed to unblock, but there can be no expectation that this does not break further in the current state.

vin-huang:
the failed is at when building sles docker missing nija.
[mci/precheckin(main)](http://math-ci.amd.com/job/main/job/precheckin/job/hipBLASLt/job/PR-1969/1/display/redirect)

KKyang (approved):

PR #1971, opened 2025-04-22
Fix waitcnt and enable PGR2 + PLR for DTL

b-shi:
Fix waitcnt and enable PGR2 + PLR for DTL
 - Removed previous restriction
 - Moved vmcnt for PGR2 to end of main loop (previously in preloop) and only used for DTL. 
   - Fix ported from Tensile [here](https://github.com/ROCm/Tensile/blob/develop/Tensile/KernelWriter.py#L2179).


tensilelite/Tensile/KernelWriter.py
```diff
@@ -2812,11 +2812,6 @@ def kernelBody( self, kernel, tensorParametersA, tensorParametersB ):
       # force to generate 2 loop bodies
       loopCopies = 2
 
-    if kernel["PrefetchGlobalRead"] == 2:
```
msujon-AMD, line 2815: Is it specific to DTL? We don't have any side effect on other cases, right?

tensilelite/Tensile/Tests/common/gemm/gfx950/dtl.yaml
```diff
@@ -378,4 +378,4 @@ BenchmarkProblems:
           - Exact: [256, 256, 1, 128]
           - Exact: [256, 256, 1, 256]
           - Exact: [256, 256, 1, 512]
-          - Exact: [4096, 8192, 1, 128]
\ No newline at end of file
+          - Range: [[4096], [8192], [1], [64,64,384]]
```
msujon-AMD, line 381: Does this yaml exercise this special case? If we had failure for any specific size, you can mark it as well.

tensilelite/Tensile/KernelWriter.py
b-shi: Yes, it should be DTL specific. I previously added that to fix a DTL-related issue mentioned [here](https://github.com/ROCm/hipBLASLt/pull/1833#issuecomment-2796483368). It was only failing for DTL case, so adding this for non-DTL might cause some perf impact for latency-bound cases.

tensilelite/Tensile/Tests/common/gemm/gfx950/dtl.yaml
b-shi: Yeah, that M/N size triggered the [issue](https://github.com/ROCm/hipBLASLt/pull/1833#issuecomment-2796483368) previously.

tensilelite/Tensile/KernelWriter.py
msujon-AMD: Okay, it was introduced in previous PR and we don't need it anymore (since it is moved to NLL with DTL specific checking). LGTM.

msujon-AMD (dismissed):

tensilelite/Tensile/KernelWriter.py
```diff
@@ -2852,6 +2847,11 @@ def kernelBody( self, kernel, tensorParametersA, tensorParametersB ):
 
     module.addComment1("Before NLL: Check VGPR.checkin for INT8 LW")
 
+    if kernel["PrefetchGlobalRead"] == 2 and (kernel["DirectToLdsA"] or kernel["DirectToLdsB"]):
```
nakajee: We should not need this here.
This will make prefetch less effective.
This should be before this code.
https://github.com/ROCm/hipBLASLt/blob/develop/tensilelite/Tensile/KernelWriter.py#L2183

I think you can refer Tensile code.
I can see wait is added at this location in Tensile case.

nakajee:
I did not realize this change in your previous PR.
By the way, why only PLR1 failed? PLR>1+PGR2 passed without any wait?

tensilelite/Tensile/KernelWriter.py
nakajee: We might have some bug fixes not ported from Tensile to tensilelite (especially DTL related).
Please make sure you checked logic in Tensile.

b-shi: Ah, thanks for the pointer! That is a better place in the NLL to place the wait instead of at the beginning.

Edit: Copied over the fix in Tensile [here](https://github.com/ROCm/Tensile/blob/develop/Tensile/KernelWriter.py#L2179).

b-shi:
> I did not realize this change in your previous PR. By the way, why only PLR1 failed? PLR>1+PGR2 passed without any wait?

`Context`: in the previous PR, the no global load loop was missing a wait for the second set of global loads. This would cause failures for small K, where we prefetched but did not enter the main loop. The fix was to either add a wait+barrier in the preloop or in the no global load loop. In hindsight it should have been added in the NGLL.

For PLR1 + DTL, the waitcnt [here](https://github.com/ROCm/hipBLASLt/blob/develop/tensilelite/Tensile/KernelWriter.py#L2574) for DTL is incorrect (fix [here](https://github.com/ROCm/hipBLASLt/pull/1977)) since codegen is still emitting some global loads after that wait. This causes some race condition to occur in the NGLL (with the initial fix). Moving the wait from the previous PR to the NGLL fixes both issue.

jichangjichang (approved):

msujon-AMD (approved):

PR #1977, opened 2025-04-24
Fix waitcnt in DTL main loop

b-shi:
- Fix incorrect `s_waitcnt` placement for DTL in main loop. Previously the `s_waitcnt` in the main loop did not wait for all global loads.
- For PGR2 increase `vmcnt` value to wait only on buffer loads from previous iteration.
  - This was a significant performance bottleneck and improves PGR2+PLR1+DTL performance by 20-40%
- Add option to skip waitcnt + barrier in DTL loads for second param.

### DIFF
--- a/tensilelite/Tensile/KernelWriter.py
+++ b/tensilelite/Tensile/KernelWriter.py
@@ -1277,13 +1277,8 @@ class KernelWriter(metaclass=abc.ABCMeta):
         ####
         # scheduled pointer
         ####
-        isDTL = kernel["DirectToLdsA"] or kernel["DirectToLdsB"]
         if mfmaIndex == self.states.lwEndMfmaIndex:
           iterCode.add(pointerLWCode)
-          # Wait for previous set of global reads to complete
-          if isDTL and not NLLlast:
-            vmcntVal = 1 if kernel["PrefetchGlobalRead"] == 2 else 0
-            iterCode.add(self._wait(kernel, tPA, tPB, vmcntVal, -1, -1, "wait for previous set of global reads"))
         if i == numMfmaPerIter - 1:
           iterCode.add(pointerLRCode)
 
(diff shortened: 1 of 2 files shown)


tensilelite/Tensile/KernelWriter.py
```diff
@@ -1274,8 +1274,13 @@ def hasAnyDependency(lr: DSLoadInstruction, insts: List[Instruction]):
         ####
         # scheduled pointer
         ####
+        isDTL = kernel["DirectToLdsA"] or kernel["DirectToLdsB"]
         if mfmaIndex == self.states.lwEndMfmaIndex:
           iterCode.add(pointerLWCode)
+          # Wait for previous set of global reads to complete
+          if isDTL and not NLLlast:
+            vmcntVal = 1 if kernel["PrefetchGlobalRead"] == 2 else 0
+            iterCode.add(self._wait(kernel, tPA, tPB, vmcntVal, -1, -1, "wait for previous set of global reads"))
```
nakajee: Is this generated later than the previous location?
DTL case, global read wait should come with syncThread.
If you would like to move wait for DTL, you should move both wait and syncThread.

nakajee: It might be better to generate this wait code at syncPlrMfmaIndex.
We can skip syncThread at the previous location in DTL case.

nakajee: Sorry. syncThread code is stored in syncCode.
This is OK.

nakajee: Probably, we need this wait only for PGR2. PGR1 case, I think wait is generated at syncThread. Please confirm.

b-shi: > Is this generated later than the previous location?

Yeah, previously the wait was generated at the loop iteration boundary. This caused issues when the local writes crossed that boundary.

> Probably, we need this wait only for PGR2. PGR1 case

We still need this wait for PGR1 + DTL. I initially tried the wait with only PGR2+DTL but there were failures with PGR1 cases in the current dtl tests.

> I think wait is generated at syncThread. 

Okay, `syncThread` does generate a wait for PGR1 but there are edge cases where it does not ([example 1wave/WG](https://github.com/ROCm/hipBLASLt/blob/develop/tensilelite/Tensile/KernelWriterModules.py#L131)) which caused the failures I mentioned above.

msujon-AMD (dismissed):

tensilelite/Tensile/KernelWriter.py
nakajee: I think I figured out what was wrong with your previous change.
You do not need to move wait generation code to _makeSubIterSchedule.
Instead,  you need to put the wait code in waitLWCode.
You added wait code to module. In that case, the wait code will be generated before scheduling (which is not ideal).

nakajee: Another thing is about redundant wait code for PGR1 and num Wave > 1.
Would you please add some logic to avoid generating duplicated wait in that case?
Either removing wait in Sync for DTL or using skipForceWaitcnt0 for DTL.

b-shi: Thanks for the suggestions! :) 
Fixed.

tensilelite/Tensile/KernelWriter.py
```diff
@@ -2593,6 +2591,11 @@ def _loopBody( self, kernel, tensorParametersA, tensorParametersB, pack, lc, loo
           pointerLWCode.addComment1("local write swap offsets b")
           pointerLWCode.add(self.localWriteSwapOffsets(kernel, expand, tensorParametersB))
 
+        # DTL: Wait for previous set of global reads to complete
+        if kernel["DirectToLdsA"] or kernel["DirectToLdsB"]:
```
nakajee: I think this should be the same location as before your change.
I think we need this condition.
  if u == localWriteEndIter+1

Is there any reason you moved the code here?

b-shi: Ooops, you're right, moved to the waitcnt to the previous location. I just realized the previous waitcnt for DTL was adding to `module`.. changed to `waitLWCode` instead.

nakajee: Thanks for your confirmation.
That is what I said in my previous comment.
The current change looks good.
I do not have any further comment for now.

jichangjichang (approved):

msujon-AMD (approved):

PR #1978, opened 2025-04-24
Added CVT_SCALE_SR_FP8_F16 and CVT_SCALE_SR_BF8_F16

adityalj:
I have added support for stochastic rounding convert instructions F16 to F8 as well as BF8. F16 to F8 works perfectly when tested with given .yaml file upon setting ScaleAB=0. GlobalReadVectorWidthA =1 result in  newBlockWidth == 0.5 condition, [2,4,8] trigger else condition.  F16 to BF8 code is generated on setting datatype: b8 however upon running it throws error: Data type not implemented.  Document also has instructions for converting BF16 into F8/B8 stochastically, however our current code base does not have placeholders to add that support.


tensilelite/Tensile/KernelWriterAssembly.py
```diff
@@ -9103,30 +9103,49 @@ def localWriteBody(tP):
...
-                      localWriteCVTCode.add(VPrngB32(dst=vgpr(vRand),src=vgpr(vgprTmp),comment="Psudo Random Number Generator"))
-                    else:
-                      vTemp0 = vgprTmp+2
-                      vTemp1 = vgprTmp+3
-                      localWriteCVTCode.add(MacroInstruction(name="PRND_GENERATOR", args=[vRand, vgprTmp, vTemp0, vTemp1]))
-
+                      localWriteCVTCode.add(VPrngB32(dst=vgpr(vRand),src=paramList[0],comment="Psudo Random Number Generator"))
                     if (toF8):
-                      localWriteCVTCode.add(VCvtSRF32toFP8(dst=paramList[0], src0=vgpr(vgprTmp), src1=vgpr(vRand), vop3=VOP3PModifiers(op_sel=[0,0,sel]), comment="Convert to FP8"))
+                      if isHigh16Bits:
```
b-shi: minor: we could simplify this a bit
```python
vop3Mod = VOP3PModifiers(op_sel=([1,0,1,0] if isHigh16Bits else [0,0,0,0]))
if toF8:
  localWriteCVTCode.add(VCvtScaleSRF16toFP8(dst=paramList[0], src0=paramList[0], src1=vgpr(vRand),..., vop3Mod,...
else:
  localWriteCVTCode.add(VCvtScaleSRF16toBF8(dst=paramList[0], src0=paramList[0], src1=vgpr(vRand),..., vop3Mod,...  
```
And alias the convert instruction (see below comment)

tensilelite/Tensile/KernelWriterAssembly.py
```diff
@@ -9103,30 +9103,49 @@ def localWriteBody(tP):
...
+                    if kernel["ProblemType"]["StochasticRounding"]:
+                      vRand = vgprTmp+1 #seed
+                      if self.states.asmCaps["v_prng_b32"]:
+                        localWriteCVTCode.add(VPrngB32(dst=vgpr(vRand),src=vgpr(vgprTmp),comment="Psudo Random Number Generator"))
+                      else:
+                        vTemp0 = vgprTmp+2
+                        vTemp1 = vgprTmp+3
+                        localWriteCVTCode.add(MacroInstruction(name="PRND_GENERATOR", args=[vRand, vgprTmp, vTemp0, vTemp1]))
+
+                      if (toF8):
```
b-shi, line 9130: We can probably simplify all the `F8/BF8` and remove duplicate code by aliasing the convert instruction (early on and reuse), something like:
```Python
VCvtSRF32toX = VCvtSRF32toFP8 if toF8 else VCvtSRF32toBF8
...
  localWriteCVTCode.add(VCvtSRF32toX(dst=paramList[0], src0=vgpr(vgprTmp),... 
...
  localWriteCVTCode.add(VCvtSRF32toX(dst=paramList[0], src0=vgpr(vgprTmp),... 
```

adityalj: I agree. I have pushed these changes

b-shi: Thanks for the change, I see two more that could be simplified (i.e. `VCvtSRF32toFP8`..), but those aren't really blockers since they seem to be only used in those specific places. Up to you.

tensilelite/Tensile/Source/client/include/TypedId.hpp
```diff
@@ -226,12 +226,24 @@ namespace TensileLite
     using TypedGemm_F8H_H_H_S     = TypedGemm<Float8, Half, Half, Half, float, float, Half>;
     using TypedGemm_HF8_H_FP8_S   = TypedGemm<Half, Float8, Float8, Float8, float, float, Half>;
     using TypedGemm_F8H_H_FP8_S   = TypedGemm<Float8, Half, Float8, Float8, float, float, Half>;
+    using TypedGemm_HB8_H_S_S     = TypedGemm<Half, BFloat8, float, float, float, float, Half>;
```
msujon-AMD, line 229: If we don't have real life use-case with these GEMM cases, do we want to add so far new cases? 
Can we enable only one or very few to test the functionality of the new feature?

PR #1985, opened 2025-04-25
Add next-cmake build system

bstefanuk:
This PR introduces a new build system alongside the existing source tree, located in the next-cmake directory.

The proposed build system enhances the coherence of the source tree by organizing substantial components into dedicated directories. As the source code remains unchanged by this PR, the new build system integrates with the existing source tree using custom, temporary path variables (e.g., `TEMP_CLIENT_SOURCE_DIR`, `_CMAKE_CURRENT_SOURCE_DIR`, etc.). These variables are intended to be replaced with the correct ones once the source code migration is undertaken.

## Defined components

- host-library: Source and header files required to build libhipblaslt.so.
- device-library: Python files necessary for building tensilelite's kernels (including .co, .hsaco, and .dat files).
- rocisa: Source and header files needed to build the rocisa C++ header-only target and Python-compliant nanobind modules.
- clients: Source and header files required to build hipblaslt-bench, hipblaslt-test, and other sample executables.
- cmake: CMake helper modules.

Detailed build commands and additional resources on the new build system's functionality are documented in next-cmake/README.rst.

## Outstanding work

- Installation support
- Packaging support
- Build and installation testing to ensure install tree is a drop-in replacement for develop
- Update README
- Resolve source code changes in separate PR

This work was conducted in collaboration with @ellosel.


stellaraccident:
Thanks. I will review in detail shortly. I do want to discuss more the requirement to separately install the python packages. That shouldn't be necessary. I'll have a concrete suggestion as I proceed with the review.

next-cmake/CMakeLists.txt
```diff
@@ -0,0 +1,194 @@
...
+project(hipblaslt VERSION 0.15.0 LANGUAGES CXX)
+
+list(APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake)
+
+include(GNUInstallDirs)
+include(CMakePackageConfigHelpers)
+include(GenerateExportHeader)
+include(CMakeDependentOption)
+
+set(ENABLE_DEVICE ON CACHE BOOL "Build hipBLASLt device libraries.")
```
stellaraccident: To support composable builds, we will want to prefix all project specific options. Can do in a search/replace follow-up as we will need to pick through what is project specific. This is the only note I'll leave on this.

next-cmake/CMakeLists.txt
```diff
@@ -0,0 +1,194 @@
...
+endif()
+
+set(TEMP_TENSILE_HOST_SOURCE_DIR "${CMAKE_CURRENT_SOURCE_DIR}/../tensilelite/Tensile/Source/lib")
+set(HIPBLASLT_LIB_DIR "${CMAKE_CURRENT_SOURCE_DIR}/../library")
+set(ROCBLASLT_LIB_DIR "${HIPBLASLT_LIB_DIR}/src/amd_detail/rocblaslt")
+
+find_package(hip REQUIRED)
+if(ENABLE_HOST)
+    find_package(hipblas-common REQUIRED)
+    find_package(BLIS REQUIRED)
```
stellaraccident, line 79: For a follow-up, we need to norm this up with the usual way of finding blas libraries. Every project is doing it differently and that is causing trouble for the overall build system.

next-cmake/CMakeLists.txt
```diff
@@ -0,0 +1,194 @@
...
+endif()
+if(ENABLE_CLIENT)
+    find_package(Threads REQUIRED) # TODO: Where is this used, should it be moved up a level?
```
stellaraccident: It's generally fine to just unconditionally use this. It's basically cmake boilerplate.

next-cmake/CMakeLists.txt
stellaraccident: Example: https://github.com/ROCm/TheRock/blob/main/patches/amd-mainline/hipBLASLt/0004-Fix-finding-BLAS-and-LAPACK.patch

next-cmake/CMakeLists.txt
```diff
@@ -0,0 +1,194 @@
...
+    find_package(cblas REQUIRED)
+endif()
+if(ENABLE_OPENMP)
+    find_package(OpenMP REQUIRED)
+endif()
+if(ENABLE_MSGPACK)
+    find_package(msgpack REQUIRED NAMES msgpack msgpack-cxx msgpack-c)
```
stellaraccident: Need to use this service: https://github.com/ROCm/TheRock/blob/main/patches/amd-mainline/hipBLASLt/0001-Update-find_package-for-msgpack-to-work-with-5.x-and.patch

(Can be follow up)

next-cmake/clients/common/CMakeLists.txt
```diff
@@ -0,0 +1,25 @@
...
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+#
+# ########################################################################
+
+add_subdirectory(include)
```
stellaraccident, line 1: You can drop copyright headers for structural files that have no copyrightable content. Init.py files, cmake files that just have add_subdirectory, etc.

next-cmake/clients/common/include/CMakeLists.txt
```diff
@@ -0,0 +1,59 @@
...
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+#
+# ########################################################################
+
+set(_CMAKE_CURRENT_SOURCE_DIR "${TEMP_CLIENTS_SOURCE_DIR}/include")
+
+target_sources(hipblaslt-clients-common
+    PRIVATE
+        $<BUILD_INTERFACE:${_CMAKE_CURRENT_SOURCE_DIR}/allclose.hpp>
```
stellaraccident: I don't think the build interface stuff is buying us anything for private sources? Those are implicitly only part of the build.

stellaraccident: (here and elsewhere as this is repeated a lot)

stellaraccident (commented):
Some initial comments

next-cmake/CMakeLists.txt
```diff
@@ -0,0 +1,194 @@
...
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+#
+# ########################################################################
+
+# Need a list of old compilation flags, definitions and link flags
+# to confirm if they are required and document what we dropped if not.
+
+cmake_minimum_required(VERSION 3.22...3.25.2)
```
apwojcik: CMake 4 removes compatibility with versions of CMake older than 3.5. Do we want to limit ourselves to such an old CMake as 3.25? CMake 4 will stop here and report a fatal error.

stellaraccident: Agreed. But I think they just copied this from the existing `cmake_minimum_required(VERSION 3.22...3.25.2 FATAL_ERROR)`.

I think we should take out the upper bound. Ideally, we are compatible with all future CMake versions. And if we are not, those are bugs that need to be fixed (and having a hard cutoff will just make that really hard to fight the inertia).

next-cmake/clients/CMakeLists.txt
```diff
@@ -0,0 +1,68 @@
...
+
+add_subdirectory(common)
+add_subdirectory(bench)
+
+if(BUILD_TESTING)
+    find_package(GTest REQUIRED)
+    add_executable(hipblaslt-test)
+    target_link_libraries(hipblaslt-test PRIVATE hipblaslt-clients-common GTest::gtest rocm_smi64)
+    if(ENABLE_BLIS)
+        target_link_libraries(hipblaslt-test PRIVATE BLIS::BLIS) # necessary?
```
stellaraccident, line 62: I'm carrying a patch that has to work around this. Let's land as is but I'll need to immediately rationalize it.

next-cmake/device-library/extops/CMakeLists.txt
```diff
@@ -0,0 +1,150 @@
...
+            "${CMAKE_CURRENT_BINARY_DIR}/S_8_32_${arch}.s"
+            "${CMAKE_CURRENT_BINARY_DIR}/S_16_16_${arch}.s"
+            "${CMAKE_CURRENT_BINARY_DIR}/S_4_64_${arch}.s"
+            "${CMAKE_CURRENT_BINARY_DIR}/S_2_128_${arch}.s"
+            "${CMAKE_CURRENT_BINARY_DIR}/S_1_256_${arch}.s"
+            "${CMAKE_CURRENT_BINARY_DIR}/A_S_S_256_4_${arch}.s"
+            "${CMAKE_CURRENT_BINARY_DIR}/A_H_H_256_4_${arch}.s"
+            "${CMAKE_CURRENT_BINARY_DIR}/A_H_S_256_4_${arch}.s"
+            "${CMAKE_CURRENT_BINARY_DIR}/A_S_H_256_4_${arch}.s"
+        COMMAND ${Python3_EXECUTABLE} LayerNormGenerator.py -o "${CMAKE_CURRENT_BINARY_DIR}/L_256_4_1_${arch}.s" -w 256 -c 4 --sweep-once 1 --arch ${arch}
```
stellaraccident, line 67: Fine to do in a followup, but we really need to support a hermetic build option here that does not require separate installation of tensilelist and rocisa as Python packages.

What I propose:

Add a HIPBLASLT_BUNDLE_PYTHON_DEPS (default to ON). When ON, that would enable building of rocisa and population of the tensilelite python path. Have a cmake/hipblaslt_setup_python.cmake which you include at the top level to act on that, do the top level `find_package(Python3)` and set some project wide variables like HIPBLASLT_PYTHON_COMMAND. That variable would include the Python3_EXECUTABLE and any additional wrapper needed to setup the PYTHONPATH, enable safe python path processing, pass ASAN flags, etc.

Then project wide, instead of just invoking a loose `${Python3_EXECUTABLE}` in tools, use `${HIPBLASLT_PYTHON_COMMAND}`.

next-cmake/host-library/src/tensilelite/hip/CMakeLists.txt
```diff
@@ -0,0 +1,32 @@
...
+
+set(_CMAKE_CURRENT_SOURCE_DIR "${TEMP_TENSILE_HOST_SOURCE_DIR}/source")
+
+target_sources(hipblaslt
+    PRIVATE
+        ${_CMAKE_CURRENT_SOURCE_DIR}/hip/HipSolutionAdapter.cpp
+        ${_CMAKE_CURRENT_SOURCE_DIR}/hip/HipHardware.cpp
+)
+
+unset(_CMAKE_CURRENT_SOURCE_DIR)
```
stellaraccident: All of these are unnecessary (I know they are also just a hack to deal with the out of tree build files): `add_subdirectory()` introduces a new CMake scope so any variables defined in the target file will not leak to the thing including them (this is different than `include()`).

Here and elsewhere (only noting this one).

next-cmake/rocisa/rocisa-cpp/CMakeLists.txt
```diff
@@ -0,0 +1,41 @@
...
+target_sources(rocisa-cpp
+    INTERFACE
+        $<BUILD_INTERFACE:${rocisa_root}/rocisa/include/enum.hpp>
+)
+target_include_directories(rocisa-cpp
+    INTERFACE
+        $<BUILD_INTERFACE:${rocisa_root}>
+)
+
+unset(rocisa_root)
```
stellaraccident: Not necessary.

next-cmake/rocisa/setup.py
```diff
@@ -0,0 +1,66 @@
...
+from setuptools.command.build_ext import build_ext
+
+ROCISA_ROOT= str(Path(__file__).parent.parent.parent / "tensilelite" / "rocisa")
+class CMakeBuild(build_ext):
+    def run(self):
+        try:
+            os.makedirs(self.build_temp, exist_ok=True)
+            os.makedirs(self.build_lib, exist_ok=True)
+            source_dir = os.path.abspath(os.path.dirname(__file__))
+            rocm_path = os.environ.get('ROCM_PATH', '/opt/rocm')
```
stellaraccident: PSA: Need to burn out all `/opt/rocm` paths.

(fine to leave for now)

next-cmake/rocisa/setup.py
```diff
@@ -0,0 +1,66 @@
...
+            compilerpath = os.path.join(rocm_path, 'bin/amdclang++')
```
stellaraccident: PSA: `amdclang++` isn't a real thing. It is just a convenience symlink we provide in some builds.

(fine to leave for now)

stellaraccident (dismissed):
This looks like a good step forward and the proper layout it brings will be a big productivity boon (I waste so much time on issues related to incorrect project layering making the project very painful to develop/use).

Marking approve with conditions:

1. As mentioned: Land the C++ changes separately.
2. Please either address the comments on the review or keep track of them for followup.

Given that this is a side by side build, I'd rather see us land this big fragile part and then iterate on smaller chunks vs aiming for perfection on every detail in the first step.

Let me know when you want a final stamp to land.

next-cmake/CMakeLists.txt
bstefanuk: As far as I'm aware, the \<max\> version, here 3.25.2, identifies the latest version that the build has been tested against. It won't halt a build if the version exceeds this but will instead warn of potential compatibility issues. If we prefer to remove it altogether, I'm fine with that.

next-cmake/clients/common/include/CMakeLists.txt
bstefanuk: @stellaraccident and @ellosel I think you two should touch base about how we want to standardize this as I've been given conflicting advice.

next-cmake/CMakeLists.txt
davidd-amd: Thoughts on providing our own find module or a helper target at the monorepo level that unifies dependency management for BLAS/LAPACK?

next-cmake/CMakeLists.txt
davidd-amd: This is a simple fix. We will include it.

next-cmake/device-library/extops/CMakeLists.txt
davidd-amd: We can almost already do this without a tensilelite install but will need to update the existing tree. rocisa will have to be dealt with as you've described.

next-cmake/clients/common/include/CMakeLists.txt
davidd-amd: It's fine to drop `$<BUILD_INTERFACE:>`  from arguments with PRIVATE linkage. I was trying to be uniform because folks looking at this might not know when it is ok to drop the genex's. Just to clarify and make sure we all have the same mental model:

- `PRIVATE` usage requirement for me
- `INTERFACE` usage requirement for my consumers
- `PUBLIC` usage requirement for me and my consumers

Things that can propagate to consumers may change depending on whether or not it is used in a build or install context. Therefore, we use `$<BUILD_INTERFACE>` and/or `$<INSTALL_INTERFACE>` to handle the different contexts.

next-cmake/rocisa/setup.py
davidd-amd: I think this will get deleted.

next-cmake/CMakeLists.txt
apwojcik: It will halt a build for CMake 4.0. See the documentation: https://cmake.org/cmake/help/latest/command/cmake_minimum_required.html

> Changed in version 4.0: Compatibility with versions of CMake older than 3.5 is removed. Calls to [cmake_minimum_required(VERSION)](https://cmake.org/cmake/help/latest/command/cmake_minimum_required.html#command:cmake_minimum_required) or [cmake_policy(VERSION)](https://cmake.org/cmake/help/latest/command/cmake_policy.html#version) that do not specify at least 3.5 as their policy version (optionally via ...<max>) will produce an error in CMake 4.0 and above.

next-cmake/CMakeLists.txt
davidd-amd: Done

next-cmake/clients/common/CMakeLists.txt
davidd-amd: Done

next-cmake/clients/common/include/CMakeLists.txt
davidd-amd: Done

next-cmake/host-library/src/tensilelite/hip/CMakeLists.txt
davidd-amd: Done

next-cmake/CMakeLists.txt
davidd-amd: Commented out for now until we can prove that it isn't needed.

next-cmake/CMakeLists.txt
davidd-amd: Using 3.25 for now - I know rocroller requires 3.25 and we will want to turn that on by default in the near term.

stellaraccident: I've personally not found that the upper bound ever provides useful or accurate information. Typically projects build at the min ver and rely on cmake being generally forward compatible and don't even *know* how or whether things work at later versions -- just taking a policy to always fix things when found or reported. 

FYI - page in TheRock which attempts to explain: https://github.com/ROCm/TheRock/blob/main/docs/environment_setup_guide.md

next-cmake/clients/CMakeLists.txt
davidd-amd: Issue created https://github.com/ROCm/hipBLASLt/issues/1989

next-cmake/CMakeLists.txt
davidd-amd: Issue created https://github.com/ROCm/hipBLASLt/issues/1989

next-cmake/device-library/extops/CMakeLists.txt
davidd-amd: Issue created https://github.com/ROCm/hipBLASLt/issues/1988

stellaraccident (approved):

PR #1993, opened 2025-04-27
Hermetic build

davidd-amd:
To enable hermetic builds in the new build system, we moved the Ops scripts to a directory that doesn't require a tensilelite installation and can be invoked without a tensile module import error. A new variable was added called `HIPBLASLT_BUNDLE_PYTHON_DEPS` that is used to toggle between using rocisa bindings from the binary tree (default `ON`). If this option is `ON` the `HIPBLASLT_PYTHON_COMMAND` will be populated with the PYTHONPATH referring to the binary tree and the python 3 executable. Otherwise, the variable will be set to the python 3 executable.


next-cmake/CMakeLists.txt
```diff
@@ -102,260 +103,265 @@ if(HIPBLASLT_ENABLE_LLVM)
 endif()
 if(HIPBLASLT_ENABLE_DEVICE)
     find_package(Python3 REQUIRED COMPONENTS Interpreter)
+    if(HIPBLASLT_BUNDLE_PYTHON_DEPS)
+    set(HIPBLASLT_PYTHON_COMMAND "PYTHONPATH=${CMAKE_CURRENT_BINARY_DIR}/rocisa;${Python3_EXECUTABLE}")
```
stellaraccident: You need to do this as `${CMAKE_COMMAND} -E env ...envvar... --` for portability vs just prefixing the command with the env (posix only).

I expect also that if invoking tensile, you need to include the tensilelite source dir in the Python path. And because tensile uses the system path to find many things (and fixing that is outside the scope of this patch), you probably also need to propagate the configure time path to the build time environment (since the build time should not operate differently based on post-configure env changes). There is an example of how to do this on one of TheRock patches.

stellaraccident (dismissed):
Looks approximately right and ok to land as is. Left one comment that is probably needed for portability and from when the build shell env is different from the configure shell env. Fine to address those in a follow-up.

(Apologies for brevity: on Mobile)

next-cmake/CMakeLists.txt
davidd-amd: We should probably make `HIPBLASLT_PYTHON_COMMAND` a cache variable that can be customized as well.

stellaraccident: I've not had super great experience with that vs just teaching the build how to deal with the options. Let's see if it is needed.

PR #1995, opened 2025-04-28
Sparse, when compressing matrix at client, if cpuInput.current is not set use cpuInput.valid instead.

vin-huang:
* cpuInput.current will be set when validation is enabled.


jichangjichang (approved):

vin-huang:
The CI failed at 
1) compiling rhel was caused by compile node was disconnected.
2) sels+gfx90a testing was caused by the test script error.

those are not related to this change.

alex391a (approved):

PR #1997, opened 2025-04-28
[cmake-next] Make final tweaks needed to build as part of TheRock.

stellaraccident:
* Lets us drop all TheRock patches for hipBLASLt.
* Fixes the bundled python build mode:
  * Adds the tensilelite source directory.
  * Adds a __main__.py file so that TensileCreateLibrary can be invoked with `python -m Tensile.TensileCreateLibrary` instead of the hacks that were in place before. (this should let the bin/TensileCreateLibrary be dropped for the normal build since the console script should just work now, whereas it was pointing to a non executable package)
  * Propagates the configure time PATH to build time, making builds resilient to being invoked from a different shell environment.
  * Adds dependencies on the rocisa build to all Python invocations.
* Ports the BLAS/LAPACK finding patch to the new build system and removes hardcoded library deps on specific BLAS/LAPACK libraries.
* Fixes rocm_smi package finding (I suspect this was being tested on a system with some /opt/rocm state which was letting this work before?).
* Uses `${CMAKE_COMMAND}` instead of loose `cmake`.

With this, the hipblaslt build runs start to finish in a single ninja invocation. Null configure/build on my machine with a hot ccache is 34s for gfx1100 (56s with cold ccache -- I think more caching could easily be achieved with the build setup like this).

### DIFF
--- a/next-cmake/cmake/hipblaslt_python.cmake
+++ b/next-cmake/cmake/hipblaslt_python.cmake
@@ -1,0 +1,64 @@
+# ########################################################################
+# Copyright (C) 2025 Advanced Micro Devices, Inc.
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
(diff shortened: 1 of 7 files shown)


next-cmake/cmake/hipblaslt_python.cmake
davidd-amd, line 1: Thank for doing this.

davidd-amd (approved):

PR #2002, opened 2025-04-29
Clean up for super build

davidd-amd:
- Fixes typo in logic for including hipblsalt-test
- Removes -P in python command variable because the option requires python 3.11 or newer
- Use the hip::hipblas-common target as roc::hipblas-common was removed
- Check if hipblas-common target exists before attempting to find it


stellaraccident (approved):
Thanks!

bstefanuk (approved):

PR #2003, opened 2025-04-29
Compile kernel helpers in parallel

bstefanuk:
The kernel helpers were previously being code-generated into a monolithic file, Kernels.cpp. This update separates the kernel helpers into independent files, thereby allowing the build generator to parallelize compilation. The effect of this change scales with the number of kernel helpers, which has been seen to grow when building for bloated targets such as gfx942.

**For all targets (i.e. install.sh -c)**:

- Pipeline build time before: 2h45m
- Pipeline build time after: 1h45

### DIFF
--- a/tensilelite/Tensile/TensileCreateLibrary/Run.py
+++ b/tensilelite/Tensile/TensileCreateLibrary/Run.py
@@ -67,7 +67,7 @@ from Tensile.KernelWriterBase import (
 from Tensile.SolutionLibrary import MasterSolutionLibrary
 from Tensile.SolutionStructs import Solution
 from Tensile.Toolchain.Assembly import makeAssemblyToolchain, buildAssemblyCodeObjectFiles
-from Tensile.Toolchain.Source import makeSourceToolchain, SourceToolchain, buildSourceCodeObjectFiles, buildSourceCodeObjectFile
+from Tensile.Toolchain.Source import makeSourceToolchain, SourceToolchain, buildSourceCodeObjectFiles, buildSourceCodeObjectFilesNEW
 from Tensile.Toolchain.Validators import (
     ToolchainDefaults,
     validateToolchain,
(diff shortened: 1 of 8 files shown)


tensilelite/Tensile/Toolchain/Component.py
```diff
@@ -45,10 +45,10 @@ def _invoke(args: List[str], desc: str=""):
   """
   #print1(f"{desc}: {' '.join(args)}")
   try:
-      out = check_output(args, stderr=STDOUT)
-  except CalledProcessError as err:
+      out = check_output(args, stderr=STDOUT, cwd=working_dir)
+  except:
       raise RuntimeError(
-          f"Error with {desc}: {err.output}\n"
+          #f"Error with {desc}: {err.output}\n"
```
davidd-amd: remove

tensilelite/Tensile/TensileCreateLibrary/Run.py
```diff
@@ -390,17 +445,16 @@ def assemble(ret):
...
+    unaryWriteHelpers = functools.partial(writeHelper, outputPath)
+    srcFiles = ParallelMap2(unaryWriteHelpers,
+                            kernelHelperObjs,
+                            "Generating Kernel Helper Source",
+                            multiArg=False,
+                            return_as="list")
+    kernelsLib = str(objectTmpPath / "Kernels.so")
+
+    srcToolchain.compiler(srcFiles, kernelsLib, str(outputPath), cmdlineArchs)
+    buildSourceCodeObjectFilesNEW(srcToolchain, outputPath / "library", kernelsLib)
```
davidd-amd: Where is this?

davidd-amd (approved):

KKyang (approved):

stellaraccident:
Hi folks - we think this patch is breaking downstreams. Can we revert and the roll it forward. I hate to do that because the impact looks quite nice, but it is going to be a scramble to fix forward that would be easier to address with some changes to how this was done. I'll prepare a revert PR with notes from a cursory review and see if it gets the CI back to green, then we can land.

stellaraccident:
Revert PR: https://github.com/ROCm/hipBLASLt/pull/2028

PR #2005, opened 2025-04-29
Smaller scale for inf output

mahmoodw:
Limiting scale when output is fp16 to avoid inf and nans in validation

### DIFF
--- a/clients/include/testing_matmul.hpp
+++ b/clients/include/testing_matmul.hpp
@@ -1929,7 +1929,7 @@ void testing_matmul_with_bias(const Arguments& arg,
         if(arg.scaleA == hipblaslt_scaling_format::Scalar
            || arg.scaleA == hipblaslt_scaling_format::Vector)
         {
-            if(To==HIP_R_16F && !arg.unit_check)
+            if(!arg.unit_check)
                 hipblaslt_init_small(hScaleA[i].buf(), size_scaleAVec[i], 1, size_scaleAVec[i], Talpha);
             else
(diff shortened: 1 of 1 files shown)


AlexBrownAMD (dismissed):

jichangjichang (commented):
Actually, it should apply small init for output datatype, fp16,f8,bf8, which has small range.
To make code simpler, or we can just always use small init for scaleA/B.
" !arg.unit_check" doesn't mean that arg.norm_check is enabled

mahmoodw:
I will simplify to using smaller init irrespective of output type, we only do not want to use it when testing with unit_check since it will introduce error when checking for exact values.

clients/include/testing_matmul.hpp
```diff
@@ -1929,13 +1929,19 @@ void testing_matmul_with_bias(const Arguments& arg,
         if(arg.scaleA == hipblaslt_scaling_format::Scalar
            || arg.scaleA == hipblaslt_scaling_format::Vector)
         {
-            hipblaslt_init(hScaleA[i].buf(), size_scaleAVec[i], 1, size_scaleAVec[i], Talpha);
+            if(!arg.unit_check)
```
jichangjichang: why didn't just use "if(arg.norm_check)    hipblaslt_init_small() "?

mahmoodw: I am indifferent, the only one of concern is unit check. In rocBLAS, we have other valid options such as res_check. I'll change it.

jichangjichang (approved):

AlexBrownAMD (approved):

PR #2006, opened 2025-04-29
Increase vmcnt for DTL+PGR2 when PLR=0

b-shi:
This PR increases the `vmcnt` for the DTL+PGR2 case when PLR=0 (example, DU=128, MI=16x16x128). Previously `vmcnt(0)` is always emitted in this case. 

This is too conservative for PGR2 since we don't need to wait for the most recent set of buffer_loads (only the previous set) - also impacts performance significantly.

### DIFF
--- a/tensilelite/Tensile/KernelWriter.py
+++ b/tensilelite/Tensile/KernelWriter.py
@@ -2330,7 +2330,8 @@ class KernelWriter(metaclass=abc.ABCMeta):
       module.addComment2("Unrolled Loop %u/%u - Begin" % (lc+1, loopCopies))
     if kernel["PrefetchGlobalRead"] and not self.states.numItersPLR and not kernel["ScheduleIterAlg"] == 2:
       if kernel["DirectToLdsA"] or kernel["DirectToLdsB"]:
-        module.add(self._wait(kernel, tensorParametersA, tensorParametersB, 0, -1, -1, "11wait for global read"))
+        vmcntVal = 1 if kernel["PrefetchGlobalRead"] == 2 else 0
+        module.add(self._wait(kernel, tensorParametersA, tensorParametersB, vmcntVal, -1, -1, "11wait for global read"))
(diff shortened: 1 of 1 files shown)


msujon-AMD (dismissed):

tensilelite/Tensile/KernelWriter.py
```diff
@@ -2330,7 +2330,8 @@ def _loopBody( self, kernel, tensorParametersA, tensorParametersB, pack, lc, loo
       module.addComment2("Unrolled Loop %u/%u - Begin" % (lc+1, loopCopies))
     if kernel["PrefetchGlobalRead"] and not self.states.numItersPLR and not kernel["ScheduleIterAlg"] == 2:
       if kernel["DirectToLdsA"] or kernel["DirectToLdsB"]:
-        module.add(self._wait(kernel, tensorParametersA, tensorParametersB, 0, -1, -1, "11wait for global read"))
+        vmcntVal = 1 if kernel["PrefetchGlobalRead"] == 2 else 0
+        module.add(self._wait(kernel, tensorParametersA, tensorParametersB, vmcntVal, -1, -1, "11wait for global read"))
```
nakajee, line 2336: You can apply the same logic to NoLoadLoop (for NGLL).
Another thing is that you can skip the local write wait below for DTL (you can use "NoLdsWriteCode" for check).

I did not care much about PLR=0 case.
There might be some more chances for optimizations for PLR=0.
By the way, I think it is better to change the title (PLR=0 instead of DU=MI-K).

b-shi: Thanks for the suggestions, added! 

> There might be some more chances for optimizations for PLR=0

Right, currently addressing the low hanging fruits.. :sweat_smile:

nakajee: Thanks for your quick update.
Looks good.

msujon-AMD (approved):

KKyang (approved):

PR #2007, opened 2025-04-30
Skip clock reset if rocm-smi is not available.

stellaraccident:
* Pinning clocks was already conditioned on this, so the cleanup logic should also be.
* Also removes a windows.h `#define` that was causing trouble.


KKyang:
~~Library will work without freq, but TensileLite will not work properly if freq is not present. If freq is none then efficiency calculation is not supported. On Windows, it's better to add a message that eff is not support on this platform.~~

Edit: The freq has changed using hip python API, so the only difference is to set the GPU status to do the benchmarks.

ScottTodd (approved):

marbre (approved):

KKyang (approved):

PR #2009, opened 2025-04-30
Enabling LDS>64K for LocalReadVALU

adityalj:
Previously only MFMA Local Read was enabled and tested to use LDS>64K. However, while testing dot2_bf16 kernel we realized LocalReadVALU was not enabled to handle LDS>64K by adjusting offsets and using 2 or 3 base registers instead of just 1 as ds_read/write instruction can support maximum 64K of address offsets.

### DIFF
--- a/tensilelite/Tensile/Components/LocalRead.py
+++ b/tensilelite/Tensile/Components/LocalRead.py
@@ -92,12 +92,16 @@ class LocalReadVALU(LocalRead):
                     #     tP["bpe"], \
                     #     paramList[-1]))
                 # paramTuple = tuple(paramList)
+                num = paramList[0] //65536
+                paramList[0] = paramList[0] - num * 65536
+                srcVgpr=vgpr("LocalReadAddr%s+%d"%(tc,num))
+
                 if numOffsets == 1:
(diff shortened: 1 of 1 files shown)


tensilelite/Tensile/Components/LocalRead.py
```diff
@@ -92,12 +92,21 @@ def __call__(self, writer, kernel, bufferIdx, iui, epsi, tP):
                     #     tP["bpe"], \
                     #     paramList[-1]))
                 # paramTuple = tuple(paramList)
+                if(paramList[0] >=131072):
```
nakajee: I think you can simplify the code like this.

num = paramList[0] // 65536
srcVgpr=vgpr("LocalReadAddr%s+%d"%(tc, num))

adityalj: It's a good suggestion. I will make these changes

KKyang (approved):

msujon-AMD (approved):

nakajee (approved):
Thanks for your update.
Looks good.
No need to change in this PR, but it might be good to have a parameter to store a maximum offset for lds load/store (instead of putting 65536 at many locations)

PR #2015, opened 2025-05-01
 LDS>64K enablement for LSU

adityalj:


tensilelite/Tensile/Components/LSU.py
```diff
@@ -312,11 +314,15 @@ def writeReadReduction(self, writer, kernel):
                     for r in range(0, kernel["LocalSplitU"]):
                         regIdx = (i * numInstPerVW + v) * regsPerStore
                         offset = r * ldsStride + regIdx * (bpr * kernel["WavefrontSize"])
+                        srcvgpr = vgpr(addr)
+                        if offset>=65536:
```
msujon-AMD: can we have offset more than 128K? We need to handle that case as well then.

tensilelite/Tensile/Components/LSU.py
```diff
@@ -227,7 +227,7 @@ def writeReadReduction(self, writer, kernel):
             dataPerWave = numAccVgpr * kernel["WavefrontSize"] * 4
             ldsStride   = dataPerWave * numWaves
 
-            addr = writer.vgprPool.checkOut(1,"addr")
+            addr = writer.vgprPool.checkOut(2,"addr")
```
nakajee: It is not a good idea to always increase the number of vgpr register here.
At least, we should be able check if max LDS size is larger than 65536 or not.

Ideally, we can calculate all offset values used here in advance and check the max offset value.
Then, we can decide how many addr registers are actually needed here.

msujon-AMD (dismissed):

tensilelite/Tensile/Components/LSU.py
```diff
@@ -245,6 +242,16 @@ def writeReadReduction(self, writer, kernel):
                 DSLoadBX     = DSLoadB32
                 numInstPerVW = bytesPerVector // 4
                 regsPerStore = 1
+            maxOffset =  kernel["LocalSplitU"] * ldsStride + ((numVgprPerLSU // self.LSUfullVw) * numInstPerVW + numInstPerVW) * regsPerStore * (bpr * kernel["WavefrontSize"])
+            if  writer.states.archCaps["HasLDSGT64K"]:
+                if maxOffset >= 65536*2:
```
nakajee: I suggest to simplify the code a bit more.
We do not need if statements for offset.

numAddr = maxOffset // 65536 + 1
addr = writer.vgprPool.checkOut(numAddr ,"addr")

tensilelite/Tensile/Components/LSU.py
```diff
@@ -298,6 +305,15 @@ def writeReadReduction(self, writer, kernel):
                     comment="lsu offset = lsu_id * LSU Process Offset"))
                 module.add(VAddU32(dst=vgpr(addr), src0=vgpr(addr), src1=vgpr(tmpVgpr), \
                     comment="addr += lsu offset"))
+                if  writer.states.archCaps["HasLDSGT64K"]:
+                    if maxOffset >= 65536*2:
```
nakajee: Same here.
We can use for loop instead of if statements.

for i in range(1, numAddr):
    offset = 0x10000 * i
    module.add(VAddU32(vgpr(addr+i),hex(offset), vgpr(addr), \
                            comment="addr += %d"%offset))

tensilelite/Tensile/Components/LSU.py
```diff
@@ -312,11 +328,20 @@ def writeReadReduction(self, writer, kernel):
                     for r in range(0, kernel["LocalSplitU"]):
                         regIdx = (i * numInstPerVW + v) * regsPerStore
                         offset = r * ldsStride + regIdx * (bpr * kernel["WavefrontSize"])
+                        srcvgpr = vgpr(addr)
```
nakajee: we can simplify this like following.
srcvgpr = vgpr(addr + %d)%(offset//65536)

tensilelite/Tensile/Components/LSU.py
```diff
@@ -245,6 +242,12 @@ def writeReadReduction(self, writer, kernel):
                 DSLoadBX     = DSLoadB32
                 numInstPerVW = bytesPerVector // 4
                 regsPerStore = 1
+            maxOffset =  kernel["LocalSplitU"] * ldsStride + ((numVgprPerLSU // self.LSUfullVw) * numInstPerVW + numInstPerVW) * regsPerStore * (bpr * kernel["WavefrontSize"])
```
nakajee: I think this is not accurate.
This should be
kernel["LocalSplitU"] -> kernel["LocalSplitU"] - 1
(numVgprPerLSU // self.LSUfullVw) -> (numVgprPerLSU // self.LSUfullVw) - 1
numInstPerVW -> numInstPerVW - 1

Would you please make sure this matches the actual maximum values?

hcman2:
A minor suggestion. Can we change the HasLDSGT64K into something like Has128KLDS for more specific naming?

msujon-AMD (dismissed):

tensilelite/rocisa/rocisa/include/hardware_caps.hpp
```diff
@@ -301,7 +301,7 @@ inline std::map<std::string, bool> initArchCaps(const IsaVersion& isaVersion)
     rv["HasEccHalf"]
         = checkInList(isaVersion, {{9, 0, 6}, {9, 0, 8}, {9, 0, 10}, {9, 4, 2}, {9, 5, 0}});
     rv["Waitcnt0Disabled"] = checkInList(isaVersion, {{9, 0, 8}, {9, 0, 10}, {9, 4, 2}, {9, 5, 0}});
-    rv["HasLDSGT64K"]      = checkInList(isaVersion, {{9, 5, 0}});
+    rv["Has160KLDS"]      = checkInList(isaVersion, {{9, 5, 0}});
```
jichangjichang: are we allowed to set this new hw capability into code?

jichangjichang (commented):
can we replace "65536" as a variable for lds max constant offset?

tensilelite/rocisa/rocisa/include/hardware_caps.hpp
AlexBrownAMD: For future proofing, shouldn't we have a cap that retrieves LDS size rather than a flag that it is either equal to 64K or greater than 64K?

tensilelite/Tensile/Components/LSU.py
nakajee: I think your change is still inaccurate.

This is the code to calculate the offset
                        regIdx = (i * numInstPerVW + v) * regsPerStore
                        offset = r * ldsStride + regIdx * (bpr * kernel["WavefrontSize"])

You should replace "v" with numInstPerVW - 1.
Would you add some debug code (print or compare with max value you calculate) and make sure the max value is accurate?

tensilelite/rocisa/rocisa/include/hardware_caps.hpp
nakajee: I agree with this. 
Essentially, what we need is 
- LDS size
- maxLDS offset

We do not need Has16KLDS or HasLDSGT64K.
(HasLDSGT64K means LDS size > maxLDS offset).

Moreover, I suggest to unify LDS size and depthUConfig.maxLDS.
We should have a single source that describes LDS size.

tensilelite/Tensile/Components/LSU.py
```diff
@@ -245,6 +242,12 @@ def writeReadReduction(self, writer, kernel):
                 DSLoadBX     = DSLoadB32
                 numInstPerVW = bytesPerVector // 4
                 regsPerStore = 1
+            maxOffset =  (kernel["LocalSplitU"] -1) * ldsStride + ((numVgprPerLSU // self.LSUfullVw -1) * (numInstPerVW-1) + numInstPerVW) * regsPerStore * (bpr * kernel["WavefrontSize"])
+            numAddr =maxOffset // maxLDSConstOffset + 1
+            if  writer.states.archCaps["Has160KLDS"]:
```
hcman2: Could we just use writer.vgprPool.checkOut(numAddr,"addr") only?
numAddr can be 1 so we don't need this if-else.

tensilelite/Tensile/Components/LSU.py
```diff
@@ -245,6 +242,12 @@ def writeReadReduction(self, writer, kernel):
                 DSLoadBX     = DSLoadB32
                 numInstPerVW = bytesPerVector // 4
                 regsPerStore = 1
+            maxOffset =  (kernel["LocalSplitU"] -1) * ldsStride + ((numVgprPerLSU // self.LSUfullVw -1) * (numInstPerVW-1) + numInstPerVW) * regsPerStore * (bpr * kernel["WavefrontSize"])
```
hcman2: 2 spaces before "(kernel["LocalSplitU..."
And the next line, 0 space before "maxOffset // ...."
We have no python code format checking so the code owner needs to take care by himself.

tensilelite/Tensile/Components/LSU.py
```diff
@@ -298,6 +301,10 @@ def writeReadReduction(self, writer, kernel):
                     comment="lsu offset = lsu_id * LSU Process Offset"))
                 module.add(VAddU32(dst=vgpr(addr), src0=vgpr(addr), src1=vgpr(tmpVgpr), \
                     comment="addr += lsu offset"))
+                if  writer.states.archCaps["Has160KLDS"] and numAddr > 1:
```
hcman2: Could we remove the if here?

tensilelite/Tensile/Components/LSU.py
```diff
@@ -298,6 +301,10 @@ def writeReadReduction(self, writer, kernel):
...
+                    for i in range(1,numAddr):
+                        module.add(VAddU32(vgpr(addr+i), 0x10000*i, vgpr(addr), \
```
hcman2: Use maxLDSConstOffset or another variable name instead of 0x10000

tensilelite/Tensile/Components/LSU.py
nakajee: For this PR, I think we can remove all if statements related to Has160KLDS.

tensilelite/Tensile/Components/LSU.py
hcman2: No need to check Has160KLDS here. Use nakajee's suggestion.

tensilelite/Tensile/Components/LSU.py
```diff
@@ -245,6 +242,12 @@ def writeReadReduction(self, writer, kernel):
                 DSLoadBX     = DSLoadB32
                 numInstPerVW = bytesPerVector // 4
                 regsPerStore = 1
+            maxOffset = (kernel["LocalSplitU"] -1) * ldsStride + ((numVgprPerLSU // self.LSUfullVw -1) * (numInstPerVW-1) + numInstPerVW) * regsPerStore * (bpr * kernel["WavefrontSize"])
+            numAddr = maxOffset // maxLDSConstOffset + 1
+            if  writer.states.archCaps["Has160KLDS"]:
+               addr = writer.vgprPool.checkOut(numAddr,"addr")
+            else:
+                addr = writer.vgprPool.checkOut(1,"addr")
```
hcman2: Why we need "if writer.states.archCaps["Has160KLDS"]:" ?
numAddr can be 1 for 64K LDS case.

tensilelite/rocisa/rocisa/include/hardware_caps.hpp
adityalj: maxLDS offset depends on ISA, we can update this when new ISA supports higher offset value. We are not using Has160KLDS in this PR now. We will update it to use depthUConfig.maxLDS at other places in the repo.

nakajee: Would you revert this change?
HasLDSGT64K will be removed in #2035.
No need to make change (to avoid conflict).

tensilelite/Tensile/Components/LSU.py
```diff
@@ -312,11 +315,17 @@ def writeReadReduction(self, writer, kernel):
                     for r in range(0, kernel["LocalSplitU"]):
                         regIdx = (i * numInstPerVW + v) * regsPerStore
                         offset = r * ldsStride + regIdx * (bpr * kernel["WavefrontSize"])
+                        srcvgpr = vgpr(addr)
```
nakajee: We do not need this line.
Line 322 will overwrite this.

tensilelite/Tensile/Components/LSU.py
```diff
@@ -245,7 +242,10 @@ def writeReadReduction(self, writer, kernel):
                 DSLoadBX     = DSLoadB32
                 numInstPerVW = bytesPerVector // 4
                 regsPerStore = 1
-
+            maxOffset = (kernel["LocalSplitU"] -1) * ldsStride + ((numVgprPerLSU // self.LSUfullVw -1) * (numInstPerVW-1) + numInstPerVW) * regsPerStore * (bpr * kernel["WavefrontSize"])
```
nakajee: This is still incorrect.
Please carefully check the original calculation.
                        regIdx = (i * numInstPerVW + v) * regsPerStore
                        offset = r * ldsStride + regIdx * (bpr * kernel["WavefrontSize"])

v part is incorrect. You changed wrong location.

hcman2 (approved):
Pretty good!

nakajee (approved):
Thanks for your update.
Looks good.

jichangjichang (approved):

PR #2016, opened 2025-05-01
Separate device library build from target

bstefanuk:
Enable incremental device library builds.


next-cmake/device-library/CMakeLists.txt
```diff
@@ -42,18 +42,22 @@ set(TENSILE_CREATE_LIBRARY_COMMAND
...
-    DEPENDS ${HIPBLASLT_PYTHON_DEPS}
+    DEPENDS  ${HIPBLASLT_PYTHON_DEPS}
     # Because the command can contain special characters
     VERBATIM
     # Because this can be very long running and difficult to debug deadlocks
     # without streaming.
     USES_TERMINAL
 )
 
+add_custom_target(tensilelite-device-libraries ALL
```
stellaraccident, line 58: Thanks for fixing this - indeed in the prior work, I got too aggressive in collapsing (I was working on this due to a couple of other bugs and then was like, this can be simplified more, but was one notch too far). The way I had it was always triggering the commands since the custom target is always invalid.

stellaraccident (approved):

stellaraccident (approved):

PR #2021, opened 2025-05-03
[next-cmake] Add option for keep-build-tmp

bstefanuk:
Add a CMake option (`HIPBLASLT_DEVICE_KEEP_TEMP`), which sets `--keep-build-tmp` on tensilelite. Added as a request from developers.

Other notes:
- Update docs with the new option
- Minor printing changes
- Use `--log-level=VERBOSE` to see the python command (previously shown) and TensileCreateLibrary command (previously not printed)


next-cmake/CMakeLists.txt
```diff
@@ -60,6 +60,7 @@ endif()
 if(HIPBLASLT_ENABLE_DEVICE)
     set(HIPBLASLT_BUNDLE_PYTHON_DEPS ON CACHE BOOL "Build python dependencies requied for device code generation.")
     set(HIPBLASLT_DEVICE_JOBS "" CACHE STRING "Number of CPU threads to use for building device libraries (will use nproc if unset).")
+    set(HIPBLASLT_DEVICE_KEEP_TEMP OFF CACHE STRING "Keep temporary build directory for device libraries (turning this ON  bloat the build size).")
```
davidd-amd:
```suggestion
    set(HIPBLASLT_DEVICE_KEEP_TMP OFF CACHE STRING "Keep temporary build directory for device libraries (turning this ON  bloat the build size).")
```

next-cmake/README.rst
```diff
@@ -89,6 +108,7 @@ Options
 *Device libraries options:*
 
 * `HIPBLASLT_DEVICE_JOBS:` Allow N jobs generating device code libraries (default empty, use nproc jobs)
+* `HIPBLASLT_DEVICE_KEEP_TEMP:` Keep temporary build files (default `OFF`)
```
davidd-amd:
```suggestion
* `HIPBLASLT_DEVICE_KEEP_TMP:` Keep temporary build files (default `OFF`)
```

davidd-amd (dismissed):

next-cmake/device-library/CMakeLists.txt
```diff
@@ -55,6 +59,9 @@ add_custom_command(
     USES_TERMINAL
 )
 
+string(REPLACE ";" " " TCL "${TENSILE_CREATE_LIBRARY_COMMAND}")
```
stellaraccident: This is just to provide a nicer formatting of the immediately following message, right?

This had me looking around for where TCL might get used like this.

Because we never want this to escape, maybe surround it in a block() and name it "FORMATTED_TCL" or something so future us don't accidentally try to use it for anything else.

Also `list(JOIN` is a bit more canonical for this case.

bstefanuk: > This is just to provide a nicer formatting of the immediately following message, right?

That's right.

Great idea, thanks for the advice. I'll implement it as you've described.

bstefanuk: @stellaraccident Updated. Let me know if there are any further improvements you'd like to see.

stellaraccident (approved):

stellaraccident (approved):
Thanks. Landing with broken CI because this tree does not have CI coverage yet and must be non-impacting.

PR #2027, opened 2025-05-06
Enable DTL for non power of 2 macro tiles

b-shi:
This PR enables DTL for non power of 2 macro tiles. The previously check/restriction is no longer needed.


tensilelite/Tensile/SolutionStructs/Solution.py
```diff
@@ -834,13 +834,6 @@ def isDirectToLdsDoable(state, tc, isaInfoMap, printRejectionReason: bool):
       reject(state, printRejectionReason, "DirectToLds%c does not supports NumLoadsCoalesced%c > 1 for zgemm"%(tc, tc))
       return False
 
-    # DirectToLds does not work if MacroTile is not power of 2
```
nakajee, line 837: By the way, the reason DTL was disabled for non-power of 2 was not just about address swapping.
I think this comment was added by me long time ago.
Previously, 64bit or 128bit DTL case, we had to use some special address conversion (rotating some bits).
That logic was based on the fact that MT size was a power of 2.
This comment was intended to mention that special address conversion logic.

tensilelite/Tensile/SolutionStructs/Solution.py
```diff
@@ -2584,9 +2577,13 @@ def subCheckLdsBlockSizePerPad(tc, idx):
...
       offsetBlk = state["LdsOffsetB"] + ldsNumBytesAlignedB
 
+      mtA = state["MacroTileA"]
+      mtB = state["MacroTileB"]
+      isNonPow2MT = (mtA & (mtA - 1) != 0) or (mtB & (mtB - 1) != 0)
       state["StoreSwapAddr"] = (state["PrefetchGlobalRead"] == 2) and \
         (state["1LDSBuffer"] == 0) and \
-        (offsetBlk + int(2**(math.ceil(math.log(offsetBlk, 2)))) > depthUConfig.maxLDS)
+        (offsetBlk + int(2**(math.ceil(math.log(offsetBlk, 2)))) > depthUConfig.maxLDS) or \
+        (state["DirectToLdsA"] or state["DirectToLdsB"] and isNonPow2MT)
```
nakajee: I think you are missing () for DTLA or DTLB.
(state["DirectToLdsA"] or state["DirectToLdsB"] and isNonPow2MT)
->((state["DirectToLdsA"] or state["DirectToLdsB"]) and isNonPow2MT)

However, I think we should separately check for A and B.
For A, it does not matter if MTB is power of 2 or not.
Furthermore, it is not always true that both DTLA and DTLB are true.
(if we enable DirectToVgprA, it can be DTVA + DTLB (DTV has higher priority than DTL).

Hence, this should be like (DTLA and isNonPow2MTA) or (DTLB and isNonPow2MB).

tensilelite/Tensile/SolutionStructs/Solution.py
b-shi: > Previously, 64bit or 128bit DTL case, we had to use some special address conversion (rotating some bits).

This was due to DTL limitations for 64b and 128b loads right? I'm assuming they are not needed anymore?

tensilelite/Tensile/SolutionStructs/Solution.py
b-shi: Good points, fixed.

tensilelite/Tensile/SolutionStructs/Solution.py
nakajee: Yes. I have already removed old style address conversion code for 64b or 128b.
Then, we no longer need any special restrictions for DTL (in terms of global address calculation).

Why do we still need special treatment for DTL here?
Would you please explain a bit more detail about that?

b-shi: > Why do we still need special treatment for DTL here?

Sorry, I don't follow, which special treatment are you referring to?

nakajee: The condition you added below (for StoreSwapAddr).

b-shi: Ah.. okay I think I was confused what this comment was referring to. 
```
# LDS offset swap/rotate logic works only when MacroTile is power of 2
```
I originally thought that comment was referring to the xor offset swapping when switching between multiple LDS buffers. After re-reading your comments it seems that comment was referring to something else - so we don't need the extra checks in StoreSwapAddr. I have removed them.

nakajee: OK. Thanks for the confirmation.
Assuming all tests passed with removing the extra condition.
I do not have further comment as long as this does not cause any fails.

msujon-AMD (approved):

jichangjichang (approved):

PR #2028, opened 2025-05-06
Revert "Compile kernel helpers in parallel (#2003)"

stellaraccident:
This reverts commit 5d6a428e336b99067b4ba8bf62cfb1a28d389623.

This patch breaks multiple downstream CIs due to different issues:

* External CI: missing CPAN deps and unexpected parse errors reading the roc-obj output.
* Windows: roc-obj-* are actually perl scripts and not EXE files. They aren't very automation friendly and will need special handling to use. Also, the hard-coded shared library suffixes look like they probably won't work on Windows, which minimally, does not name its files to end in `.so`. In addition, it looks like it has a stray change to the windows name to hipinfo.

Reading through the code, I think it is a good idea but we may want to iterate on the approach as it is doing some advanced stuff with tooling that isn't necessarily hardened for it. Would be good to get some more iterations on the review.

### DIFF
--- a/tensilelite/Tensile/KernelWriterBase.py
+++ b/tensilelite/Tensile/KernelWriterBase.py
@@ -55,17 +55,17 @@ class KernelWriterBase(ABC):
 
 
   @abstractmethod
-  def getKernelName(self) -> str:
+  def getKernelName(self):
     pass
 
 
   @abstractmethod
-  def getHeaderFileString(self) -> str:
+  def getHeaderFileString(self):
     pass
 
 
   @abstractmethod
-  def getSourceFileString(self) -> str:
+  def getSourceFileString(self):
     pass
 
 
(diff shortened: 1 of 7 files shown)


stellaraccident:
Confirmed that this revert fixes the windows build.

stellaraccident:
Please also note that the roc-obj-ls and roc-obj-extract tools are deprecated. See: 

* https://github.com/ROCm/clr/blob/64d6f5714a1d4b3c792b5d734e31c935257af435/hipamd/bin/roc-obj-extract
* https://github.com/ROCm/clr/blob/64d6f5714a1d4b3c792b5d734e31c935257af435/hipamd/bin/roc-obj-ls

Note that they each print the recommended alternative that is available as part of the LLVM toolchain used to compile.

stellaraccident:
@KKyang if this passes CI during your day, would you mind landing so that we don't end up in a state that is difficult to revert? I can work with the author tomorrow on a path to land without using deprecated tooling.

jayhawk-commits (approved):

KKyang (approved):

PR #2029, opened 2025-05-06
[next-cmake] Alternate library logic path

bstefanuk:
Allow the next-cmake build system to handle alternative paths to library logic files by adding a CMake customization point (`HIPBLASLT_DEVICE_LIBLOGIC_PATH`). Added as a request from developers.


next-cmake/device-library/CMakeLists.txt
```diff
@@ -35,13 +35,16 @@ endif()
 if(HIPBLASLT_DEVICE_KEEP_TMP)
     set(TENSILE_BUILD_OPTS ${TENSILE_BUILD_OPTS} "--keep-build-tmp")
 endif()
+if(HIPBLASLT_DEVICE_LIBLOGIC_PATH STREQUAL "")
```
stellaraccident: Nit: I usually just see NOT and treating such things as truthy vs an explicit string compare.

stellaraccident (dismissed):

next-cmake/device-library/CMakeLists.txt
bstefanuk: Done.

amd-jnovotny (approved):

next-cmake/device-library/CMakeLists.txt
```diff
@@ -35,13 +35,16 @@ endif()
 if(HIPBLASLT_DEVICE_KEEP_TMP)
     set(TENSILE_BUILD_OPTS ${TENSILE_BUILD_OPTS} "--keep-build-tmp")
 endif()
+if(NOT HIPBLASLT_DEVICE_LIBLOGIC_PATH)
+    set(HIPBLASLT_DEVICE_LIBLOGIC_PATH "${CMAKE_CURRENT_SOURCE_DIR}/../../library")
+endif()
```
davidd-amd, line 40: Is there a reason to prefer this rather than defaulting the cache variable?

stellaraccident: Personally, if I was fiddling with CMake flags in a live build, like I often do, I would expect that if I set this back to empty, it should use its default vs making me need to go back through the code and find what the official location is or do a clean reconfigure. Could go either way though.

PR #2030, opened 2025-05-06
use roc::hipblas-common

davidd-amd:
Removing barriers to progress.


stellaraccident (approved):
Thanks. I think it is best to match current like this. We can update the namespace once we sync more.

PR #2031, opened 2025-05-06
Fix macro vgpr calculation in bufferStore with large loads

daineAMD:


KKyang (dismissed):

tensilelite/Tensile/KernelWriterAssembly.py
```diff
@@ -11978,8 +11978,12 @@ def bufferStoreImpl(tmpSgpr, mubuf):
             mubuf2.offen = False
             mubuf2.offset12 = 0
             module.add(SMovB32(dst=tmpSgpr, src=offset2, comment="large offset"))
-          module.add(BufferStoreB128(src=vgpr(int(srcVgpr +shiftRpv*i), shiftRpv), vaddr=addr0, \
-            saddr=addr1, soffset=tmpSgpr, mubuf=mubuf2, comment=comment))
+          if isinstance(srcVgpr, int):
+            module.add(BufferStoreB128(src=vgpr(int(srcVgpr + shiftRpv*i), shiftRpv), vaddr=addr0, \
```
nakajee: You do not need to copy the whole line to if and else.
You just need to convert srcVgpr to str if srcVgpr is int.

daineAMD: Simply changing srcVgpr to a str doesn't resolve the issue as the RegisterContainer indexing doesn't seem to work that way. I've removed the redundant if/else statement, but let me know if you had something else in mind to keep it more readable.

nakajee: Thanks for your update.
This version looks better.

KKyang (approved):

AlexBrownAMD (approved):

PR #2035, opened 2025-05-07
Fix device LDS size

AlexBrownAMD:
Fix a bug that would cause different assembly code for kernels to be generated during the tuning phase than would be generated in the final build. Issue was caused by LDS size being incorrect for build step.

This change moves LDS size out of global parameters since they are not configuration values, and they affect kernel generation.

DeviceLDS is now in ArchCaps, since it is a property of the device.

MaxLDS is moved to ForkParameters since it affects kernel generation. MaxLDS has a default value of -1, which tells the code gen to use the max amount set in DeviceLDS. MaxLDS can be set to a lower value if it is required to test out generating a kernel with reduced LDS usage for some reason.


KKyang (approved):
Good change!

davidd-amd (approved):

msujon-AMD (approved):

PR #2038, opened 2025-05-08
Fix checkLocalReadFIFO not checking metadata

KKyang:


hcman2 (dismissed):
@vin-huang please help to confirm that local read scheduling of metadata is the same as local read A/B.

aazz44ss (approved):

hcman2 (approved):

PR #2046, opened 2025-05-08
Add DTL tail loop support for 16b data types

b-shi:
This PR updates the tail loop code to use same buffer load width as main loop for DTL+16b types.

To avoid premature OOB-zeroing of values when using 32b/128b loads - alignment checks are added to handle different alignment cases.

For 4B aligned 16b arrays: add 2 bytes to the size record to allow correct direct reading of last two bytes in array.
For 2B but not 4B aligned 16b arrays: shift pointers left by 2 bytes to indirectly read last two bytes in array.

### DIFF
--- a/tensilelite/Tensile/KernelWriterAssembly.py
+++ b/tensilelite/Tensile/KernelWriterAssembly.py
@@ -58,7 +58,7 @@ from rocisa.instruction import BranchInstruction, BufferLoadB128, BufferLoadB32,
   SLShiftRightB64, SLoadB32, SLoadB64, SMFMAInstruction, SMemLoadInstruction, SMinI32, \
   SMinU32, SMovB32, SMovB64, SMulHIU32, SMulI32, SNop, SOrB32, SOrSaveExecB32, \
   SOrSaveExecB64, SSExtI16toI32, SSetPCB64, SSetRegIMM32B32, SSetPrior, SSubBU32, SSubI32, SSubU32, \
-  SWaitCnt, SWaitAlu, SXorB32, VAShiftRightI32, VAccvgprReadB32, VAccvgprWrite, VAccvgprWriteB32, \
+  SWaitCnt, SWaitAlu, SSetVSkip, SXorB32, VAShiftRightI32, VAccvgprReadB32, VAccvgprWrite, VAccvgprWriteB32, \
   VAdd3U32, VAddCCOU32, VAddCOU32, VAddF32, VAddF64, VAddLShiftLeftU32, VAddU32, VAndB32, \
   VBfeU32, VCmpEQI32, VCmpEQU32, VCmpGEI32, VCmpGEU32, VCmpGtU32, VCmpLeI32, VCmpLtI32, \
   VCmpLtU32, VCmpUF32, VCmpXGeU32, VCmpXLtU32, VCmpXLtU64, VCndMaskB32, VCvtF16toF32, \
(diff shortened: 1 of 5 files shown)


tensilelite/Tensile/KernelWriterAssembly.py
```diff
@@ -7607,6 +7690,8 @@ def globalReadGuardKBody(tP, optParams = None):
                             vgprIdx = vgprIdx + 1
                         else:
                           destVgprHi = self.vgprPool.checkOut(1, 'destVgprHi')
+                  if kernel["DirectToLds%s"%tc]:
```
nakajee, line 7693: I think you need to update numLoadVectorComp as well as numElementsPerLoad.

tensilelite/Tensile/KernelWriterAssembly.py
```diff
@@ -4125,6 +4125,89 @@ def lraSwapAddressesForDTLPad(self, kernel, tP):
...
+
+    # Skip this for 8bit types for now.
+    if kernel["ProblemType"]["DataTypeA"].isAnyFloat8() or kernel["ProblemType"]["DataTypeB"].isAnyFloat8():
+      return Module()
+
+    module.addComment0("Check if global reads need to be modified for DTL Tail Loop")
+
+    tmpSgpr = self.sgprPool.checkOut(1)
+    labelSkip = Label("SkipDTLTailLoopGRChanges", "" )
+    goodLMultiple = 2
```
nakajee, line 4153: To completely avoid out of boundary access, goodLMultiple should be GlobalReadVectorWidth.

For example,  GRVW is 4 and K % 4 is 3.
Te sfhit number will be 1 in that case.
However, GlobalRead access 4 continuous elements and it still access out of boundary elements.

tensilelite/Tensile/KernelWriterAssembly.py
```diff
@@ -4125,6 +4125,89 @@ def lraSwapAddressesForDTLPad(self, kernel, tP):
...
+      dataType = kernel["ProblemType"]["DataType%s"%tc]
+
+      if dataType.isHalf() or dataType.BFloat16():
+        tmpSgpr = self.sgprPool.checkOut(1)
+        labelEnd = Label("EndDTLTailLoopGRChanges%s"%tc, "" )
+        labelOffsetMod = Label("DTLOffsetMod%s"%tc, "" )
+
+        padValue = 2
+
+        module.add(SAndB32(dst=sgpr(tmpSgpr), src0=sgpr("Srd%s+%u"%(tc,0)), src1=hex(0x3), comment="Check if addr is 4B aligned"))
```
nakajee, line 4177: Does the alignment of Srd0 really matter?
I think lda/ldb can be unaligned number (not multiple of GlobalReadVectorWidth) and we cannot guarantee the alignment in that case.

tensilelite/Tensile/KernelWriterAssembly.py
```diff
@@ -4125,6 +4125,89 @@ def lraSwapAddressesForDTLPad(self, kernel, tP):
...
+        module.add(SCmpEQU32(src0=sgpr(tmpSgpr), src1=hex(0), comment="" )) # check if 4B aligned
+        module.add(SCBranchSCC0(labelName=labelOffsetMod.getLabelName(), comment="" ))
+
+        # Case where address is 4B aligned
+        module.add(SCmpEQU32(src0=sgpr("Srd%s+%u"%(tc,2)), src1="BufferLimit", comment="" )) # check if 4B aligned
```
nakajee, line 4182: We should not change Srd2.
That means we allow out of boundary access.

tensilelite/Tensile/KernelWriterAssembly.py
nakajee: I think what we essentially need is to avoid out of boundary access.
  mod = K % GRVW
  GlobalReadOffset -= mod * bpeGR
  LocalReadAddr += mod * bpe

I am not sure if we can use wider local read with this logic.

Another thing is that this logic is applicable only for transposed case (TLU=False).
K does not matter for TLU=True case.

tensilelite/Tensile/KernelWriterAssembly.py
nakajee: Sorry. The previous comment is not true.

nakajee: I do not think we need to check BufferLimit in TailLoop.
Srd2 should be within BufferLimit in TailLoop (since remaining K is smaller than DepthU)

nakajee: Hmm...
K < DepthU is related to TLU=True case.
TLU=False case, Srd2 should be always within BufferLimit.

tensilelite/Tensile/KernelWriterAssembly.py
b-shi: > I think lda/ldb can be unaligned number

Ah, this a good point - we need to check alignment for voff + base + scalar offset.. 
The current impl in this PR will not work.

tensilelite/Tensile/KernelWriterAssembly.py
b-shi: > That means we allow out of boundary access.

Yes, the padding is to allow potential OOB read by 2 bytes. For dword{x4} loads the OOB check work at a 32bit granularity. Given some offset a thread will read 32bits starting from that offset and if OOB occurs in between that 32bit region, the load will return zero for the entire 32bit even if part of the read is in bounds. For K odd, this will lead to some values towards the end of the matrix to not be read correctly.

tensilelite/Tensile/KernelWriterAssembly.py
nakajee: It might be better to change Srd0 and 2 instead of GlobalReadOffset.
I am not sure if GlobalReadOffset can be negative value or not.

By the way, in case GlobalReadOffset cannot be a negative value, you need to change Srd instead.
However, you cannot just change Srd0.
This is 64bit address.
You should use 64bit calculation instead.

b-shi:
@nakajee, thanks for the review/comments!

This PR was intended to be a quick fix to allow tail loop support for DTL (16bit types). But it looks like there are several issues I did not consider. Based on your [comment](https://github.com/ROCm/hipBLASLt/pull/2046#discussion_r2080748741)/observation, the alignment check logic will be more complex than I initially thought - I will investigate other options for DTL+tailloop.

PR #2048, opened 2025-05-09
Fix missing LD_PRELOAD for ExtOps

KKyang:


library/src/amd_detail/rocblaslt/src/extops/CMakeLists.txt
```diff
@@ -34,6 +34,18 @@ list(REMOVE_DUPLICATES archs)
...
+    # Must populate LD_PRELOAD with ASAN runtime if ASAN is being used.
+    # Find the ASAN RT with compiler and update env for Tensile call.
+    execute_process(
+        COMMAND ${CMAKE_CXX_COMPILER} --print-file-name=libclang_rt.asan-x86_64.so
+        OUTPUT_VARIABLE ASAN_LIB_PATH
+        COMMAND_ECHO STDOUT)
+    string(STRIP ${ASAN_LIB_PATH} ASAN_LIB_PATH)
+    set(python_launch_prefix "LD_PRELOAD=${ASAN_LIB_PATH} ${python_launch_prefix}")
+endif()
+
```
ellosel, line 48: We should put this in the file whree python_launch_prefix is set.

library/src/amd_detail/rocblaslt/src/extops/CMakeLists.txt
```diff
@@ -34,6 +34,18 @@ list(REMOVE_DUPLICATES archs)
...
+if($ENV{ENABLE_ADDRESS_SANITIZER})
+    # Must populate LD_PRELOAD with ASAN runtime if ASAN is being used.
+    # Find the ASAN RT with compiler and update env for Tensile call.
+    execute_process(
+        COMMAND ${CMAKE_CXX_COMPILER} --print-file-name=libclang_rt.asan-x86_64.so
+        OUTPUT_VARIABLE ASAN_LIB_PATH
+        COMMAND_ECHO STDOUT)
+    string(STRIP ${ASAN_LIB_PATH} ASAN_LIB_PATH)
+    set(python_launch_prefix "LD_PRELOAD=${ASAN_LIB_PATH} ${python_launch_prefix}")
```
davidd-amd: We have to be careful with how we are quoting here to be sure we don't break windows builds.

davidd-amd:
@KKyang can you explain how we enable asan with this change. We are using LD_PRELOAD when running python to code gen. Is the goal to have asan for the extop library? I would think we would want asan on when compiling/linking the generated code.

davidd-amd:
Because we are using rocisa under the hood, all python code that depends on rocisa needs the LD_PRELOAD updated accordingly.

davidd-amd (dismissed):
This build system isn't working on windows so ignore windows comments.

KKyang:
> @KKyang can you explain how we enable asan with this change. We are using LD_PRELOAD when running python to code gen. Is the goal to have asan for the extop library? I would think we would want asan on when compiling/linking the generated code.

Sorry to reply so late. We compiled a python module with C++ (rocisa) with asan enabled, so we need this LD_PRELOAD

library/src/amd_detail/rocblaslt/src/extops/CMakeLists.txt
KKyang: Fixed

library/src/amd_detail/rocblaslt/src/extops/CMakeLists.txt
```diff
@@ -34,6 +34,18 @@ list(REMOVE_DUPLICATES archs)
 set(python_venv_executable "${VIRTUALENV_BIN_DIR}/${VIRTUALENV_PYTHON_EXENAME}")
 set(python_launch_prefix "PYTHONPATH=${PROJECT_BINARY_DIR}/lib" "${python_venv_executable}")
 set(dat_depends "")
+
+if($ENV{ENABLE_ADDRESS_SANITIZER})
```
stellaraccident, line 38: I don't have a lot of investment in these build files anymore, but ftr, this is not an appropriate use of an environment variable. I realize we probably do this for compatibility with something but in the future, these things should all be cache variables.

stellaraccident (approved):
Not worried about windows with these build files: they were already hopeless. The new build files are Windows clean.

hcman2 (approved):

PR #2051, opened 2025-05-09
Remove invalid grid point.

hcman2:
Remove invalid grid point.


hcman2:
M=4784 is not in the grid set. We cannot add this kind of points.

KKyang (approved):

aazz44ss (approved):

PR #2053, opened 2025-05-09
[next-cmake] Add rocm cmake

davidd-amd:
- Restore usage of rocm-cmake.
- continue using Dependencies.cmake to pull in rocm-cmake.

### DIFF
--- a/next-cmake/CMakeLists.txt
+++ b/next-cmake/CMakeLists.txt
@@ -28,12 +28,26 @@ cmake_minimum_required(VERSION 3.25.0)
 project(hipblaslt VERSION 0.15.0 LANGUAGES CXX)
 
 list(APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake)
+list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake/rocm-cmake/share/rocmcmakebuildtools/cmake")
 
+include(ROCMInstallTargets)
+include(ROCMSetupVersion)
 include(GNUInstallDirs)
 include(CMakePackageConfigHelpers)
 include(GenerateExportHeader)
 include(CMakeDependentOption)
 include(hipblaslt_python)
+
+rocm_setup_version(VERSION ${PROJECT_VERSION})
+set(hipblaslt_SOVERSION 0.15)
+
+if(CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT)
+    if(WIN32)
+      set(CMAKE_INSTALL_PREFIX "C:/hipSDK" CACHE PATH "Install path prefix." FORCE)
+    else()
+      set(CMAKE_INSTALL_PREFIX "/opt/rocm" CACHE PATH "Install path prefix." FORCE)
+    endif()
+endif()
 
 set(HIPBLASLT_ENABLE_DEVICE ON CACHE BOOL "Build hipBLASLt device libraries.")
 set(HIPBLASLT_ENABLE_CLIENT ON CACHE BOOL "Build hipBLASLt client apps.")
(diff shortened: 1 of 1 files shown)


next-cmake/CMakeLists.txt
```diff
@@ -28,13 +28,27 @@ cmake_minimum_required(VERSION 3.25.0)
...
+rocm_setup_version(VERSION ${PROJECT_VERSION})
+set(hipblaslt_SOVERSION 0.15)
+
+if(CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT)
+    if(WIN32)
+      set(CMAKE_INSTALL_PREFIX "C:/hipSDK" CACHE PATH "Install path prefix." FORCE)
+    else()
+      set(CMAKE_INSTALL_PREFIX "/opt/rocm" CACHE PATH "Install path prefix." FORCE)
+    endif()
+endif()
```
bstefanuk, line 50: Did we agree to remove this so as to depend on the users spec

next-cmake/CMakeLists.txt
```diff
@@ -172,15 +191,17 @@ if(HIPBLASLT_ENABLE_HOST)
...
     if(HIPBLASLT_ENABLE_LAZY_LOAD)
-        target_compile_definitions(hipblaslt
-            PRIVATE
-                ROCBLASLT_TENSILE_LAZY_LOAD
-        )
+        target_compile_definitions(hipblaslt PRIVATE ROCBLASLT_TENSILE_LAZY_LOAD)
+    endif()
+
+    if(HIPBLASLT_ENABLE_SEPARATE_ARCHITECTURES)
+        target_compile_definitions(hipblaslt PRIVATE ROCBLAS_TENSILE_SEPARATE_ARCH)
```
bstefanuk: Should this be ROCBLASLT_?

next-cmake/CMakeLists.txt
```diff
@@ -190,6 +211,10 @@ if(HIPBLASLT_ENABLE_HOST)
         target_compile_definitions(hipblaslt PRIVATE TENSILE_MSGPACK)
     endif()
 
+    if(HIPBLASLT_ENABLE_MARKER)
+        target_compile_definitions(hipblaslt PRIVATE -DHIPBLASLT_ENABLE_MARKER)
```
bstefanuk: No need for the prefix -D here, target compile definitions will add it

next-cmake/CMakeLists.txt
```diff
@@ -256,150 +264,168 @@ if(HIPBLASLT_ENABLE_HOST)
...
-        "${CMAKE_CURRENT_BINARY_DIR}/hipblaslt-config.cmake"
-        DESTINATION "${CMAKE_INSTALL_LIBDIR}/cmake/hipblaslt"
-        COMPONENT hipBLASLt_Development
+endif()
+
+if(HIPBLASLT_ENABLE_DEVICE)
+    if(WIN32)
+        set(HIPBLASLT_TENSILE_LIBRARY_DIR "\${CPACK_PACKAGING_INSTALL_PREFIX}hipblaslt/bin" CACHE PATH "path to tensile library")
+    else()
+        set(HIPBLASLT_TENSILE_LIBRARY_DIR "\${CPACK_PACKAGING_INSTALL_PREFIX}${CMAKE_INSTALL_LIBDIR}/hipblaslt" CACHE PATH "path to tensile library")
```
bstefanuk, line 278: Is the $ escape necessary here? If so, why?

bstefanuk (commented):
Looks good, some comments.

next-cmake/CMakeLists.txt
davidd-amd: We agreed that we would remove it in the future.

next-cmake/CMakeLists.txt
davidd-amd: Let me check

next-cmake/CMakeLists.txt
davidd-amd: Agreed this is superstition when moving from old to new - i.e. systematically change things.

next-cmake/CMakeLists.txt
davidd-amd: This is straight from the rocm-cmake playbook. I can't say why at the moment and because I can't I didn't want to remove it.

next-cmake/CMakeLists.txt
davidd-amd: It is in fact ROCBLAS

davidd-amd: I should check that they got it right in the source code.

davidd-amd: Can't find any uses of this definition or variations thereof in the source.

stellaraccident (approved):

PR #2058, opened 2025-05-09
[next-cmake] Restore asan builds in next-cmake

davidd-amd:


next-cmake/CMakeLists.txt
```diff
@@ -119,6 +121,18 @@ if(HIPBLASLT_ENABLE_LLVM)
...
+
+    if(HIPBLASLT_ENABLE_ASAN AND NOT WIN32)
+        execute_process(
+            COMMAND ${CMAKE_CXX_COMPILER} --print-file-name=libclang_rt.asan-x86_64.so
+            OUTPUT_VARIABLE ASAN_LIB_PATH
+            COMMAND_ECHO STDOUT
+        )
+        string(STRIP ${ASAN_LIB_PATH} ASAN_LIB_PATH)
+        # Disable a few asan options to get builds going but these should be addressed
+        set(asan_opts "LD_PRELOAD=${ASAN_LIB_PATH}" "ASAN_OPTIONS=detect_leaks=0,new_delete_type_mismatch=0")
```
stellaraccident, line 159: Can you file an issue on new delete type mismatch? These are often real bugs that can lead to crashes in certain circumstances.

next-cmake/CMakeLists.txt
```diff
@@ -131,14 +145,30 @@ if(HIPBLASLT_ENABLE_DEVICE)
...
     else()
         # Just pass through to the found python executable and trust that the
         # user set it up properly.
         set(HIPBLASLT_PYTHON_COMMAND "${Python3_EXECUTABLE}")
         set(HIPBLASLT_PYTHON_DEPS)
     endif()
 endif()
+
+if(HIPBLASLT_ENABLE_ASAN)
+    add_library(hipblaslt-asan INTERFACE)
```
stellaraccident: Clever, but I'm not sure this will work right as this will leak a lot of opinionated things to users of the shared library. If XYZ target gets this as transitive usage requirements, all of the sudden, it is inheriting sanitizer build options and an override of the linker.

These are options you want to build libhipblaslt.so with, but you don't want users of it to inherit them: they have to make their own choices on those things, and there are cases where users will choose differently.

Asan is one of those annoying things that often doesn't fit right in this kind of build... Which is why you see so much ad-hoc helpers for it. I think what you actually want is not easily done with normal cmake target inheritance.

next-cmake/CMakeLists.txt
```diff
@@ -119,6 +121,18 @@ if(HIPBLASLT_ENABLE_LLVM)
 endif()
 if(HIPBLASLT_ENABLE_DEVICE)
     find_package(Python3 REQUIRED COMPONENTS Interpreter)
+
+    if(HIPBLASLT_ENABLE_ASAN AND NOT WIN32)
+        execute_process(
```
stellaraccident, line 146: FYI - if this fails, it will fail silently. Likely will not break but will silently set an empty LD_PRELOAD env var.

next-cmake/CMakeLists.txt
```diff
@@ -157,6 +172,20 @@ if(HIPBLASLT_ENABLE_HOST)
             ${CMAKE_DL_LIBS}
     )
 
+    if(HIPBLASLT_ENABLE_ASAN)
```
stellaraccident, line 201: Might want to eventually have a helper like `hipblaslt_target_configure_sanitizers(target)` or something.

stellaraccident (dismissed):
Some nits. Can be considered now or later.

next-cmake/CMakeLists.txt
davidd-amd: Adding a test to confirm that it was discovered and if not FATAL_ERROR.

PR #2059, opened 2025-05-09
[next-cmake] legacy hipblast direct

davidd-amd:
- Enables usage of hipblas api rather than hipblas-common


bstefanuk:
@ellosel for clarity, what is the context that this header is used? I haven't come across this (I don't think) in my development.

davidd-amd:
> @ellosel for clarity, what is the context that this header is used? I haven't come across this (I don't think) in my development.

This was a shim put in place when enabling hipblaslt as a backend for rocblas. I don't know why it is still in use but there are frameworks devs that still build with this option on develop.

stellaraccident (dismissed):
This is annoying. In the future, I'd like to have a look at these kinds of build variations and work to normalize. Fine for now since converging with existing practices.

PR #2061, opened 2025-05-11
Enable non-DTL loads in DTL tailloop for 16b/8b types

b-shi:
This PR enables DTL tail loop code to use non-DTL loads. 
 - Added subroutines to allocate vgprs for local write offsets and G2L for DTL tail loop.
 - Compute local write offsets to match lds write offsets used in corresponding DTL loads

### DIFF
--- a/tensilelite/Tensile/KernelWriter.py
+++ b/tensilelite/Tensile/KernelWriter.py
@@ -84,7 +84,6 @@ class ABMatrixInfo(MatrixInfo):
   numVgprValuPerBlock: int       = -1
   numVgprG2L: int                = -1
   numVgprG2LAllocated: int       = -1
-  numVgprG2LTailLoop: int        = -1
   numVgprG2LTailLoopAllocated: int= -1
   startVgprG2L: Optional[int]    = None
   numVgprLocalReadAddr:int       = -1
@@ -2984,9 +2983,8 @@ class KernelWriter(metaclass=abc.ABCMeta):
       module.add(moduleMacroG2lVgpr)
 
       # Check out VGPR for LW
-      if (kernel["DirectToLdsA"] or kernel["DirectToLdsB"]) and kernel["NonDTLTailLoop"]:
-        moduleMacroDTLLWVgpr, vgprLW = self.tailLoopAllocDTLLWVgpr(kernel)
-        module.add(moduleMacroDTLLWVgpr)
+      moduleMacroDTLLWVgpr, vgprLW = self.tailLoopAllocDTLLWVgpr(kernel)
+      module.add(moduleMacroDTLLWVgpr)
 
       module.add(self.calculateLoopNumIter(kernel, tensorParametersA, tensorParametersB, -1))
       if self.states.actualSummationLoops==1:
@@ -3017,9 +3015,17 @@ class KernelWriter(metaclass=abc.ABCMeta):
       globalReadMode1st = 3 if tensorParameters1st["isSwizzled"] else globalReadMode1st
       globalReadMode2nd = 3 if tensorParameters2nd["isSwizzled"] else globalReadMode2nd
 
-      if (kernel["DirectToLdsA"] or kernel["DirectToLdsB"]) and kernel["NonDTLTailLoop"]:
-        globalReadMode1st = 2
-        globalReadMode2nd = 2
+      if kernel["DirectToLdsA"] and kernel["NonDTLTailLoop"]:
+        if tc1 == 'A':
+          globalReadMode1st = 2
+        elif tc2 == 'A':
+          globalReadMode2nd = 2
+
+      if kernel["DirectToLdsB"] and kernel["NonDTLTailLoop"]:
+        if tc1 == 'B':
+          globalReadMode1st = 2
+        elif tc2 == 'B':
+          globalReadMode2nd = 2
 
       module.addComment1("Update M0 for DTLDS")
       moduleTmp = self.directToLdsM0Update(kernel, 1, tensorParameters1st)
@@ -3929,17 +3935,16 @@ class KernelWriter(metaclass=abc.ABCMeta):
     if not kernel["DirectToLdsA"] or self.do["KeepDirectToLdsAlloc"]:
       self.states.a.numVgprG2L = statesANumVgprG2L
       self.states.a.numVgprG2LAllocated = statesANumVgprG2LAllocated
-      self.states.a.numVgprG2LTailloop = self.states.a.numVgprG2L
       self.states.a.numVgprG2LTailloopAllocated = self.states.a.numVgprG2LAllocated
     else:
       self.states.a.numVgprG2L = 0
       self.states.a.numVgprG2LAllocated = 0
-      self.states.a.numVgprG2LTailloop = statesANumVgprG2L
       self.states.a.numVgprG2LTailloopAllocated = statesANumVgprG2LAllocated
     # using _ds_store_b8: need one more vgpr space to do lshr
     if tensorParametersA["localWriteInstruction"].blockWidth == 0.25:
       self.states.a.numVgprG2L = self.states.a.numVgprG2L * 2
-      self.states.a.numVgprG2LAllocated = self.states.a.numVgprG2LAllocated + numVgprG2LAllocatedLocal
+      self.states.a.numVgprG2LAllocated += numVgprG2LAllocatedLocal
+      self.states.a.numVgprG2LTailloopAllocated += numVgprG2LAllocatedLocal
     # double numVgprG2L if DirectToVgpr is enabled
     if kernel["DirectToVgprA"]:
       self.states.a.numVgprG2L *= 2
@@ -3970,18 +3975,16 @@ class KernelWriter(metaclass=abc.ABCMeta):
     if not kernel["DirectToLdsB"] or self.do["KeepDirectToLdsAlloc"]:
       self.states.b.numVgprG2L = statesBNumVgprG2L
       self.states.b.numVgprG2LAllocated = statesBNumVgprG2LAllocated
-      self.states.b.numVgprG2LTailloop = self.states.b.numVgprG2L
       self.states.b.numVgprG2LTailloopAllocated = self.states.b.numVgprG2LAllocated
     else:
       self.states.b.numVgprG2L = 0
       self.states.b.numVgprG2LAllocated = 0
-      self.states.b.numVgprG2LTailloop = statesBNumVgprG2L
       self.states.b.numVgprG2LTailloopAllocated = statesBNumVgprG2LAllocated
     # using _ds_store_b8: need one more vgpr space to do lshr
     if tensorParametersB["localWriteInstruction"].blockWidth == 0.25:
       self.states.b.numVgprG2L = self.states.b.numVgprG2L * 2
-      self.states.b.numVgprG2LAllocated = self.states.b.numVgprG2LAllocated + numVgprG2LAllocatedLocal
-
+      self.states.b.numVgprG2LAllocated += numVgprG2LAllocatedLocal
+      self.states.b.numVgprG2LTailloopAllocated += numVgprG2LAllocatedLocal
     # double numVgprG2L if DirectToVgpr is enabled
     if kernel["DirectToVgprB"]:
       self.states.b.numVgprG2L *= 2
--- a/tensilelite/Tensile/KernelWriter.py
+++ b/tensilelite/Tensile/KernelWriter.py
@@ -3969,7 +3969,7 @@ class KernelWriter(metaclass=abc.ABCMeta):
     if (self.states.archCaps["HasEccHalf"] or not self.states.asmCaps["HasWMMA_V1"]) and (bpeMax * vwb < self.states.bpr):
       # This check is to reserve porential usage of VGPRs for gfx12 8-bit code gen
       # We should optimize the usage for better performance.
-      statesBNumVgprG2LAllocated = statesBNumVgprG2LAllocated * (int)(self.states.bpr/(bpeMax * vwb))
+      statesBNumVgprG2LAllocated = statesBNumVgprG2L * (int)(self.states.bpr/(bpeMax * vwb))
     else:
       statesBNumVgprG2LAllocated = statesBNumVgprG2L
     if not kernel["DirectToLdsB"] or self.do["KeepDirectToLdsAlloc"]:
--- a/tensilelite/Tensile/KernelWriter.py
+++ b/tensilelite/Tensile/KernelWriter.py
@@ -3015,13 +3015,13 @@ class KernelWriter(metaclass=abc.ABCMeta):
       globalReadMode1st = 3 if tensorParameters1st["isSwizzled"] else globalReadMode1st
       globalReadMode2nd = 3 if tensorParameters2nd["isSwizzled"] else globalReadMode2nd
 
-      if kernel["DirectToLdsA"] and kernel["NonDTLTailLoop"]:
+      if kernel["DirectToLdsA"] and kernel["NonDTLTailLoopA"]:
         if tc1 == 'A':
           globalReadMode1st = 2
         elif tc2 == 'A':
           globalReadMode2nd = 2
 
-      if kernel["DirectToLdsB"] and kernel["NonDTLTailLoop"]:
+      if kernel["DirectToLdsB"] and kernel["NonDTLTailLoopB"]:
         if tc1 == 'B':
           globalReadMode1st = 2
         elif tc2 == 'B':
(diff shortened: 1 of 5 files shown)


tensilelite/Tensile/KernelWriterAssembly.py
```diff
@@ -7585,9 +7627,11 @@ def globalReadGuardKBody(tP, optParams = None):
                       eccOffset = _getEccOffset(tP["globalReadInstruction"].totalWidth, bpr=self.states.bpr, bpe=eccBpe, \
                         glvw=tP["glvw"], idx=loopCnt, numVgprG2L=numVgprG2L)
                 elif dataType.isHalf() or dataType.isBFloat16():
-                  if tP["glvw"]>1 and kernel["AssertSummationElementMultiple"] % 2 == 0:
+                  if tP["glvw"]>1 and kernel["AssertSummationElementMultiple"] % 2 == 0 and not isLds:
                   # Pack two FP16 values into a single load dword x2
                     numElementsPerLoad = 2
+                  elif isLds and kernel["AssertSummationElementMultiple"] % 2 == 0:
+                    numElementsPerLoad = kernel["GlobalReadVectorWidth%c"%tc]
```
nakajee, line 7635: Can we really do this?
Doesn't this cause out of boundary access?
For example, if ASEM=2, GRVW is 4 and K is not multiple of 4, I think this does out of boundary access since each global load accesses 4 elements at once.

tensilelite/Tensile/KernelWriterAssembly.py
```diff
@@ -4542,11 +4564,15 @@ def tailLoopAllocG2LVgpr(self, kernel):
     if not kernel["DirectToVgprA"]:
       if ("ULSGRODoubleG2L" in kernel) and kernel["ULSGRODoubleG2L"] == 1:
         numG2LA = self.states.a.numVgprG2LAllocated*2
+      elif kernel["DirectToLdsA"] and kernel["NonDTLTailLoop"]:
```
nakajee: I think this if statement is redundant.
You should be able to use numVgprG2LTailloopAllocated for any cases.

tensilelite/Tensile/KernelWriter.py
```diff
@@ -2992,6 +3000,10 @@ def kernelBody( self, kernel, tensorParametersA, tensorParametersB ):
       globalReadMode1st = 3 if tensorParameters1st["isSwizzled"] else globalReadMode1st
       globalReadMode2nd = 3 if tensorParameters2nd["isSwizzled"] else globalReadMode2nd
 
+      if (kernel["DirectToLdsA"] or kernel["DirectToLdsB"]) and kernel["NonDTLTailLoop"]:
```
nakajee: Shouldn't we check A and B separately?

tensilelite/Tensile/KernelWriterAssembly.py
```diff
@@ -9340,7 +9385,7 @@ def localWriteBody(tP):
         localWriteCode.add(SBarrier(comment="dump LDS"))
         localWriteCode.add(self.getCmpAssert(self.asmAssert.ne, sgpr("WorkGroup0"),1))
 
-    if (not kernel["DirectToLds%s"%tc]):
+    if (not kernel["DirectToLds%s"%tc]) or (kernel["NonDTLTailLoop"] and self.states.inTailLoop):
```
nakajee: Shouldn't we avoid applying this logic for Metadata?

tensilelite/Tensile/KernelWriter.py
```diff
@@ -3874,23 +3895,32 @@ def readWriteVectors(mat, vw, kernel):
...
           tpA) / (float)(self.states.bpr))
-      else:
-        self.states.a.numVgprG2LAllocated = self.states.a.numVgprG2L
     else:
+      statesANumVgprG2LAllocated = self.states.a.numVgprG2L
+
+    if not kernel["DirectToLdsA"] or self.do["KeepDirectToLdsAlloc"]:
+      self.states.a.numVgprG2L = statesANumVgprG2L
+      self.states.a.numVgprG2LAllocated = statesANumVgprG2LAllocated
+      self.states.a.numVgprG2LTailloop = self.states.a.numVgprG2L
```
nakajee: Do we need numVgprG2LTailloop?
Looks like not used anywhere.

tensilelite/Tensile/KernelWriterAssembly.py
b-shi: I think it should be okay since the size record in the buffer descriptor will prevent the OOB access.

tensilelite/Tensile/KernelWriterAssembly.py
b-shi: Good point, I've switched to use `numVgprG2LTailloopAllocated` for all A/B cases in `tailLoopAllocG2LVgpr`.

tensilelite/Tensile/KernelWriterAssembly.py
nakajee: Are you sure?
That is different from my understanding.
Let's say, address=0x1000, OOB is 0x1004, GRVW=4 (fp16, 8byte access per thread),
in this case, global read instruction accesses from 0x1000 to 0x1007.
My understanding is buffer load limit (Srd+2) does not limit the OOB access if the limit is within global load unit.

If OOB access limit really works in this case, I think all edge related logics (Edge shift, narrower load for TailLoop) are unnecessary.

tensilelite/Tensile/KernelWriter.py
```diff
@@ -3971,23 +4011,29 @@ def readWriteVectors(mat, vw, kernel):
...
+    self.states.a.numVgprLocalWriteAddrTailLoop = 0 if not (kernel["DirectToLdsA"] and kernel["NonDTLTailLoop"]) else 1 * self.states.rpla
+    self.states.b.numVgprLocalWriteAddrTailLoop = 0 if not (kernel["DirectToLdsB"] and kernel["NonDTLTailLoop"]) else 1 * self.states.rpla
 
     if self.states.archCaps["DeviceLDS"] > 65536 and not kernel["StoreSwapAddr"]:
       need128K = kernel["LdsOffsetA_Blk"]>=131072 and kernel["ExpandPointerSwap"] and not kernel["1LDSBuffer"]
       need64K = kernel["LdsOffsetA_Blk"]>=65536 and kernel["ExpandPointerSwap"] and not kernel["1LDSBuffer"]
       if need128K or kernel["LdsNumElementsAlignedA"]>=131072:
         self.states.a.numVgprLocalReadAddr *= 3
         self.states.a.numVgprLocalWriteAddr *= 3
+        self.states.a.numVgprLocalWriteAddrTailLoop *= 3
```
nakajee, line 4038: Not necessary to do this in this PR, but I suggest to refactor this part.
- replace 65536 related const values with offset limit
- calculate num of multiple factor (*3 or *2) before setting numVgprLocalRead/Write addr. Then, use the num of multiple factor when setting numVgprLocalRead/Write addr (replace 1* with num *)

This part looks redundant.
I suggest to simplify it instead of adding *3 and *2 for new parameters.

tensilelite/Tensile/Tests/common/gemm/gfx950/dtl.yaml
```diff
@@ -84,12 +84,12 @@ BenchmarkProblems:
         - SourceSwap: [1]#[0, 1]
         - LocalReadVectorWidth: [8]
         - DirectToLds: [1]
-        - AssertSummationElementMultiple: [4]
+        - AssertSummationElementMultiple: [1, 4]
```
nakajee, line 87: Do we have any test case with GSU?
I would like to know if this logic really works with GSU.
Let's say, if ASEM=4, K%DepthU=4, GSU=2, in this case, the actual K size assigned to each CU is K/GSU.
Then, K/GSU is not a multiple of 4.

I am wondering if ASEM really works as expected with GSU.

tensilelite/Tensile/KernelWriterAssembly.py
b-shi: Let me double check and get back to you. I believe all dword/dwordx2/dwordx4 have read granularity of 32 bits and in the example you provided the vgprs containing elements in 0x1004-0x1007 would be zero.

nakajee: OK. Please dump the register value and make sure if it is 0 or not.
By the way, setting value 0 to OOB does not work with dtl option.

b-shi: Confirmed, size record determines OOB cut-off with 32b granularity (for dword{x2,x4}). 

Comment: the 32b granularity makes directly reading 2 bytes not possible with dword{x2,x4} loads - i.e. if OOB is 0x1002, load will return 0 when trying to read 2 bytes from 0x1000. My previous [PR](https://github.com/ROCm/hipBLASLt/pull/2046) was trying to workaround this. 

> setting value 0 to OOB does not work with dtl option.

What do you mean by this?

nakajee: OK. Thanks for confirmation.
I did not know that.

Regarding this
> setting value 0 to OOB does not work with dtl option.

Without dtl option, global load value in register from OOB memory will be 0.
However, dtl case, looks like 0 is not written to LDS (if I remember correctly).

tensilelite/Tensile/Tests/common/gemm/gfx950/dtl.yaml
b-shi: I quickly checked some cases with `GSU > 1` and it looks like there are some predicates that enforces K so that `K / GSU` satisfies the ASEM condition. :thinking:

nakajee: OK. Thanks for confirmation.
Interesting.
Please add some GSU test cases.

tensilelite/Tensile/SolutionStructs/Solution.py
```diff
@@ -371,6 +371,14 @@ def assignProblemIndependentDerivedParameters(state, printRejectionReason: bool,
     state["tailLoopOptA"] = True
     state["tailLoopOptB"] = True
 
+    # Use nonDTL loads in DTL tail loop
+    state["NonDTLTailLoop"] = False
+    if (((state["ProblemType"]["DataType"].isHalf() or state["ProblemType"]["DataType"].isBFloat16()) and \
+        state["AssertSummationElementMultiple"] % 2 != 0) or \
```
nakajee: I suggest to check if TLU is false or not here.
Looks like DTL is not enabled in TLU=True case, but ASEM matters only if TLU=False.
It is better to add necessary check in case we enable DTL for TLU=True in the future.

b-shi: That makes sense, I've split `NonDTLTailLoop` for A/B separately and added the TLU checks.

tensilelite/Tensile/SolutionStructs/Solution.py
```diff
@@ -371,6 +371,17 @@ def assignProblemIndependentDerivedParameters(state, printRejectionReason: bool,
     state["tailLoopOptA"] = True
     state["tailLoopOptB"] = True
 
+    # Use nonDTL loads in DTL tail loop
+    state["NonDTLTailLoopA"] = False
+    state["NonDTLTailLoopB"] = False
+    is16bASEM2 = ((state["ProblemType"]["DataType"].isHalf() or state["ProblemType"]["DataType"].isBFloat16()) and \
```
nakajee: I think this should be DataTypeA/B or bpeGR.
"DataType" is for input type of MFMA.
If we need input data type conversion, DataType will be after conversion type.
For global load, we should use input type for before conversion.
I suggest to add some test cases for input type conversion (fp16->fp8, fp8->fp16) to make sure this logic works with type conversion cases.

Another thing is that what we essentially need here is data size.
We can simplify the code by using dataSize (like numByte or bpeGR).

What we need to check is something like this.
   bpeGR * ASEM is multiple of 4
We do not need to check fp16,bf16,int8,float8 separately.

b-shi: > Another thing is that what we essentially need here is data size.

Very good point, I have simplified the previous logic.

Regarding the mixed precision test cases. It looks like there are some general functional issues (not tailloop related) so I have currently added a rejection condition for those cases. This needs to be fixed and is on my todo list.

nakajee: I think you mixed up the code change here.
(I suggest to do self review before pushing the code change...)

We should check A and B separately.
Another thing is you do not need to check 1 byte and 2 bytes separately.
As I said in my previous comment, you can just do like this.
  if (DataTypeA.numBytes * ASEM) % 4 == 0):
    NonDTLTailLoopA = ...
  if (DataTypeB.numBytes * ASEM) % 4 == 0):
    NonDTLTailLoopB = ...

nakajee: Please put some more comments for the explanation about the logic.

b-shi: Thanks, comment added, NonDTLTailLoop logic condition fixed and simplified.

tensilelite/Tensile/KernelWriter.py
```diff
@@ -3057,6 +3081,9 @@ def kernelBody( self, kernel, tensorParametersA, tensorParametersB ):
       # tail: free G2L Vgpr
       module.add(self.tailLoopFreeVgpr(vgprG2L, moduleMacroG2lVgpr))
 
+      if (kernel["DirectToLdsA"] and kernel["NonDTLTailLoopA"]) or (kernel["DirectToLdsB"] and kernel["NonDTLTailLoopB"]):
```
nakajee: We do not need this if condition.
Checking vgprLW!=-1 should be enought (handled in tailLoopFreeVgpr())

tensilelite/Tensile/SolutionStructs/Solution.py
```diff
@@ -2321,6 +2334,16 @@ def calSwizzleK(state, tc):
     # LDS (load size coalesced) * LSPA must load some multiple of 256 bytes.
     # No longer support loadX2/loadx4 .
     if state["DirectToLds"]:
+
+      bpeA = state["ProblemType"]["DataTypeA"].numBytes()
+      bpeB = state["ProblemType"]["DataTypeB"].numBytes()
+
+      # TODO: Currently DTL with input types of different size is not support. There are functional issues
+      # This needs to be fixed.
+      if bpeA != bpeB:
```
nakajee: I think it is better to check this after isDirectToLdsDoable and check A and B separately.
DirectToLdsA/B can be false even if DirectToLds is true.
The condition should be like
  if bpeAbefore != bpeAafter:

aazz44ss:
Have you try setting tailLoopOpt to false in Solution.py?
state["tailLoopOptA"] = True

b-shi:
> Have you try setting tailLoopOpt to false in Solution.py? state["tailLoopOptA"] = True

I just checked, setting `state["tailLoopOptX"] = False` will still generate DTL loads in the tail loop code.

tensilelite/Tensile/SolutionStructs/Solution.py
b-shi: > if bpeAbefore != bpeAafter:

What does before and after refer to here?

nakajee: Before conversion and after conversion.
What I thought was that like
- NN case does not use DTL for A (because TLU=True)
- then, we can use type conversion for A (like f8 to f16)
- then, bpeA != bpeB can be OK

tensilelite/Tensile/KernelWriter.py
```diff
@@ -3030,21 +3048,27 @@ def kernelBody( self, kernel, tensorParametersA, tensorParametersB ):
...
-      if not kernel["NoLdsWriteCode"]:
+      if not kernel["NoLdsWriteCode"] or kernel["NonDTLTailLoopA"] or kernel["NonDTLTailLoopB"]:
         # tail: local write
-        module.addComment1("local write a")
-        tempLWCodeModA = self.localWriteDo(kernel, tensorParametersA)
-        module.add(tempLWCodeModA)
-        module.addComment1("local write b")
-        tempLWCodeModB = self.localWriteDo(kernel, tensorParametersB)
-        module.add(tempLWCodeModB)
+        if not (kernel["DirectToLdsA"] and not kernel["NonDTLTailLoopA"]):
```
nakajee: I think this if statement is also redundant.
You put if condition in localWriteDo as well.
We need only one of them.

tensilelite/Tensile/SolutionStructs/Solution.py
b-shi: Good point regarding NN case, I've moved the check back to `isDirectToLdsDoable` and added test cases for mixed precision NN, TT cases.

nakajee (approved):
Looks good.
I do not have further comments.

aazz44ss (approved):

msujon-AMD (approved):

PR #2064, opened 2025-05-12
[Feature] Generate coverage report in hipblaslt-library and tensilelite

tomchengchitang:
# Update 05-22-2025
After the build, you can just run `make coverage` to generate both hipblaslt-library and tensilelite reports.

# hipblaslt-library
- Build: `./install.sh -dc -g --codecoverage` 
- Generate the report:
  - Run programs you want. e.g.: `LLVM_PROFILE_VERBOSE=1 LLVM_PROFILE_FILE=./coverage_%m.profraw ./clients/staging/sample_hipblaslt_gemm_tuning_wgm_ext --output-on-failure`
  - Merge all .profraw files to coverage_test.profdata: `llvm-profdata merge -sparse *.profraw -o coverage_test.profdata`
  -  Show the report on terminal and filter out other code except library: `llvm-cov report -instr-profile=./coverage_test.profdata ./clients/staging/sample_hipblaslt_gemm_tuning_wgm_ext  --ignore-filename-regex=".*/samples/.*|.*/build/debug/include/.*|.*/build/debug/virtualenv/.*"`
  - The report is the first attached
# tensilelite
- Go to tensiltelite
 - Run the test: `tox -v --workdir /tmp/.tensile-tox -e py3 -- Tensile/Tests -m common --cov=. --cov-report=html`
 - The report is the 2nd attached
 
 
image
image


CMakeLists.txt
```diff
@@ -121,8 +121,9 @@ cmake_dependent_option(HIPBLASLT_ENABLE_MARKER "Enable roctx marker in hipBLASLt
 option(USE_ROCROLLER "Build with the rocRoller library" ON)
 
 if(BUILD_CODE_COVERAGE)
```
reboss: Would it be possible for you to add a new target called `coverage`?  We're currently in the process of setting standards for building and testing so each component builds/tests in the same way.

With code coverage, we'd like to be able to simply run the following to get a code coverage report:
```bash
cmake ${buildArgs} -DBUILD_CODE_COVERAGE ${projectDir}
make -j coverage
```

See how rocFFT is doing it here: https://github.com/ROCm/rocFFT/pull/551

tensilelite/tox.ini
```diff
@@ -22,8 +22,9 @@ commands =
     sh -c "cmake -S rocisa -B {envdir}/rocisa -DCMAKE_CXX_COMPILER={env:ROCM_PATH:/opt/rocm}/bin/amdclang++"
     sh -c "cmake --build {envdir}/rocisa -j  $(nproc)"
     pip install --upgrade pip
+    pip install pytest-cov
     python3 ./Tensile/bin/Tensile Tensile/Tests/build_client.yaml {envdir}/client
-    py.test -v --basetemp={envtmpdir} --junit-xml={toxinidir}/python_tests.xml --junit-prefix={envname} --color=yes -n 4 --prebuilt-client={envdir}/client/0_Build/client/tensile_client {posargs}
+    pytest -v --basetemp={envtmpdir} --junit-xml={toxinidir}/python_tests.xml --junit-prefix={envname} --color=yes -n 4 --prebuilt-client={envdir}/client/0_Build/client/tensile_client {posargs}
```
reboss, line 27: Regarding my comment above, I think we'd need to have a target to run these tests defined in CMakeLists.txt.  Is there anything that may prevent that or make it unfeasible to do?

CMakeLists.txt
tomchengchitang: Sure, let me survey rocFFT...

tomchengchitang: Created the script and building...

tomchengchitang: I was resolving some linkage issue and it has been resolved. (The building time is a bit long) Now it enters the final stage, I will push a new commit after the final test pass. Also, is it okay to call the target **all_coverage_reports**? Or is **coverage** better? Thanks.

reboss: To keep it consistent with other libraries, we'd appreciate it if you could just go with `coverage`

CMakeLists.txt
```diff
@@ -121,8 +121,29 @@ cmake_dependent_option(HIPBLASLT_ENABLE_MARKER "Enable roctx marker in hipBLASLt
...
-  add_compile_options(-fprofile-arcs -ftest-coverage)
-  add_link_options(--coverage -lgcov)
+  set(BUILD_SHARED_LIBS OFF CACHE BOOL "Build coverage reports" FORCE)
+  set(LLVM_DIR /opt/rocm/llvm/lib/cmake/llvm)
+  find_package(LLVM REQUIRED CONFIG)
```
reboss: A better approach would be to do the following:
```
  find_program(
    LLVM_PROFDATA
    llvm-profdata
    REQUIRED
    HINTS ${rocm_path}/llvm/bin
    PATHS ${rocm_path}/llvm/bin
  )

  find_program(
    LLVM_COV
    llvm-cov
    REQUIRED
    HINTS ${rocm_path}/llvm/bin
    PATHS ${rocm_path}/llvm/bin
  )
```

See comment below.

**Note:** `${rocm_bin}` resolves to /opt/rocm/bin and llvm is at the same level as bin, so we need to define `${rocm_path}` separately.

CMakeLists.txt
```diff
@@ -121,8 +121,29 @@ cmake_dependent_option(HIPBLASLT_ENABLE_MARKER "Enable roctx marker in hipBLASLt
...
+  add_link_options(-fprofile-instr-generate)
+
+  add_custom_target(coverage
+    COMMAND LLVM_PROFILE_VERBOSE=1 LLVM_PROFILE_FILE=./coverage_%m.profraw ./clients/staging/hipblaslt-test --output-on-failure
+    COMMAND LLVM_PROFILE_VERBOSE=1 LLVM_PROFILE_FILE=./coverage_%m.profraw ./clients/staging/sample_hipblaslt_ext_op_amax --output-on-failure
+    COMMAND LLVM_PROFILE_VERBOSE=1 LLVM_PROFILE_FILE=./coverage_%m.profraw ./clients/staging/sample_hipblaslt_ext_op_amax_with_scale --output-on-failure
+    COMMAND LLVM_PROFILE_VERBOSE=1 LLVM_PROFILE_FILE=./coverage_%m.profraw ./clients/staging/sample_hipblaslt_ext_op_layernorm --output-on-failure
+    COMMAND LLVM_PROFILE_VERBOSE=1 LLVM_PROFILE_FILE=./coverage_%m.profraw ./clients/staging/sample_hipblaslt_weight_swizzle_padding --output-on-failure
+    COMMAND sh -c "ls *.profraw 1>/dev/null 2>&1 && llvm-profdata merge -sparse *.profraw -o coverage_test.profdata || echo 'No .profraw files found'"
+    COMMAND sh -c "export PATH=/usr/lib/llvm-19/bin:\$PATH && llvm-cov report -instr-profile=./coverage_test.profdata ./clients/staging/sample_hipblaslt_ext_op_amax --ignore-filename-regex='.*/samples/.*|.*/build/debug/.*'"
```
reboss: We need to use the `llvm-profdata` and `llvm-cov` that matches the `clang` version we're using.  Since by default, we set the `CMAKE_CXX_COMPILER` to /opt/rocm/bin/amdclang, we need to use the llvm tools that are in /opt/rocm/llvm.

Also, if using the above `find_program` function, `${LLVM_PROFDATA}` and `${LLVM_COV}` will be set so we can use them here instead of the path inlined with the command

CMakeLists.txt
```diff
@@ -121,8 +121,29 @@ cmake_dependent_option(HIPBLASLT_ENABLE_MARKER "Enable roctx marker in hipBLASLt
...
+    COMMAND sh -c "export PATH=/usr/lib/llvm-19/bin:\$PATH && llvm-cov show ./clients/staging/sample_hipblaslt_ext_op_amax -instr-profile=./coverage_test.profdata -format=html -output-dir=coverage_report --ignore-filename-regex='.*/samples/.*|.*/build/debug/.*'"
```
reboss: I think your `llvm-cov` command needs to be:
 ```bash
llvm-cov show \
    -object ./clients/staging/sample_hipblaslt_ext_op_amax \
    -object ./clients/staging/hipblaslt-test \
    -object ./clients/staging/sample_hipblaslt_ext_op_amax_with_scale \
    -object ./clients/staging/sample_hipblaslt_ext_op_layernorm \
    -object ./clients/staging/sample_hipblaslt_weight_swizzle_padding \
    -instr-profile=./coverage_test.profdata \
    -format=html \
    -output-dir=coverage_report \
    --ignore-filename-regex='.*/samples/.*|.*/build/debug/.*'"
```

It's really annoying that `llvm-cov` doesn't allow for wildcards in the `-object` parameter.  You might be able to get away with what you have if all test executables import exactly the same header files, but I haven't tested that scenario yet.

If your coverage report fluctuates as a result of changing the number of `-object` parameters you pass, then you'll need them all.

File: CMakeLists.txt (hunk 3 above)
Comment (evetsso) [reply CMakeLists.txt]: FYI CMake convention is for `PATHS` to be hardcoded guesses (e.g. `/opt/rocm/llvm/bin`) while `HINTS` should be used for computed values.  https://cmake.org/cmake/help/latest/command/find_program.html

File: CMakeLists.txt [hunk 6]
```diff
@@ -121,8 +121,29 @@ cmake_dependent_option(HIPBLASLT_ENABLE_MARKER "Enable roctx marker in hipBLASLt
...
 if(BUILD_CODE_COVERAGE)
-  add_compile_options(-fprofile-arcs -ftest-coverage)
-  add_link_options(--coverage -lgcov)
+  set(BUILD_SHARED_LIBS OFF CACHE BOOL "Build coverage reports" FORCE)
+  set(LLVM_DIR /opt/rocm/llvm/lib/cmake/llvm)
+  find_package(LLVM REQUIRED CONFIG)
+  message(STATUS "LLVM found: ${LLVM_DIR}")
+  include_directories(${LLVM_INCLUDE_DIRS})
+  link_directories(${LLVM_LIBRARY_DIRS})
+  add_compile_options(-fprofile-instr-generate -fcoverage-mapping)
```
evetsso: Consider adding target-specific compile/link options (`target_compile_options` and `target_link_options`) instead of adding directory-wide options.  Besides having clearer scoping, this would let you limit coverage options to the hipBLASLt library itself and not any of the client programs.  You'd want the compile options to be private to hipBLASLt, and the link option to be public.

CMakeLists.txt
```diff
@@ -121,8 +121,29 @@ cmake_dependent_option(HIPBLASLT_ENABLE_MARKER "Enable roctx marker in hipBLASLt
...
+  add_link_options(-fprofile-instr-generate)
+
+  add_custom_target(coverage
+    COMMAND LLVM_PROFILE_VERBOSE=1 LLVM_PROFILE_FILE=./coverage_%m.profraw ./clients/staging/hipblaslt-test --output-on-failure
```
evetsso: The portable way to set environment variables when running programs is `${CMAKE_COMMAND} -E env`, so for example:
```cmake
COMMAND ${CMAKE_COMMAND} -E env LLVM_PROFILE_VERBOSE=1 LLVM_PROFILE_FILE=./coverage_%m.profraw ./clients/staging/hipblaslt-test --output-on-failure
```

`ls` is not portable.  I would advocate just letting `llvm-profdata` run and let it print an error on its own if no files were found.

`sh` is not portable either.  But you also shouldn't need it if you find `llvm-cov` as @reboss suggests above.

This custom target also needs dependencies.  For example, it should depend on targets like `hipblaslt-test`, `sample_hipblaslt_ext_op_amax`, and so on.  The build system needs to be told enough such that right after running CMake you can just do `make coverage` and get a coverage report out.

install.sh
```diff
@@ -739,7 +739,7 @@ pushd .
           echo "Code coverage is disabled in Release mode, to enable code coverage select either Debug mode (-g | --debug) or RelWithDebInfo mode (-k | --relwithdebinfo); aborting";
           exit 1
       fi
-      cmake_common_options="${cmake_common_options} -DBUILD_CODE_COVERAGE=ON"
+      cmake_common_options="${cmake_common_options} -DBUILD_CODE_COVERAGE=ON -DLLVM_DIR=/opt/rocm/llvm/lib/cmake/llvm -DCMAKE_IGNORE_PATH=/usr/lib/llvm-13;/usr/lib/llvm-13/lib -DCMAKE_PREFIX_PATH=/opt/rocm/llvm -DCMAKE_LIBRARY_PATH=/opt/rocm/llvm/lib"
```
evetsso: I don't think this change is necessary if all of the above suggestions are implemented.

CMakeLists.txt
stellaraccident: Also use VERBATIM

stellaraccident (commented):
Please get a review from @ellosel 

We're going to be deleting these default build files soon in favor of the rewrite, and we'll need to port this feature there or redo it later.

CMakeLists.txt
tomchengchitang: Thanks, great suggestion!

CMakeLists.txt
```diff
@@ -429,3 +408,109 @@ rocm_create_package(
...
+  set(LLVM_DIR /opt/rocm/llvm/lib/cmake/llvm)
+  find_package(LLVM REQUIRED CONFIG)
+  message(STATUS "LLVM found: ${LLVM_DIR}")
+
+  # Locate required LLVM tools (llvm-cov, llvm-profdata, etc.)
+  find_program(LLVM_PROFDATA_EXECUTABLE llvm-profdata HINTS ${rocm_path}/llvm/bin REQUIRED)
+  find_program(LLVM_COV_EXECUTABLE llvm-cov HINTS ${rocm_path}/llvm/bin REQUIRED)
+
+  # Use LLVM libraries and headers
+  set(LLVM_DIR ${rocm_path}/llvm/lib/cmake/llvm)
```
reboss: In our CI, we use `/opt/rocm/llvm/bin/llvm-profdata` and `/opt/rocm/llvm/bin/llvm-cov`

CMakeLists.txt
```diff
@@ -429,3 +408,109 @@ rocm_create_package(
...
+  find_package(LLVM REQUIRED CONFIG)
+
+  # Modern: Use target_* for compiler and linker flags
+  target_compile_options(hipblaslt PRIVATE
+    -fprofile-instr-generate
+    -fcoverage-mapping
+  )
+
+  target_link_options(hipblaslt PRIVATE
```
reboss: `target_link_options` should generally be `PUBLIC` for coverage flags.  Is there a reason you've made it `PRIVATE`?

CMakeLists.txt
```diff
@@ -429,3 +408,109 @@ rocm_create_package(
...
+  set(PROFRAW_FILES hipblaslt-test.profraw 
+                    sample_hipblaslt_ext_op_amax.profraw
+                    sample_hipblaslt_ext_op_amax_with_scale.profraw
+                    sample_hipblaslt_ext_op_layernorm.profraw
+                    sample_hipblaslt_weight_swizzle_padding.profraw)
+
+  add_custom_target(coverage
+    COMMAND ${CMAKE_COMMAND} -E env 
+            LLVM_PROFILE_VERBOSE=1 
+            LLVM_PROFILE_FILE=./hipblaslt-test.profraw
```
reboss: If any of your tests are done in parallel, this file will get corrupted because you'll have multiple processes writing to the same file.  To avoid this, the `LLVM_PROFILE_FILE` must have the %m or %p.  I think %m is preferred because %p is just the process ID and there's no guarantee that you'll get unique process IDs for every test

CMakeLists.txt
```diff
@@ -429,3 +408,109 @@ rocm_create_package(
...
+            --ignore-filename-regex=.*/samples/.*|.*/build/debug/.*
+    COMMAND ${LLVM_COV_EXECUTABLE} 
+            show 
+            -object ./clients/staging/hipblaslt-test
+            -object ./clients/staging/sample_hipblaslt_ext_op_amax
+            -object ./clients/staging/sample_hipblaslt_ext_op_amax_with_scale
+            -object ./clients/staging/sample_hipblaslt_ext_op_layernorm 
+            -object ./clients/staging/sample_hipblaslt_weight_swizzle_padding
+            -instr-profile=./coverage_test.profdata 
+            -format=html -output-dir=coverage_report
```
reboss, line 502: The folder should actually be named `coverage-report` with a dash, as that's where CI expects it to be.

clients/samples/CMakeLists.txt
```diff
@@ -135,7 +135,15 @@ foreach( exe ${sample_list_all} )
     SYSTEM PRIVATE
       $<BUILD_INTERFACE:${HIP_INCLUDE_DIRS}>
       )
-
+  if(BUILD_CODE_COVERAGE)
+    target_compile_options(${exe} PRIVATE
```
reboss: I don't think you need to instrument the samples.  They're just test/use cases, correct?

CMakeLists.txt
```diff
@@ -429,3 +408,109 @@ rocm_create_package(
...
+    COMMAND ${LLVM_COV_EXECUTABLE} 
+            show 
+            -object ./clients/staging/hipblaslt-test
+            -object ./clients/staging/sample_hipblaslt_ext_op_amax
+            -object ./clients/staging/sample_hipblaslt_ext_op_amax_with_scale
+            -object ./clients/staging/sample_hipblaslt_ext_op_layernorm 
+            -object ./clients/staging/sample_hipblaslt_weight_swizzle_padding
+            -instr-profile=./coverage_test.profdata 
+            -format=html -output-dir=coverage_report 
+            --ignore-filename-regex=.*/samples/.*|.*/build/debug/.*
```
reboss: If you don't instrument the samples/testcases themselves, then there'll be no need to ignore them here

CMakeLists.txt
```diff
@@ -429,3 +408,109 @@ rocm_create_package(
...
+            ./clients/staging/sample_hipblaslt_ext_op_layernorm 
+            --output-on-failure
+    COMMAND ${CMAKE_COMMAND} -E env 
+            LLVM_PROFILE_VERBOSE=1 
+            LLVM_PROFILE_FILE=./sample_hipblaslt_weight_swizzle_padding.profraw 
+            ./clients/staging/sample_hipblaslt_weight_swizzle_padding 
+            --output-on-failure
+    COMMAND ${LLVM_PROFDATA_EXECUTABLE}
+            merge -sparse
+            ${PROFRAW_FILES}
```
reboss: Wildcards do work for merging, unlike generating a report where they don't.  So if you end up adding the %p or %m option, then you could probably just do *.profraw here

CMakeLists.txt
```diff
@@ -49,6 +49,11 @@ if(NOT BUILD_CUDA)
 list( APPEND CMAKE_PREFIX_PATH /opt/rocm/llvm /opt/rocm )
 endif()
 
+
+if(BUILD_CODE_COVERAGE)
+  set(BUILD_SHARED_LIBS OFF CACHE BOOL "Build coverage reports" FORCE)
```
reboss: Doesn't this create a .so?  If so, unless you're losing a lot of functionality with a .so, you should build it and then just reference it when generating the report.  So instead of this below:
```
COMMAND ${LLVM_COV_EXECUTABLE} 
            report 
            -object ./clients/staging/hipblaslt-test
            -object ./clients/staging/sample_hipblaslt_ext_op_amax
            -object ./clients/staging/sample_hipblaslt_ext_op_amax_with_scale
            -object ./clients/staging/sample_hipblaslt_ext_op_layernorm 
            -object ./clients/staging/sample_hipblaslt_weight_swizzle_padding
            -instr-profile=./coverage_test.profdata 
            --ignore-filename-regex=.*/samples/.*|.*/build/debug/.*
```

You can do:
```
COMMAND ${LLVM_COV_EXECUTABLE} 
            report 
            -object ./library/src/libhipblaslt.so
            -instr-profile=./coverage_test.profdata 
```

reboss (changes requested):

clients/samples/CMakeLists.txt
tomchengchitang: Yes, you are right

CMakeLists.txt
tomchengchitang: It becomes `/opt/rocm/llvm/bin/llvm-profdata merge -sparse "*.profraw" -o coverage_test.profdata` in build.cmake, and the output shows `error: *.profraw: No such file or directory`.

And I tried to use both `file(GLOB PROFRAW_FILES "${CMAKE_BINARY_DIR}/*.profraw")` and `COMMAND ${CMAKE_COMMAND} -E sh -c...`, but 1st one doesn't work and the 2nd one is not portable.

tomchengchitang: Reference: https://github.com/ROCm/hipFFT/pull/132/files

CMakeLists.txt
tomchengchitang: NP!

CMakeLists.txt
```diff
@@ -408,61 +405,91 @@ rocm_create_package(
...
+    COMMAND ${LLVM_COV_EXECUTABLE} 
+            report 
+            -object ./library/libhipblaslt-d.so.0.15
+            -instr-profile=./coverage_test.profdata 
+            --ignore-filename-regex=.*/build/debug/.*
+    COMMAND ${LLVM_COV_EXECUTABLE} 
+            show 
+            -object ./library/libhipblaslt-d.so.0.15
+            -instr-profile=./coverage_test.profdata 
+            -format=html -output-dir=coverage_report
```
reboss:
```suggestion
            -format=html -output-dir=coverage-report 
```

CMakeLists.txt
```diff
@@ -408,61 +405,91 @@ rocm_create_package(
...
-
-  add_custom_target(coverage DEPENDS coverage_output)
-
-  #
-  # Coverage cleanup
-  #
-  add_custom_target(coverage_cleanup
-    COMMAND find ${CMAKE_BINARY_DIR} -name *.gcda -delete
+  # Force CMake to use ROCm's LLVM
+  set(LLVM_DIR ${rocm_path}/llvm/lib/cmake/llvm)
```
reboss: This should be set to `/opt/rocm/llvm/bin` for CI.  You may want to consider adding an `option()` with a default value of `/opt/rocm/llvm/bin` so that you can override it with the path to a custom llvm library, in case `/opt/rocm/llvm/bin` doesn't exist.

CMakeLists.txt
```diff
@@ -408,61 +405,91 @@ rocm_create_package(
...
+  find_package(LLVM REQUIRED CONFIG)
+  message(STATUS "LLVM found: ${LLVM_DIR}")
+
+  # Locate required LLVM tools (llvm-cov, llvm-profdata, etc.)
+  find_program(LLVM_PROFDATA_EXECUTABLE llvm-profdata HINTS ${rocm_path}/llvm/bin/llvm-profdata REQUIRED)
```
reboss:
```suggestion
  find_program(LLVM_PROFDATA_EXECUTABLE llvm-profdata HINTS ${LLVM_DIR} REQUIRED)
```

CMakeLists.txt
```diff
@@ -408,61 +405,91 @@ rocm_create_package(
...
+  find_program(LLVM_COV_EXECUTABLE llvm-cov HINTS ${rocm_path}/llvm/bin/llvm-cov REQUIRED)
```
reboss:
```suggestion
  find_program(LLVM_COV_EXECUTABLE llvm-cov HINTS ${LLVM_DIR} REQUIRED)
```

reboss (changes requested):

reboss (approved):

vin-huang (approved):
LGTM

PR #2069, opened 2025-05-12
Modify GEMM benchmarking when use_gpu_timer is enabled.

pghysels:
When using gpu timers, each GEMM is timed individually with 2 hipEvent_t's. Then the mean time is computed after removing outliers. The mean GEMM time as originally reported is significantly affected by outliers (regardless of the number of iterations). Removing the outliers makes the reported times more consistent.

Options bench_time (float) and cold_bench_time (float) have been added (only supported when running with --use_gpu_timer). The timing loop then runs for at least a given amount of time (cold_bench_time, bench_time) (by default 0., 0.) and with a minimum of (cold_iters, iters) iterations. For accurate results, the timing loop should run long enough such that the device thermals and clock frequencies can stabilize.

Now when using gpu timers, since GEMMs are timed individually, flushing the instruction cache is no longer included in the timing. When not using gpu timers, and when running with --flush, the time for flush_icache is subtracted from the benchmark time at the end (in argument_model.hpp), but this can occasionally lead to negative runtime or gflops. The time for flush_icache is computed separately by running it for (2x) 100000 iterations, which takes ~2 seconds.

I would now recommend running with --use_gpu_timer (but didn't make this default), because the reported times are much more consistent, and the benchmarks run faster with --flush enabled. However, the hipEvents introduce some overhead, measured to be ~2 microseconds per GEMM. This could skew the results for small GEMMs.

Default behavior should not be impacted.

(includes feedback from Tomoko.Kurotobi@amd.com)


docs/conceptual/hipblaslt-clients.rst
```diff
@@ -96,24 +97,31 @@ For more information, run the command with the ``--help`` option. The output of
...
    --scaleAlpha_vector                Apply scaleAlpha vector
+   --scaleABlockRowSize <value>       Set the row size of scale block for A                                               (Default value is: 32)
+   --scaleABlockColSize <value>       Set the column size of scale block for A                                            (Default value is: 1)
+   --scaleBBlockRowSize <value>       Set the row size of scale block for B                                               (Default value is: 1)
+   --scaleBBlockColSize <value>       Set the column size of scale block for B                                            (Default value is: 32)
    --amaxScaleA                       Apply scale for A buffer by abs max of A buffer
    --amaxScaleB                       Apply scale for B buffer by abs max of B buffer
    --amaxD                            Output Amax of intermediate D matrix
    --use_e                            Apply AUX output/ gradient input
+   --aux_type <value>                 Used with --use_e. Precision of AUX output (matrix E).Options: f16_r, default (same with D type)
```
amd-jnovotny:
```suggestion
   --aux_type <value>                 Used with --use_e. Precision of AUX output (matrix E). Options: f16_r, default (same with D type)
```

pghysels: Thanks. Fixed.

pghysels:
These changes make timings more consistent, as shown in the below figure. We ran 17 different GEMMs (different shapes/sizes), with the default 2 cold, 10 hot iterations, launching hipblaslt-bench 500 times to get the distribution of the times. With these changes, the standard deviation of the reported times is reduced.
The more consistent times help with our machine learning based solution selection efforts. 
times_0 02_0 1_default_freqFalse_flushFalse npy

TomokoKurotobi (approved):
I believe this PR is very helpful in terms of making benchmarking time per solution very consistent. I was suggested to run a solution for 20ms for warming up and 100ms for actual measurement (hot iterations) for consistent benchmarking result.

One challenge that I have seen with fixed number of iteration per problem size is below:
Let's say cold iteration count for a problem size is set to a number that GEMM would spend at least 20ms calculated from hardware spec or roofline model. I have seen some solutions taking 100x or more time than the best solution for the problem. In this case, even with skipping slow solution feature (warming up time is inevitable), warming up time of those slow solution can be 20ms * 100 = 2 seconds. Since the library has a lot of solutions, this can be a bottle neck to complete numbers of benchmarking to find the best solution out of all possible solutions for the problem size.

I have tried this PR branch, and I saw significant time improvement of benchmarking when I ran a lot of sizes on mi300x, which has a great number of GEMM solutions.

yenong-amd (approved):

hcman2:
I didn't get the plots. Of course, if we use remove_outliers, the variance will become smaller. We should compare this patch with a larger iter numbers (which means the stabilized state). If their plots are similar to each other, then we can say that remove_outliers can represent the stabilized performance.

pghysels:
> I didn't get the plots. Of course, if we use remove_outliers, the variance will become smaller. We should compare this patch with a larger iter numbers (which means the stabilized state). If their plots are similar to each other, then we can say that remove_outliers can represent the stabilized performance.

I observed that if you run for longer, there are just more outliers. Some iterations take way longer, and this still happens after running the GEMM for several seconds. But I agree that for reliable results, you should run for a certain amount of time so that the thermals and frequency can stabilize.

jayhawk-commits:

PR #2071, opened 2025-05-12
[next-cmake] Find and link CBLAS

marbre:


davidd-amd (approved):

stellaraccident (approved):
Thanks. Iiuc, order here is important, right: the BLAS provider may also provide the cblas package so must be found first.

PR #2075, opened 2025-05-13
Filesystem dependency update

davidd-amd:
Resolves linking issue on rhel systems with gcc < 9. See https://github.com/ROCm/MIOpen/pull/3126 for additional details.

### DIFF
--- a/cmake/CheckCXXLinkerFlag.cmake
+++ b/cmake/CheckCXXLinkerFlag.cmake
@@ -1,0 +1,61 @@
+################################################################################
+#
+# MIT License
+#
+# Copyright (c) 2024 Advanced Micro Devices, Inc.
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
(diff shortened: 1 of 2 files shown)


cmake/CheckCXXLinkerFlag.cmake
```diff
@@ -0,0 +1,61 @@
...
+)
+
+include (CheckCXXSourceCompiles)
+
+function(check_cxx_linker_flag _flag _var)
+    set (_source "int main() { return 0; }")
+    set(CMAKE_REQUIRED_LINK_OPTIONS "${_flag}")
+    check_cxx_source_compiles("${_source}" "_${_var}" ${check_cxx_linker_flag_patterns})
+    set(${_var} "${_${_var}}" PARENT_SCOPE)
+endfunction()
```
bstefanuk, line 61: Newline EOF

cmake/CheckCXXLinkerFlag.cmake
```diff
@@ -0,0 +1,61 @@
+################################################################################
+#
+# MIT License
+#
+# Copyright (c) 2024 Advanced Micro Devices, Inc.
```
bstefanuk, line 5: 2025?

bstefanuk (approved):
The contents of CheckCXXLinkerFlag.cmake looks copy-pasted from somewhere, is this the case? Where from?

This looks like it's guarding against adding stdc++fs on a compiler that doesn't support it. Could you add to the current PR description how this is achieving your goal?

PR #2078, opened 2025-05-13
[next-cmake] Disable marker library on Windows

marbre:


davidd-amd (approved):
Awesome - thanks @marbre!

stellaraccident (approved):
Thank you for making this a default vs hardcoded. Gives us a path to enable on Windows eventually (versus the hard coding I see often done on this).

PR #2079, opened 2025-05-13
Add roc namespace to hipblas-common

davidd-amd:
- adds roc:: to hipblas(-common) targets

### DIFF
--- a/next-cmake/CMakeLists.txt
+++ b/next-cmake/CMakeLists.txt
@@ -198,7 +198,7 @@ if(HIPBLASLT_ENABLE_HOST)
 
     target_link_libraries(hipblaslt
         PUBLIC
-            ${hipblas_target}
+            roc::${hipblas_target}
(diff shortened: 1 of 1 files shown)


marbre (approved):
Confirmed to fix an build issue in TheRock!

next-cmake/CMakeLists.txt
```diff
@@ -198,7 +198,7 @@ if(HIPBLASLT_ENABLE_HOST)
 
     target_link_libraries(hipblaslt
         PUBLIC
-            ${hipblas_target}
+            roc::${hipblas_target}
```
stellaraccident, line 201: Nit for the future: when using non-optional string substitution like this, it can be a little better to quote it. That way you at least get an error about some library not being found at generation time vs missing usage requirements.

stellaraccident (approved):

PR #2087, opened 2025-05-15
[next-cmake] Required updates to next-cmake discovered during install script integration

davidd-amd:
- Adds missing sample test
- Adds missing benchmark executable targets
- Fixes open mp dependency
- Integrates required Tensile options
- Cleans up rocm-cmake packaging usage


next-cmake/CMakeLists.txt
```diff
@@ -194,6 +200,7 @@ if(HIPBLASLT_ENABLE_HOST)
         add_library(hipblaslt STATIC)
         target_compile_definitions(hipblaslt PRIVATE HIPBLASLT_STATIC_LIB)
     endif()
+    rocm_set_soversion(hipblaslt ${hipblaslt_SOVERSION})
```
bstefanuk, line 203: This appears to duplicate line 42: `set(hipblaslt_SOVERSION 0.15)`
Is it achieving something different? If so, can they be collected?

next-cmake/CMakeLists.txt
```diff
@@ -409,6 +440,7 @@ if(HIPBLASLT_ENABLE_SAMPLES)
         TARGETS
             sample_hipblaslt_gemm
             sample_hipblaslt_gemm_ext
+            sample_hipblaslt_gemm_ext_deprecated
```
bstefanuk, line 443: Is this deprecated sample required for some specific reason?

next-cmake/CMakeLists.txt
```diff
@@ -453,7 +485,7 @@ if(HIPBLASLT_ENABLE_SAMPLES)
     )
     rocm_package_setup_client_component(samples)
 endif()
-
+set(BUILD_SHARED_LIBS ${HIPBLASLT_BUILD_SHARED_LIBS})
```
bstefanuk, line 491: Should we prefer to build shared libs on a per target basis? If I understand correctly, this will cause all libraries to be built as shared instead of just the hipblaslt host lib target.

bstefanuk (dismissed):

next-cmake/CMakeLists.txt
davidd-amd: This is setting the so version on the hipblaslt target so we get something lie libhipblaslt.so.0.15. On line 42 we are setting the variable so version to 0.15.

next-cmake/CMakeLists.txt
davidd-amd: yeah testing expects it to be there

next-cmake/CMakeLists.txt
davidd-amd: I should add a comment as to why we are doing this but what a target is depends on how this is set at the time that we call add_library. Given that we are calling it after adding all of our libs they won't be impacted.

next-cmake/clients/CMakeLists.txt
```diff
@@ -65,11 +65,40 @@ target_link_libraries(hipblaslt-bench
...
+
+add_executable(hipblaslt-bench-extop-softmax "${TEMP_CLIENTS_SOURCE_DIR}/benchmarks/client_extop_softmax.cpp")
+target_link_libraries(hipblaslt-bench-extop-softmax PRIVATE hipblaslt-clients-common hip::device)
+target_compile_definitions(hipblaslt-bench-extop-softmax PRIVATE ROCBLASLT_INTERNAL_API )
+
+add_executable(hipblaslt-bench-extop-amax "${TEMP_CLIENTS_SOURCE_DIR}/benchmarks/client_extop_amax.cpp")
+target_link_libraries(hipblaslt-bench-extop-amax PRIVATE hipblaslt-clients-common hip::device)
+target_compile_definitions(hipblaslt-bench-extop-amax PRIVATE ROCBLASLT_INTERNAL_API )
+
+find_package(LLVM REQUIRED)
```
stellaraccident, line 95: Not for now, but this is a really shady way to be consuming YAML outside of LLVM. We likely want to sever this dep at some point.

stellaraccident (approved):

PR #2090, opened 2025-05-16
Fix SIA bug in gfx950

KKyang:


tensilelite/Tensile/KernelWriter.py
```diff
@@ -1125,8 +1124,8 @@ def calculateRangeAndUpdateCounter(itemCounter, writeCounters, length):
             if (mfmaIndex >= self.states.lwStartMfmaIndex) and not countGlobalRead(globalReadCode):
               writeItemLength = (localWriteCodeCounts[-1] - itemCounter) if localWriteCodeCounts else 0
               writeItemLength = min(writeItemLength, self.states.numLocalWriteModPerMfma)
-              numLoops, itemCounter = calculateRangeAndUpdateCounter(itemCounter, localWriteCodeCounts, writeItemLength)
-              for j in range(numLoops):
+              numLoops, _ = calculateRangeAndUpdateCounter(itemCounter, localWriteCodeCounts, writeItemLength)
```
b-shi, line 1127: Is it okay to not return the new `itemCounter` value? The value seems to be updated in `calculateRangeAndUpdateCounter`.

KKyang: Shouldn't return here because it's not popping items. The original code just calculates the latencyLeft here and starts from the beginning to pop items.

b-shi (approved):

aazz44ss (approved):

PR #2104, opened 2025-05-20
logics: temp workaround to avoid solutions from being rejected with xccg -1

solaslin:
SWDEV-528780
SWDEV-522233
SWDEV-533436


solaslin:
We have another alternation solution: https://github.com/ROCm/hipBLASLt/pull/2115 which I think is better. Once we merge 2115, I'll close this one.

solaslin:
#2115 is a better solution for these issues. Let's close this one.

PR #2108, opened 2025-05-20
Fixing build support for RockyLinux and add documentation

G-Ragghianti:
There doesn't seem to be a reason that RHEL and almalinux would not be supported in install.sh but rockylinux would not.  Adding rockylinux (rocky).  

Also adding a note that the python development headers are required.


KKyang (approved):

jayhawk-commits:

PR #2119, opened 2025-05-24
Enable mbsk spmm

vin-huang:
1. Added requiredSynchornizerSizes() 
2. Enable MBSK for SPMM
3. Add MBSK test for SPMM


vin-huang:
The Azure CI was failed due to cannot find proper device.

Hip error: 'no ROCm-capable device is detected'(100) at /agent/_work/1/s/library/src/amd_detail/hipblaslt.cpp:171

tensilelite/Tensile/Common/RegisterPool.py
```diff
@@ -236,8 +237,11 @@ def checkOutAligned(self, size, alignment, tag="_untagged_aligned_", preventOver
       oldSize = len(self.pool)
       if self.occupancyLimitSize > 0:
         if newSize > self.occupancyLimitSize and newSize <= self.occupancyLimitMaxSize:
-          print("newSize", newSize, "OldSIze", oldSize, "Limit", self.occupancyLimitSize)
-          assert self.occupancyLimitSize >= newSize
+          print(self.type, "newSize", newSize, "OldSIze", oldSize, "Limit", self.occupancyLimitSize)
+          if self.states == None:
+            assert self.occupancyLimitSize >= newSize
+          else:
+            self.states.overflowedResources = 4
```
KKyang: Don't bypass the guard, this is meant to fail. The calculation should be done outside and shouldn't exceed this limitation.

vin-huang: No, this kernel will be rejected.
Writer will check the overflowedResources and reject this kernel.

KKyang: No, I mean don't reject it. It should fail. When seeing this means the occupancy calculation is incorrect.

vin-huang: Removed the rejection way. Trying to prepare the required sgpr/vgpr before calculating the occupancy limit.
[c1c8201e](https://github.com/ROCm/hipBLASLt/pull/2119/commits/c1c8201e6189a44e79f0749a46defd3e445e6d99)

KKyang (dismissed):

tensilelite/Tensile/Components/GSU.py
```diff
@@ -867,14 +867,19 @@ def GSUSynccodegen(self, writer, kernel, tmpVgpr, tmpVgprSize, tmpVgprDynamic, b
                     vmcnt = SyncloadedData = SyncloadedData -1
                     module.add(SWaitCnt(lgkmcnt=lgkmcnt, vmcnt=vmcnt, vscnt=vscnt, comment="(wait for buffer ready)"))
 
-                    if ((gwvw % 2) == 1):
+                    if kernel["ProblemType"]["DataType"].isInt8() or kernel["ProblemType"]["DataType"].isInt32():
```
jichangjichang, line 870: should we check kernel["ProblemType"]["ComputeDataType"].isInt32() ?

vin-huang: In the original case, which didn't consider the output datatype is int32 from mfma/smfma instruction.
So, here I use kernel["ProblemType"]["DataType"] to identify the output datatype.

jichangjichang (approved):

KKyang (approved):

PR #2126, opened 2025-05-27
Origami solution selection

AlexBrownAMD:
Add a new solution selection mechanism using the Origami library to select the best performing kernel.

This PR does not include any Origami library logic files. Sample libraries are available, and initial libraries will be checked in later in a separate PR.


aazz44ss (dismissed):

hcman2:
We have both select_best_macro_tile_size and select_best_grid_size. Will we use predict MT and Gridbased combination in the future?

tensilelite/Tensile/Source/lib/source/analytical/AnalyticalGemm.cpp
```diff
@@ -0,0 +1,917 @@
...
+            // printf("MT_M:%d MT_N:%d MT_K:%d arith:%d\n", MT_M, MT_N, MT_K, mt_arith);
+            // size_t arith = ((M * N * K * 2) / (M * K + N * K + M * N));
+            size_t L_MT = L_MI * N_MI;
+
+            //TN
+            if(transA && !transB)
+            {
+                //We want to penalize tiles that can't be coalesced for T,N where K is contiguous dimension.
+                //In this case, that's when the K dimension is less than 128 bytes.
+                if(MT_K * safe_ceil_div(element_size_A, 8) % 128 != 0)
```
hcman2, line 118: This condition is somehow different from the description above. 
Why not use MT_K*element_size_A < 128 which means the K dimension is less than 128 bytes.

ryanswann-amd: Yes, the comment should be updated. In reality, it's important that it is a multiple of 128-bytes. Being less than 128 bytes will never be a multiple of 128-bytes, but there are larger numbers that are also not multiples. In testing, tile dimensions which are not a multiple of 128-bytes in the contiguous dimension performed poorly.

ryanswann-amd:
> We have both select_best_macro_tile_size and select_best_grid_size. Will we use predict MT and Gridbased combination in the future?

select_best_macro_tile_size : Determines the Macro Tile dimensions and DepthU (Internally called MT_M,MT_N,MT_K)

select_best_grid_size : Is an example of a dynamic mode for streamk, this is used to predict the number of workgroups it's best to launch for streamk, which then determines StreamK's splitting.

pghysels:
There are two slightly conflicting declarations of `compute_reuse_in_block_gemm`, but they are both not used? Looks like nothing in Reuse.hpp/.cpp is currently used?

tensilelite/Tensile/Source/lib/include/Tensile/PredictionLibrary.hpp
```diff
@@ -0,0 +1,224 @@
...
+
+#include <set>
+#include <vector>
+
+#include <Tensile/analytical/Utils.hpp>
+
+namespace TensileLite
+{
+    template <typename Value>
+    struct PredictionEntry
```
pghysels: PredictionEntry is not used?

ryanswann-amd:
> There are two slightly conflicting declarations of `compute_reuse_in_block_gemm`, but they are both not used? Looks like nothing in Reuse.hpp/.cpp is currently used?

This is legacy code that was determined to be too slow. Will Remove

tensilelite/Tensile/Source/lib/include/Tensile/PredictionLibrary.hpp
ryanswann-amd: I see it being used here : https://github.com/ROCm/hipBLASLt/blob/6122908a497b1538da5bd1ecc9782e3043b2fd70/tensilelite/Tensile/Source/lib/include/Tensile/Serialization/Containers.hpp#L211-L215

yenong-amd: This was legacy from previous implementation of the library. I can remove them.

tensilelite/Tensile/TensileCreateLibrary/Run.py
```diff
@@ -386,6 +391,12 @@ def assemble(ret):
         multiArg=False,
         return_as="list"
     )
+    # passPostKernelInfoToSolution(
+    #     ret, uniqueAsmKernels, solutions, splitGSU
+    # )
+    # result.src is very large so let garbage collector know to clean up
```
bstefanuk: Delete

tensilelite/Tensile/TensileCreateLibrary/Run.py
```diff
@@ -90,6 +90,11 @@ class KernelCodeGenResult(NamedTuple):
     pgr: int
     mathclk: int
 
+class KernelMinResult(NamedTuple):
+    err: int
+    cuoccupancy: int
+    pgr: int
+    mathclk: int
```
bstefanuk: Nit: I wonder if there's a better name we can use for this that's more descriptive of it being the minimum code gen result data needed for the prediction model?

tensilelite/Tensile/TensileCreateLibrary/Run.py
```diff
@@ -332,6 +335,7 @@ def writeSolutionsAndKernelsTCL(
     outputPath,
     asmToolchain,
     srcToolchain,
+    solutions,
```
bstefanuk: Looks like this new parameter is unused. This should be removed alongside the other commented code in this function (where it appeared to have been used).

tensilelite/Tensile/TensileCreateLibrary/Run.py
```diff
@@ -707,6 +719,15 @@ def validLogicFile(p: Path):
...
     splitGSU = False
 
+
+    solDict = {}
+    for solution in solutions:
+        solutionKernels = solution.getKernels()
+        for kernel in solutionKernels:
+            kName = getKeyNoInternalArgs(kernel, False)
+            if kName not in solDict:
+                solDict["%s"%kName] = kernel
```
bstefanuk: It would be nice to extract this into a function. Also, that way, the intent would be clearer.

tensilelite/Tensile/TensileCreateLibrary/Run.py
```diff
@@ -332,6 +335,7 @@ def writeSolutionsAndKernelsTCL(
     outputPath,
     asmToolchain,
     srcToolchain,
+    solutions,
     kernels,
     kernelHelperObjs,
     kernelWriterAssembly,
```
bstefanuk: Since this function now has a return value, could you type hint this with `) -> KernelMinResult:`, or whatever you choose to name that struct.

bstefanuk (commented):
Looks good, a few minor comments with regards to the TensileCreateLibrary code.

bethune-bryant (dismissed):

neoblizz (dismissed):

tensilelite/Tensile/Source/lib/source/analytical/AnalyticalGemm.cpp
```diff
@@ -0,0 +1,895 @@
...
+                                   size_t          element_size_out,
+                                   double          H_mem1,
+                                   bool            debug)
+        {
+            // Compute total FLOPs
+            double total_FLOPs = 2.0 * M * N * K; // For GEMM, each multiply-add is 2 FLOPs
+            // Compute total time in seconds
+            double cycles_per_second
+                = hardware.compute_clock_ghz * 1e9; // 1 GHz = 1e9 cycles per second
+            size_t WGM                = 6;
```
hcman2: Is this WGM=6 a HW dependent default value? Different HW has different numbers of CUs  and XCDs. Is it OK to use WGM=6 for all the HWs?

ryanswann-amd: It is a HW dependent value. I think for this specific function, it should be an input parameter (since it's a parameter of the solution, and the solution is problem-size-dependent and not hardware-dependent. But I think that _generally_ the default should be sqrt(active_cus_per_l2) or sqrt(total_cus_per_l2) since that's the largest creatable square, if there needs to be a default. I'll change this to an input parameter for now though.

pghysels:
Perhaps you should add a description for `TENSILE_STREAMK_DYNAMIC_GRID=4,5` in ValidParameters.py and for `TENSILE_STREAMK_DYNAMIC_WGM` in ValidParameters.py and in how-to-use-streamk.rst?

Is `TENSILE_STREAMK_DYNAMIC_GRID=3` still a good default for Origami?

ryanswann-amd:
> Perhaps you should add a description for `TENSILE_STREAMK_DYNAMIC_GRID=4,5` in ValidParameters.py and for `TENSILE_STREAMK_DYNAMIC_WGM` in ValidParameters.py and in how-to-use-streamk.rst?
> 
> Is `TENSILE_STREAMK_DYNAMIC_GRID=3` still a good default for Origami?

I think that we wanted to move from TENSILE_STREAMK_DYNAMIC_GRID=3 as the default to either 4 or 5. Thoughts @AlexBrownAMD ? I can also add Descriptions where you mentioned.

neoblizz:
> > Perhaps you should add a description for `TENSILE_STREAMK_DYNAMIC_GRID=4,5` in ValidParameters.py and for `TENSILE_STREAMK_DYNAMIC_WGM` in ValidParameters.py and in how-to-use-streamk.rst?
> > Is `TENSILE_STREAMK_DYNAMIC_GRID=3` still a good default for Origami?
> 
> I think that we wanted to move from TENSILE_STREAMK_DYNAMIC_GRID=3 as the default to either 4 or 5. Thoughts @AlexBrownAMD ? I can also add Descriptions where you mentioned.

I think we should start using `5` for now, there's a future PR that should consolidate all of these.

tensilelite/Tensile/Source/lib/source/analytical/Utils.cpp
```diff
@@ -0,0 +1,660 @@
...
+                                     size_t          mx_block_size,
+                                     double          H_L2,
+                                     bool            debug,
+                                     size_t          WGM,
+                                     size_t          biggest_allowable_split)
+        {
+
+            // compute how many 32×32 tiles are needed in each dim,
+            // then multiply to get total grid size:
+            size_t grid = ((M + MT_M - 1) / MT_M) * ((N + MT_N - 1) / MT_N);
```
pghysels: `size_t grid = ((M + MT_M - 1) / MT_M) * ((N + MT_N - 1) / MT_N) * batch;`

ryanswann-amd:
> > > Perhaps you should add a description for `TENSILE_STREAMK_DYNAMIC_GRID=4,5` in ValidParameters.py and for `TENSILE_STREAMK_DYNAMIC_WGM` in ValidParameters.py and in how-to-use-streamk.rst?
> > > Is `TENSILE_STREAMK_DYNAMIC_GRID=3` still a good default for Origami?
> > 
> > 
> > I think that we wanted to move from TENSILE_STREAMK_DYNAMIC_GRID=3 as the default to either 4 or 5. Thoughts @AlexBrownAMD ? I can also add Descriptions where you mentioned.
> 
> I think we should start using `5` for now, there's a future PR that should consolidate all of these.

This will Impact all streamk kernels though right? Should we check other libraries being built with StreamK function well with this before changing it?

tensilelite/Tensile/Source/lib/include/Tensile/PredictionLibrary.hpp
```diff
@@ -0,0 +1,222 @@
...
+                *(pAMDGPU->analyticalHardware),
+                tile_list,
+                elementSizeA_bits,
+                elementSizeA_bits,
+                elementSizeC_bits,
+                0, //mx_block_size -> MX Data types come from rocroller.
+                0.8,
+                debug,
+                false,
+                6);
```
bethune-bryant: @ryanswann-amd , I think you meant to use this here?

```suggestion
                WGM);
```

bethune-bryant (approved):

pghysels (approved):

aazz44ss (approved):

bnemanich (approved):

PR #2128, opened 2025-05-27
Changes to support using rocroller next-cmake

davidd-amd:
- conditionally propagate rocroller when enabled
- restore building mxDataGen.cpp
- add new object lib to set rocroller specific config
- removed duplicate symbol


next-cmake/CMakeLists.txt
```diff
@@ -273,9 +273,15 @@ if(HIPBLASLT_ENABLE_HOST)
     add_subdirectory(host-library)
 endif()
 
+if(HIPBLASLT_ENABLE_ROCROLLER)
+    target_link_libraries(hipblaslt PRIVATE roc::rocroller)
+    target_compile_definitions(hipblaslt PUBLIC USE_ROCROLLER)
```
stellaraccident: Can you make a note to followup: unqualified CPP defines like `USE_ROCROLLER` will age badly with multiple projects. Should always be scoped with the project name.

next-cmake/clients/CMakeLists.txt
```diff
@@ -53,8 +53,13 @@ target_link_libraries(hipblaslt-clients-common
     PRIVATE
         hip::device
 )
+
 target_compile_definitions(hipblaslt-clients-common PRIVATE CLIENTS_NO_FORTRAN) # necessary?
 
+if(HIPBLASLT_ENABLE_ROCROLLER)
+    target_compile_definitions(hipblaslt-clients-common PRIVATE USE_ROCROLLER)
```
stellaraccident: Ditto on define.

stellaraccident (dismissed):

next-cmake/CMakeLists.txt
davidd-amd: Good idea - will add issue.

library/src/amd_detail/rocblaslt/src/handle.cpp
```diff
@@ -50,9 +50,9 @@ _rocblaslt_handle::_rocblaslt_handle()
     asic_rev = 0;
 #endif
 
-#ifdef USE_ROCROLLER
+#ifdef HIPBLASLT_USE_ROCROLLER
     rocroller_create_handle(&rocroller_handle);
-    const char* rocRollerEnvVal = std::getenv("HIPBLASLT_USE_ROCROLLER");
+    const char* rocRollerEnvVal = std::getenv("HIPBLASLT_HIPBLASLT_USE_ROCROLLER");
```
stellaraccident: Find/replace gaffe? Back this line out?

stellaraccident (changes requested):

library/src/amd_detail/rocblaslt/src/handle.cpp
davidd-amd: Sad face - good catch!

stellaraccident (approved):
Thanks for this and for addressing comments.

PR #2132, opened 2025-05-28
Rejecting solution for StoreSwapAddr with LDS buffer size >= 64K

adityalj:


tensilelite/Tensile/Components/LocalRead.py
```diff
@@ -595,11 +595,17 @@ def __call__(self, writer, kernel, bufferIdx, iui, epsi, tP):
     
     
                         if(paramList[0] >=131072):
-                            paramList[0] = paramList[0] -131072
-                            srcAddr=vgpr("LocalReadAddr%s+2"%tc)
+                            if numVgprLocalReadAddr ==3:
```
nakajee, line 598: Do we really need to do error check here?
It should be in Solution.py.
Is there any reason why you implemented the check here?

tensilelite/Tensile/SolutionStructs/Solution.py
```diff
@@ -2604,6 +2604,12 @@ def subCheckLdsBlockSizePerPad(tc, idx):
       state["LdsOffsetB"] = state["LdsOffsetMetadata"] + ldsNumBytesAlignedMetadata
       ldsNumBytesAB = state["LdsOffsetB"] + ldsNumBytesB
 
+    need64K = state["LdsOffsetA_Blk"]>=65536 and state["ExpandPointerSwap"] and not state["1LDSBuffer"]
```
nakajee, line 2607: I think it is better to use maxLDSConstOffset instead of 65536.

PR #2136, opened 2025-05-29
Add back old design of scale A/B vector temporarily

jichangjichang:
Keep both old design and standard way 
I will remove it once framework migrate to standard way.


jichangjichang:
Use old way to call scale vector and test pass
[==========] 54611 tests from 13 test suites ran. (2239374 ms total)
[  PASSED  ] 54611 tests.
hipBLASLt version: 1500

aazz44ss (approved):

Serge45 (approved):

PR #2162, opened 2025-06-02
Use hip f8 header for rocm 6.3+

b-shi:
Update macro checks so hip f8 headers is used for rocm6.3+

### DIFF
--- a/library/include/hipblaslt/hipblaslt-types.h
+++ b/library/include/hipblaslt/hipblaslt-types.h
@@ -33,7 +33,7 @@
 #define _HIPBLASLT_TYPES_H_
 
 #if(HIP_VERSION_MAJOR == 6 && HIP_VERSION_MINOR == 2 && HIP_VERSION_PATCH > 42130) \
-    || (HIP_VERSION_MAJOR == 6 && HIP_VERSION_MINOR >= 3) //tmp before gfx94 use hip f8 header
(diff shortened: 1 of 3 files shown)


library/include/hipblaslt/hipblaslt-types.h
```diff
@@ -33,7 +33,7 @@
 #define _HIPBLASLT_TYPES_H_
 
 #if(HIP_VERSION_MAJOR == 6 && HIP_VERSION_MINOR == 2 && HIP_VERSION_PATCH > 42130) \
-    || (HIP_VERSION_MAJOR == 6 && HIP_VERSION_MINOR >= 3) //tmp before gfx94 use hip f8 header
+    || (HIP_VERSION_MAJOR >= 6 && HIP_VERSION_MINOR >= 3) //tmp before gfx94 use hip f8 header
```
nakajee, line 36: This is not correct.
It should be
  MAJOR >=7 || (MAJOR==6 && MINOR >= 3)

b-shi: Ah... really silly mistake on my part.. :facepalm:

b-shi:
The current plan is to remove the backward compatibility headers ([PR](https://github.com/ROCm/hipBLASLt/pull/2165) here). So these macro checks are no longer needed. Closing.

PR #2163, opened 2025-06-03
Bump version to 1.0

KKyang:


TorreZuk (dismissed):
Looks good

CHANGELOG.md
```diff
@@ -2,6 +2,13 @@
 
 Full documentation for hipBLASLt is available at [rocm.docs.amd.com/projects/hipBLASLt](https://rocm.docs.amd.com/projects/hipBLASLt/en/latest/index.html).
 
+## hipbBLASLt 1.0.0 for ROCm 7.0.0
```
jichangjichang: hipBLASLt

CHANGELOG.md
```diff
@@ -2,6 +2,13 @@
...
+
+### Changed
+
+* ``HIPBLASLT_MATMUL_DESC_A_SCALE_POINTER_VEC_EXT`` and ``HIPBLASLT_MATMUL_DESC_B_SCALE_POINTER_VEC_EXT`` are deprecated, ``use ROCBLASLT_MATMUL_DESC_A_SCALE_MODE`` and ``ROCBLASLT_MATMUL_DESC_B_SCALE_MODE`` attributes to set scalar (``HIPBLASLT_MATMUL_MATRIX_SCALE_SCALAR_32F``) or vector (``HIPBLASLT_MATMUL_MATRIX_SCALE_OUTER_VEC_32F``).
```
jichangjichang: use ``

jichangjichang: VEC_EXT should be in  Deprecations section

CHANGELOG.md
KKyang: Fixed

CHANGELOG.md
KKyang: Fixed

jichangjichang (approved):

davidd-amd (approved):

CHANGELOG.md
```diff
@@ -2,6 +2,17 @@
 
 Full documentation for hipBLASLt is available at [rocm.docs.amd.com/projects/hipBLASLt](https://rocm.docs.amd.com/projects/hipBLASLt/en/latest/index.html).
 
+## hipBLASLt 1.0.0 for ROCm 7.0.0
+
+### Changed
+
+* ``HIPBLASLT_MATMUL_DESC_A_SCALE_POINTER_VEC_EXT`` and ``HIPBLASLT_MATMUL_DESC_B_SCALE_POINTER_VEC_EXT`` are deprecated, use ``ROCBLASLT_MATMUL_DESC_A_SCALE_MODE`` and ``ROCBLASLT_MATMUL_DESC_B_SCALE_MODE`` attributes to set scalar (``HIPBLASLT_MATMUL_MATRIX_SCALE_SCALAR_32F``) or vector (``HIPBLASLT_MATMUL_MATRIX_SCALE_OUTER_VEC_32F``).
+* The non-V2 APIs in the Cpp header are all deprecated and replaced by V2 APIs.
```
bnemanich, line 10: I think this is a little off. The non-V2 APIs were actually removed, not deprecated. If they were just deprecated, they would still be able to be used for now.

I think we also removed the existing V2 APIs, which we probably didn't want to do. We should probably leave in the existing V2 APIs for now, but mark them as deprecated and to be removed in a future release. Hopefully we can just implement them with `using` statements pointing to the new "V1" classes.

I think the CHANGELOG should also be explicit about which classes were removed/deprecated. Instead of saying the "non-V2" APIs, please list out all of the class names that were removed/replaced.

I also think that `[[deprecated]]` attributes should be added to everything that is deprecated. This way it is much more obvious that something will be removed in the future and builds will spit out warnings when trying to use deprecated functions.

PR #2164, opened 2025-06-03
[rocisa] Add RegisterPool to rocisa

KKyang:
also fixed a bug when HIP kernels may accidentally get removed.

Root cause:

``names`` is not sorted but ``kho`` is, so they may have different order and cause the kernel accidentally been removed.

### DIFF
--- a/tensilelite/Tensile/TensileCreateLibrary/Run.py
+++ b/tensilelite/Tensile/TensileCreateLibrary/Run.py
@@ -468,6 +468,8 @@ def generateKernelHelperObjects(solutions: List[Solution], cxxCompiler: str, isa
             buildMask = []
             names = callable(solution)
             if names:
+                sortByEnum = lambda x: ("Enum" in x, x)
+                sorted(names, key=sortByEnum, reverse=True)
                 for name in names:
                     if name not in visited:
                         visited.add(name)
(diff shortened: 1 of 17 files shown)


hcman2 (dismissed):
ok if ci passed

tensilelite/Tensile/TensileCreateLibrary/Run.py
```diff
@@ -476,9 +476,12 @@ def generateKernelHelperObjects(solutions: List[Solution], cxxCompiler: str, isa
                         buildMask.append(False)
                 if any(buildMask):
                     kho = initHelperKernelObjects(solution, kernelHelperType, cxxCompiler, isaInfoMap)
-                    kho = list(itertools.compress(kho, buildMask))
-                    if kho:
-                        khos.extend(kho)
+                    # Pair each kho with its name, then sort to match names order
```
KKyang: @ellosel There is a chance that buildMask is [A, B] while kho is [B, A]. Leading incorrect ko been pushed to khos.

davidd-amd: @KKyang I don't have any objections with the change but can you explain in greater detail in the description what the problem is and how this resolves the issue. I want to be sure I understand so I don't break things in the future. Thanks!

KKyang: Done

aazz44ss (approved):

hcman2 (approved):

KKyang:
image

PR #2167, opened 2025-06-03
Enable building on Windows with necessary modifications

amd-mtrifuno:
* Add script to download Windows dependencies
* Introduce rmake.py script for Windows builds
* Disable roctracer and rocroller - unsupported on Windows
* Update packaging configurations for Windows
* Include OpenBLAS library for Windows builds
* Adapt environment variable settings for Windows
* Implemented cross-platform libraries copying
* Fix linker issues related to platform differences


clients/gtest/CMakeLists.txt
```diff
@@ -89,32 +94,44 @@ target_compile_definitions( hipblaslt-test PRIVATE ROCM_USE_FLOAT16 HIPBLASLT_IN
...
-#  if (NOT BUILD_FORTRAN_CLIENTS)
-#    list( APPEND COMMON_LINK_LIBS "-lgfortran") # for lapack
-#  endif()
-#else()
-#  list( APPEND COMMON_LINK_LIBS "libomp")
-#endif()
 
 target_link_libraries( hipblaslt-test PRIVATE ${COMMON_LINK_LIBS} )
 
+if (WIN32)
```
stellaraccident, line 118: Sorry, we can't land changes like this anymore. It is never appropriate to be copying files from system32 or using a platform gate to move files around by absolute path. Getting these things lined up is a job for whatever is building this project, not just something to be done unconditionally here.

library/CMakeLists.txt
```diff
@@ -78,6 +78,22 @@ endif()
 
 # Target compile definitions
 if(NOT BUILD_CUDA)
+  if (WIN32)
```
stellaraccident, line 81: No - I don't see why this is win32 guarded.

library/CMakeLists.txt
```diff
@@ -207,6 +223,10 @@ set_target_properties(hipblaslt PROPERTIES CXX_VISIBILITY_PRESET "hidden" VISIBI
 set_target_properties(hipblaslt PROPERTIES RUNTIME_OUTPUT_DIRECTORY "${PROJECT_BINARY_DIR}/staging")
 set_target_propertieS(hipblaslt PROPERTIES DEBUG_POSTFIX "-d")
 
+if (WIN32 AND BUILD_CLIENTS)
```
stellaraccident, line 226: Cannot land things like this anymore. This is not win32 generally but related to the expectations of your build system.

library/CMakeLists.txt
```diff
@@ -239,7 +259,7 @@ install(
 
 if ( NOT BUILD_CUDA )
     if (WIN32)
-      set( HIPBLASLT_TENSILE_LIBRARY_DIR "\${CPACK_PACKAGING_INSTALL_PREFIX}hipblaslt/bin" CACHE PATH "path to tensile library" )
+      set( HIPBLASLT_TENSILE_LIBRARY_DIR "\${CPACK_PACKAGING_INSTALL_PREFIX}/bin/hipblaslt" CACHE PATH "path to tensile library" )
```
stellaraccident, line 262: Fyi - we're aligning these libraries to ask be based on the posix paths. The hipBLASLt host code has already been updated to use lib as the location for all platforms.

library/src/amd_detail/rocblaslt/src/extops/CMakeLists.txt
```diff
@@ -142,21 +142,31 @@ foreach(arch IN LISTS archs)
...
-    COMMAND ${CMAKE_COMMAND} -E copy "${CMAKE_CURRENT_BINARY_DIR}/*.co" ${PROJECT_BINARY_DIR}/Tensile/library
-    COMMAND ${CMAKE_COMMAND} -E copy "${CMAKE_CURRENT_BINARY_DIR}/*.dat" ${PROJECT_BINARY_DIR}/Tensile/library
-    COMMENT "Copying ExtOp Library"
-)
+add_custom_target(ExtOpCp ALL  
+    DEPENDS ${dat_depends} TENSILE_LIBRARY_TARGET  
+    COMMENT "Copying .co and .dat files"
+) 
+
+file(GLOB FILES_TO_COPY "${CMAKE_CURRENT_BINARY_DIR}/*.co" "${CMAKE_CURRENT_BINARY_DIR}/*.dat")
```
stellaraccident: I don't understand how this can work: the glob will be evaluated at configure time but it is trying to find files that are generated at build time. I think this will only work after the first build, which would be wrong.

toolchain-windows.cmake
```diff
@@ -0,0 +1,73 @@
+# ########################################################################
+# Copyright (C) 2022-2023 Advanced Micro Devices, Inc.
```
stellaraccident, line 2: TheRock is ROCm's official windows build. If hipsdk needs a custom toolchain, then it is local to that, not global for Windows

stellaraccident (changes requested):
Fyi, we will shortly be deleting all of the default cmake files and switching to those in the next-cmake directory. The next-cmake build is tested on Windows via TheRock and likely shouldn't need any changes for basic Windows compat. The default cmake build has a number of portability problems that we do not want to address. 

I'd recommend using next-cmake files for the time being until we get them switched. 

Also, we can't accept changes upstream anymore which presume the hipsdk directory layout. If those are necessary, they are generally part of the hipsdk build, not part of the OSS build.

jayhawk-commits:

PR #2176, opened 2025-06-04
[next-cmake] Bump version

marbre:
The hipBLASLt (so)version was bumped with commit d3b681b.

### DIFF
--- a/next-cmake/CMakeLists.txt
+++ b/next-cmake/CMakeLists.txt
@@ -25,7 +25,7 @@
 # to confirm if they are required and document what we dropped if not.
 
 cmake_minimum_required(VERSION 3.22.1)
-project(hipblaslt VERSION 0.15.0 LANGUAGES CXX)
+project(hipblaslt VERSION 1.0.0 LANGUAGES CXX)
 
 list(APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake)
 list(APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/../cmake)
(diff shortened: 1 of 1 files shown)


next-cmake/CMakeLists.txt
```diff
@@ -39,7 +39,7 @@ include(hipblaslt_python)
 include(hipblaslt_target_configure_sanitizers)
 
 rocm_setup_version(VERSION ${PROJECT_VERSION})
-set(hipblaslt_SOVERSION 0.15)
+set(hipblaslt_SOVERSION 1.0)
```
bstefanuk, line 42:
```suggestion
set(hipblaslt_SOVERSION "${PROJECT_VERSION_MAJOR}.${PROJECT_VERSION_MINOR})
```
This sets the project version as the single source of truth as I don't see why the SO version and project version would be out of sync.

bstefanuk (approved):
Thanks for patching this. One minor improvement comment.

next-cmake/CMakeLists.txt
marbre: I think this depends on the used versioning scheme (which I am not aware of). If it is a semantic version number it indeed reflects the ABI version but if not, the ABI version and the version of the package might differ. No strong feelings though and for now I just ported what the commit applied to the "old" CMake config.

davidd-amd (approved):

next-cmake/CMakeLists.txt
marbre: I merged as is to unblock your rocBLAS next-cmake efforts @bstefanuk, but we can iterate and change that later if needed.

PR #2180, opened 2025-06-04
Cache tuning override file instead of opening each time

jsf0:
In hipblaslt.cpp , if HIPBLASLT_TUNING_OVERRIDE_FILE is used, the function hipblasLtMatmulAlgoGetHeuristic() calls override_path_compare_git_version(). This opens the file specified by HIPBLASLT_TUNING_OVERRIDE_FILE.

However, this appears to open the file on every GEMM kernel initialization, due to hipblasLtMatmulAlgoGetHeuristic() being called.

We noticed that when the tuning.txt file lives on an NFS share, performance drops dramatically. The following outputs are from Megatron-LM on a MI325X machine:

With tuning.txt on NFS:
```
throughput per GPU: 358.6
elapsed time per iteration: 21165.5
tokens/GPU/s: 6192.719284
mem usages: 0.7992
```

With tuning.txt on local NVMe:
```
throughput per GPU: 695.6
elapsed time per iteration: 10913.1
tokens/GPU/s: 12010.519467
mem usages: 0.7992
```

This PR would instead cache the file and avoid doing a full open and close cycle on every GEMM init.


jayhawk-commits:

PR #2181, opened 2025-06-04
Remove setting CMAKE_BUILD_TYPE in rocisa

davidd-amd:
- Remove setting the build type in rocisa


KKyang (approved):

tensilelite/rocisa/CMakeLists.txt
```diff
@@ -26,9 +26,6 @@ set(CMAKE_CXX_STANDARD 20)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_CXX_EXTENSIONS OFF)
 
-set(CMAKE_BUILD_TYPE Debug CACHE STRING "Build type" FORCE)
-message(STATUS "CMAKE_BUILD_TYPE set to ${CMAKE_BUILD_TYPE}")
-
 if(DEFINED Python_EXECUTABLE AND Python_EXECUTABLE)
```
marbre, line 29: An alternative could be to set it to Debug only if the build type isn't set. In IREE we have https://github.com/iree-org/iree/blob/33958f20dfd73c1a2e4b25bf6ba077a3c4138dfb/CMakeLists.txt#L41-L45 but deleting those lines here works as well.

marbre (approved):

tensilelite/rocisa/CMakeLists.txt
KKyang: This is added for testing only and forgot to remove it.

PR #2179, opened 2025-06-04
[next-cmake] Adds missing files and tweaks rocroller integration

davidd-amd:
- Adds new files to build resolving missing symbol error
- Use roc namespace for mxDataGenerator
- Attempt to find then fetch rocroller

PR #2182, opened 2025-06-04
[next-cmake] Add source and headers from Origami

bstefanuk:
Adds TensileLite::analytical symbols to the next-cmake build.


bstefanuk:
Duplicated by #2179

PR #2188, opened 2025-06-05
perf. script: Avoid timeout and make this runnable on gfx950

solaslin:
- [x] api_overhead: getAllSolution always exceed timeout. Reduced the iteration.
- [x] matmul: fixed no solution found on gfx950 due to f8n type. 

Changing f8n to f8 works on both gfx950 and gfx942 since client.cpp automatically converts f8 to f8n when gfx942 is detected.
Though we may run different problem-sets on different arch in the future. This PR is a temporary change for testing gfx950.


hcman2 (approved):
ok if ci passed

KKyang (approved):

PR #2189, opened 2025-06-05
Add deprecated V2 (force compile error)

KKyang:


hcman2 (approved):

jichangjichang (commented):
Carson decide that we should keep v2 and make v2 is identical with v1 but with deprecate attibute.

PR #2195, opened 2025-06-05
Refine the method for caculating sparse mfma offset at tile section

vin-huang:


vin-huang:
Those failures on gfx950 are known cases.

jichangjichang (approved):

Serge45 (approved):

PR #2199, opened 2025-06-06
Do not use sysconf when on windows

davidd-amd:
- Only import sysconf if not windows.


ScottTodd (approved):

marbre (approved):

bstefanuk (approved):
Thanks for patching this.

jichangjichang (approved):

PR #2202, opened 2025-06-08
[next-cmake] msgpack target management and use object library for internal libs

davidd-amd:
- Change hipblaslt-clients-common to and OBJECT library. The primary motivation for changing this target to an OBJECT library is consistency with the broader BLAS libraries updates where we are shifting to OBJECT libraries to remove synchronization points. In this case, we aren't exposing more parallelism because it is only consumed internal to the project but consistency adds value.
- Updates treatment of msgpack see description: https://github.com/ROCm/Tensile/pull/2145


next-cmake/CMakeLists.txt
```diff
@@ -263,11 +252,19 @@ if(HIPBLASLT_ENABLE_HOST)
...
+        find_package(msgpack-cxx CONFIG)
         if(msgpack-cxx_FOUND)
-            target_link_libraries(hipblaslt PRIVATE msgpack-cxx)
+            message(STATUS "Found msgpack-cxx (>=6.x)")
+            set(msgpack_target msgpack-cxx)
         else()
-            target_link_libraries(hipblaslt PRIVATE msgpackc)
+            find_package(msgpackc-cxx CONFIG REQUIRED NAMES msgpackc-cxx msgpack)
+            message(STATUS "Found msgpack (<=5.x)")
+            set(msgpack_target msgpackc)
```
stellaraccident, line 264: This approach massively sketches me out as a way to take a header only dep on this lib. Iiuc, in 6.x you can just depend on msgpackc-cxx vs reaching in to get its headers, but I would need to look closely at it (I am mainly deriving this from the fact that we have been building it this was for 4 months and do not have concrete library dep like we do for 5.x).

If this is true, we should code the 6.x branch the right way (ie. How it was) and only include the header only hack for the 5.x branch. That way, future us just deletes the 5.x branch vs trying to untangle again what was done here.

I'm going to go ahead and approve/land this, but I would like us to converge on a future correct way before propagating it. 

Once we do, please update the dependencies.md file in TheRock to note the canonical way we chose to do it.

next-cmake/clients/CMakeLists.txt
```diff
@@ -29,7 +29,7 @@ set(TEMP_TENSILE_CLIENT_SOURCE_DIR "${CMAKE_CURRENT_SOURCE_DIR}/../../tensilelit
 set(TEMP_TENSILELITE_DIR "${CMAKE_CURRENT_SOURCE_DIR}/../../tensilelite")
 
 add_executable(hipblaslt-bench)
-add_library(hipblaslt-clients-common STATIC)
+add_library(hipblaslt-clients-common OBJECT)
```
stellaraccident, line 32: Can you please comment on the rationale for this change in the PR description vs just saying it is being done. I don't disagree -- just want to see it documented.

stellaraccident (approved):

next-cmake/CMakeLists.txt
stellaraccident: Ok, confirmed by reading the msgpack>=6.x code (https://github.com/ROCm/TheRock/blob/main/docs/development/dependencies.md). In the >=6.x branch, `msgpack-cxx` is an interface library suitable for header only use. Therefore, the right thing there is to just use it with `target_link_libraries()`. Note also that it does set various interface options that do influence the behavior, so we really should be using the interface library vs just stealing its include directories.

Recommend changing the 6.x branch back to what it was:

```
            target_link_libraries(hipblaslt PRIVATE msgpack-cxx)
            message(STATUS "Found msgpack-cxx (>=6.x)")
```

Then in the 5.x branch, do the INTERFACE_INCLUDE_DIRECTORIES if that is the right way (I know that is the historic way we have done it, but I'm going to look at the code in a second to determine whether that is the right way).

stellaraccident: 5.x code: https://github.com/msgpack/msgpack-c/blob/cpp-5.0.0/CMakeLists.txt#L7

There, the interface library is `msgpackc-cxx`. So you have two options, neither of which I care much about because my opinion is that we shouldn't use old libs vs catering to them like this. However, if being conservative and preserving the letter of how we have historically done it, the 5.x branch becomes:

```
            find_package(msgpackc-cxx CONFIG REQUIRED NAMES msgpackc-cxx msgpack)
            message(STATUS "Found msgpack (<=5.x)")
            get_target_property(msgpack_inc msgpackc-cxx INTERFACE_INCLUDE_DIRECTORIES)
            target_include_directories(hipblaslt PRIVATE "${msgpack_inc}")
```

Although, I think it is also perfectly reasonable to just `target_link_libraries(hipblaslt PRIVATE msgpackc-cxx)`. I understand, though, if you want to do it exactly as it was done historically given how little testing we have for these old versions.

I'd just like the new version branch to be correct and not a hack.

davidd-amd: > This approach massively sketches me out as a way to take a header only dep on this lib. Iiuc, in 6.x you can just depend on msgpackc-cxx vs reaching in to get its headers, but I would need to look closely at it (I am mainly deriving this from the fact that we have been building it this was for 4 months and do not have concrete library dep like we do for 5.x).
> 
> If this is true, we should code the 6.x branch the right way (ie. How it was) and only include the header only hack for the 5.x branch. That way, future us just deletes the 5.x branch vs trying to untangle again what was done here.
> 
> I'm going to go ahead and approve/land this, but I would like us to converge on a future correct way before propagating it.
> 
> Once we do, please update the dependencies.md file in TheRock to note the canonical way we chose to do it.

Originally, I wanted to implement as follows:

```
find_package(msgpack-cxx CONFIG)
    if(msgpack-cxx_FOUND)
    message(STATUS "Found msgpack-cxx (>=6.x)")
else()
    find_package(msgpackc-cxx CONFIG REQUIRED NAMES msgpackc-cxx msgpack)
    message(STATUS "Found msgpack (<=5.x)")
    add_library(msgpack-cxx ALIAS msgpackc-cxx)
endif()
target_link_libraries(foo PRIVATE msgpack-cxx)
````

However, the mathci environments install multiple versions of msgpack and the erions that cmake prefers is an older version without the header only target (frown). I notified the mathci team and when they update the environments I can update how we manage msgpack. Alternatively, we can have a two-pronged approach.

next-cmake/clients/CMakeLists.txt
davidd-amd: See the updated description.

next-cmake/CMakeLists.txt
stellaraccident: What I wrote gets it there, right? (Ie. Move the target_link_libraries or target_include_directories into the branches of the version switch vs trying to normalize the.

PR #2236, opened 2025-06-13
Increase coverage for set_get_attr

tomchengchitang:
- Added test cases to increase the coverage
  - cover `hipblasLtMatmulDescSetAttribute` and `hipblasLtMatmulDescGetAttribute` in **rocblaslt_auxiliary**
- Increase the coverage from 51.1% to 54.56% (~400 lines)
 
image


hcman2 (approved):

library/src/amd_detail/hipblaslt.cpp
```diff
@@ -63,30 +63,6 @@ bool override_path_compare_git_version(OverrideSingleton& override, hipblasLtHan
     return false;
 }
 
-hipblasStatus_t hipErrorToHIPBLASStatus(hipError_t status)
```
jichangjichang, line 66: remove this API from hipblaslt_internal.hpp as well

jichangjichang (approved):

library/src/amd_detail/hipblaslt.cpp
tomchengchitang: I will remove this part in next PR.

PR #2241, opened 2025-06-13
New mode for stream-k launch grid prediction

AlexBrownAMD:
Adds a new mode for improved stream-k launch grid prediction. Performance analysis work is available.

### DIFF
--- a/tensilelite/Tensile/Source/lib/include/Tensile/Serialization/PredictionLibrary.hpp
+++ b/tensilelite/Tensile/Source/lib/include/Tensile/Serialization/PredictionLibrary.hpp
@@ -89,7 +89,8 @@ namespace TensileLite
                                 solution->sizeMapping.depthU, // MT_K
                                 solution->sizeMapping.matrixInstruction[0], // MI_M
                                 solution->sizeMapping.matrixInstruction[1], // MI_N
-                                solution->sizeMapping.matrixInstruction[2]); // MI_K
+                                solution->sizeMapping.matrixInstruction[2], // MI_K
+                                solution->sizeMapping.CUOccupancy); // Occupancy
                             lib.tile_list.emplace_back(solution_tuple);
                             lib.tile_map.insert(std::make_pair(solution_tuple, index));
                         }
--- a/tensilelite/Tensile/Source/lib/source/ContractionSolution.cpp
+++ b/tensilelite/Tensile/Source/lib/source/ContractionSolution.cpp
@@ -43,6 +43,8 @@
 #include <cstdlib>
 #include <random>
 
+#include <stdio.h>
+
 #ifdef ENABLE_ROCTX
 #include <roctracer/roctx.h>
 #endif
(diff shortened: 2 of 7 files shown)


tensilelite/Tensile/Source/lib/source/ContractionSolution.cpp
```diff
@@ -43,6 +43,8 @@
 #include <cstdlib>
 #include <random>
 
+#include <stdio.h>
```
pghysels: This seems unnecessary.

aazz44ss (dismissed):

tensilelite/Tensile/Source/lib/include/Tensile/Serialization/PredictionLibrary.hpp
```diff
@@ -89,7 +89,8 @@ namespace TensileLite
                                 solution->sizeMapping.depthU, // MT_K
                                 solution->sizeMapping.matrixInstruction[0], // MI_M
                                 solution->sizeMapping.matrixInstruction[1], // MI_N
-                                solution->sizeMapping.matrixInstruction[2]); // MI_K
+                                solution->sizeMapping.matrixInstruction[2], // MI_K
+                                solution->sizeMapping.CUOccupancy); // Occupancy
```
neoblizz, line 93: Does this impact the python-bindings in any way?

bethune-bryant: It impacts anywhere we're using the python-bindings. We'll need to update the macro-tile list that's being passed in as an argument to include occupancy.

yenong-amd (dismissed):

tensilelite/Tensile/Source/lib/include/Tensile/Serialization/PredictionLibrary.hpp
bethune-bryant: I think we can fix this in the places we're using it. This shouldn't block the merge.

bethune-bryant (dismissed):

tensilelite/Tensile/Source/lib/source/ContractionSolution.cpp
AlexBrownAMD: Yea, accidentally left some debug code in. Thanks for catching this, just posted an update.

pghysels:
Does this require retuning the grid?

AlexBrownAMD:
> Does this require retuning the grid?

This should provide a boost in performance for existing grid tuning, but a retune with this option will also be done.

hcman2:
Why do we need so many modes? Can we run different modes at the same time?

AlexBrownAMD:
> Why do we need so many modes? Can we run different modes at the same time?

Different modes are mostly for testing at the moment. Once we have a completed best solution for launch grid, we will reduce the number of modes to just that solution as default and the debug mode.

KKyang (approved):

hcman2 (approved):

PR #2263, opened 2025-06-19
Add fixed WGM mode for debugging

AlexBrownAMD:
Adds an environment variable that allows you to override the value of WGM. Useful for debugging Origami's dynamic WGM mode. This change also extends dynamic WGM to work with all kernels. Dynamic WGM is off by default, but can be tested by setting the environment variable to enable it.

### DIFF
--- a/tensilelite/Tensile/Source/lib/source/ContractionSolution.cpp
+++ b/tensilelite/Tensile/Source/lib/source/ContractionSolution.cpp
@@ -1240,7 +1240,7 @@ namespace TensileLite
         if(internalArgsSupport.useUniversalArgs)
         {
             auto defaultWGM = sizeMapping.workGroupMapping;
-            // Enable Origami dynamic WGM mapping for all kernels
+            // Allow Origami dynamic WGM mapping for all kernels (enabled by setting TENSILE_STREAMK_DYNAMIC_WGM)
             {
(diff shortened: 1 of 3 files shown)


tensilelite/Tensile/Source/lib/source/ContractionSolution.cpp
```diff
@@ -1240,10 +1240,12 @@ namespace TensileLite
         if(internalArgsSupport.useUniversalArgs)
         {
             auto defaultWGM = sizeMapping.workGroupMapping;
-            if(sizeMapping.streamK != 0)
+            // Allow Origami dynamic WGM mapping for all kernels (enabled by setting TENSILE_STREAMK_DYNAMIC_WGM)
```
hcman2, line 1243: Why remove if(sizeMapping.streamK != 0) here? Is it not necessary?

neoblizz: I think before it only worked for where SK != 0, i.e. the non-Stream-K kernel could not use DynamicWGM. Now this enables it for all kernels, even the ones that do not set Stream-K as enabled.

neoblizz (approved):

idass1990:
Closing the PR in this repo. Please refer to the migrated PR for updates.

PR #2265, opened 2025-06-20
[OPT] Tail Loop reload loop optimization

briannwu:
Details:
1. Tail loop adds a reload loop to check if other load tiles need to be reloaded in commit #62c84ce. This commit modify the way to do the reload loop. Using vgprAddr + sgprOffset can calculate the start offset of each thread in the load tile. Add the start offset calculated and load range of each thread(GRVW * bpe) - 1 can get the end offset. If the last valid element falls within the range means that the load tile need to be reloaded.
2. Fix wrong sLoadTileIdx calculation for LSU > 1


idass1990:
Closing the PR in this repo. Please refer to the migrated PR for updates.

PR #2298, opened 2026-03-11
Fix int32 overflow in rotating buffer size calculation

Z-Y00:
## Motivation
This is Lorri Rao from AIG model team (lorri.rao@amd.com), trying to fix tensile tunning error for grouped gemm.

## Technical Details
When rotating buffer sizes exceed ~2GB, the int32_t multiplication of rotatingNum * rotatingSize wraps negative (e.g. 838926336 * 3 = -1778188288), causing "Insufficient rotating buffer size" abort. 
Very similar like 4GB overflow error for 32 sized datatype in memory allocation.
Use int64_t for rotatingNum and totalRotatingSizeNeeded to handle large grouped GEMM buffer allocations correctly.

## Test Plan
NA

## Test Result
NA

## Submission Checklist

- [ y] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

### DIFF
--- a/tensilelite/client/src/DataInitialization.cpp
+++ b/tensilelite/client/src/DataInitialization.cpp
@@ -2504,21 +2504,21 @@ namespace TensileLite
             {
                 auto    castInputs   = static_pointer_cast<ContractionInputs>(inputs);
                 size_t  rotatingSize = getRotatingSize(*gemmProblem, *castInputs);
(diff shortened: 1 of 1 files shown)


Z-Y00:
Shall I submit PR in the mono repo instead?

